PR #8, opened 2016-05-30
Fast Fourier Transform widget

stuart-cls:
Here is the FFT widget I've been working on. 

Still outstanding:
  - There are no tests, we haven't really talked about that yet.
  - There is no documentation, the docs folder is missing from this example add-on 
    vs the one that I forked, is that intentional?
  - build_spec_table() was broken out into a utils.py in my add-on since it is
    useful for a lot of different widgets, thoughts?
  - Currently requires the proper Orange.data.Table format as input, which means
    it needs an importer for columnar data (DPT-like, very common for this
    data type). I used a transposing widget I made for now to import data.


orangecontrib/infrared/widgets/owfft.py
```diff
@@ -0,0 +1,594 @@
...
+    # independant variables (or "attributes" as Orange calls them)
+    domain = Orange.data.Domain(wn_vars)
+
+    # Finally, build the table using the damain and intensity arrays:
+    table = Orange.data.Table.from_numpy(domain, intensities)
+    return table
+
+class OWFFT(OWWidget):
+    # Widget's name as displayed in the canvas
+    name = "FFT"
```
borondics: I would call it something else than FFT. Maybe IFG->SP would be a good one, since it has many options/operations unrelated to FFT and the functionality is that...

orangecontrib/infrared/widgets/owfft.py
```diff
@@ -0,0 +1,594 @@
...
+    # An icon resource file path for this widget
+    # (a path relative to the module where this widget is defined)
+    icon = "icons/fft.svg"
+
+    # Define inputs and outputs
+    inputs = [("Interferogram", Orange.data.Table, "set_data")]
+    outputs = [("Spectra", Orange.data.Table)]
+
+    # Define widget settings
+    laser_wavenumber = settings.Setting(15797.337544)
```
borondics: We should keep the doors open for NanoIR instruments. They don't use lasers to measure the mirror position.

orangecontrib/infrared/widgets/owfft.py
```diff
@@ -0,0 +1,594 @@
...
+
+    def calculateFFT(self):
+        """
+        Calculate FFT from input interferogram(s).
+        This is a handler method for
+          - bad data / data shape
+          - splitting the array in the case of two interferogram sweeps per dataset.
+          - multiple input interferograms
+
+        Based on mertz module by Eric Peach, 2014
```
borondics: Thanks for acknowledging Eric!

Mertz was a person, right?

markotoplak:
Great, thanks! 

Do you have a file I can test it with? Maybe you could create a small example .tab file and put it into the orangecontrib/infrared/datasets folder.

You are in fact encouraged to put common functionalities out of the widget so that also other (widgets or users just running the code) could use it. For example, build_spec_table could be perhaps in orangecontrib/infrared/data.py.

The widget provides some FFT functions that could be also provided to users as a library. If they would be added to orangecontrib/infrared/fft.py they could be used both from the widget or just from an ordinary python library.

Is FFT for spectral data something specific orwhy don't you use an available library to perform it? I see you use some fft from numpy internally, which is great. You mention a mertz module: is it possible to run it directly? This might be a stupid question as I know very little about FFT.

The doc folder will come back ...

stuart-cls:
I have a test data file that is ~750 kB, I'll convert it to .tab format and add to the PR.

I'm unsure about breaking out into an fft module: while it's easy to do I don't see another application for the code. I suppose someone might want to use as an import in some othercode so might as well?

As for existing libraries, I tried my hardest to use existing numpy libraries. Feri is right: it's not really an FFT widget so much as an interferogram -> spectrum processor: the actual FFT is the easiest part and I didn't write my own. The apodization and to a lesser extent phase correction are pretty infrared-specfic. I had to do the zero-filling myself because the existing numpy zero-filling puts the zeros in the wrong place.

I'll upload the other changes discussed.

orangecontrib/infrared/widgets/owfft.py
stuart-cls: You're quite right. IFG->SP seems like a lot of acronyms to me though.

orangecontrib/infrared/widgets/owfft.py
stuart-cls: Interesting. I could change this setting to be directly the dx data point spacing, however that would be awkward for users with lasers specified in cm<sup>-1</sup>. How is the spacing defined in the NanoIR instrument?

orangecontrib/infrared/widgets/owfft.py
borondics: I didn't mean exactly that name, just wanted to add something that you
understand. ;)
You could call it Interferogram to Spectrum Conversion :)

orangecontrib/infrared/widgets/owfft.py
borondics: ​They measure the point spacing with an optical encoder usually, so they
just give you a spacing number in nm (or meter)​

stuart-cls:
Okay, I'm on board with breaking out the fft helper functions. What is the correct way to do this import? If I do `from orangecontrib.infrared.irfft import *` then it's not clear in the widget code where the functions are coming from but if I do `import orangecontrib.infrared.irfft` then the function calls are all really long in the code.

markotoplak:
Stuart,

```
from orangecontrib.infrared import irfft
irfft.yourfunction(parameters)
```

markotoplak:
If you have a smaller file, that still makes sense that would be even better. Can perhaps decrease resolution by keeping only n-th line?

stuart-cls:
Interestingly, the spectral resolution is decreased by chopping the wings off the interferogram, rather than increasing dx spacing. Anyway, I can do that easily.

markotoplak:
Ah, now I remember Feri explaining me about the beamlines with mirrors that move for even more than a few meters. Thanks for reminding me!

borondics:
​Yes, the length of the interferogram is related to the spectral resolution
and the spacing of the points or point density ​determines the highest
frequency you can detect.

stuart-cls:
I broke out some of the easier functions into a new `infrared.irfft` module. If you think this is a good approach, I can tackle the more difficult phase computation / correction and maybe the single_sweep function.

markotoplak:
Stuart, I think this is a good approach. 

The optimal division between code in a module and in a widget (for me) would be that the widget only handles user input/output, while all computation is in a module. Such modules are then independent of the user interface and as such we can also run them from a scripts. Orange power users frequently use Orange as a library (not widgets) and for them it is nice to have all Orange's functionality at their disposal. This also facilitates easier testing. 

Usually, when we develop Orange widgets, we first write the functionally as a module and then build the  widget.

markotoplak:
Thanks for your code!

I have rebased your commits in #9 and will accept that one.

PR #10, opened 2016-06-03
[owpreproc] Rubberband baseline subtraction

stuart-cls:
This calculates a rubberband baseline using the convex hull approach.

Notes:
  - I imported getx() from owcurves to avoid duplicating this (very) important function, but I think it should be sanity-checked and put into data.py or similar
  - I included an option to only show the calculated baseline without subtracting. Obviously not useful in repetitive calculations but I'm nervous about including baseline subtraction without a way for the user to check that a reasonable baseline was built.
  - The scipy interp1d function can do other kinds of interpolation than linear, is this a useful option to expose in the GUI?
  - Line 232 does a `copy.copy(data)`, which I based off the Gaussian preproc. Do we need to do this copy (doubles memory usage)? And if we do, why use `copy.copy(data)` instead of `data.copy()`?

### DIFF
--- a/orangecontrib/infrared/widgets/owpreproc.py
+++ b/orangecontrib/infrared/widgets/owpreproc.py
@@ -39,8 +39,11 @@ from Orange.widgets.data.owpreprocess import (
 import numpy as np
 
 from scipy.ndimage.filters import gaussian_filter1d
+from scipy.spatial import ConvexHull
 
 import copy
+
+from orangecontrib.infrared.widgets.owcurves import getx
 
 class GaussianSmoothing():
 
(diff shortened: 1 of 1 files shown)


markotoplak:
Stuart, thanks.

Yes, getx should be moved into data.py. I'll do it right away.

This widget is still missing the preview functionality that will allow us to see what is happening (this is what I am playing with now).

I think you should decide which options of interp1d should we expose. Which do you think are useful?

The data object has to be somehow copied. If not, we would directly modify the input data object of loaded date. I chose copy instead of constructing a new data table because everything but the contents of X stayed the same in my case (the same domain, the same other meta variables). I only used copy.copy() as I did not know that data tables also allow .copy() directly. I'll change this, thanks.

I did get some exceptions while testing your widget though, but these are not problematic at this stage. Try using your preprocessors in weird ways: for example with invalid input data or try to remove input data from it and so on.

stuart-cls:
Okay great.

I'll try to stress-test it and the next one a little more. I admit I'm not 100% confident with the logic of the preprocessor widget yet.

For the interp1d options, I think I'll leave the interpolation mode to the default (and most-used) linear unless there is a concrete use case presented.

I love the idea of a built in previewer, looking forward to seeing that.

PR #12, opened 2016-06-08
[owpreproc] Normalization preprocessor

stuart-cls:
Here is a (spectral) normalization preprocessor. It implements:
 - Simple min-max scaling
 - Vector normalization
 - Offset correction
with the option of limiting the normalization criteria to a specific spectral range.

Another thing we need to do with this widget is normalize to a meta attribute, for example scaling by a weight percent to correct for KBr pellet thickness. I was unable to figure out a good way of filling a combo box with the meta/class attributes of the data, since the BaseEditor isn't aware of what data is connected to it as far as I can tell. Help?

### DIFF
--- a/orangecontrib/infrared/widgets/owpreproc.py
+++ b/orangecontrib/infrared/widgets/owpreproc.py
@@ -282,6 +282,8 @@ class Normalize():
     def __call__(self, data):
         x = getx(data)
 
+        data = data.copy()
+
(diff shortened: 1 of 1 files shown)


markotoplak:
I got your question, but I do not have a good answer now. It is also connected with smart filewise limits that would come in handy for default values.

PR #15, opened 2016-06-22
[WIP][owpreproc] Remove markings when preprocessor is not focused.

stuart-cls:
This adds a few functions to support removing the markings that a preprocessor has added when it looses focus.

Motivation: I love that the markings appear when you click on a preprocessor, but it's going to get very busy as we add more markings for other preprocessors if they all hang around forever. I am planning to use this to make some of my preprocessors more interactive but wanted to make sure we're on the same page first.

PS Sorry about the whitespace noise, I just noticed it now.

### DIFF
--- a/orangecontrib/infrared/widgets/owcurves.py
+++ b/orangecontrib/infrared/widgets/owcurves.py
@@ -73,7 +73,7 @@ def searchsorted_cached(cache, arr, v, side="left"):
     if key not in cache:
         cache[key] = np.searchsorted(arr, v, side=side)
     return cache[key]
-        
+
 
 def distancetocurve(array, x, y, xpixel, ypixel, r=5, cache=None):
     if cache is not None and id(x) in cache:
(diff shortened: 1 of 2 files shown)


stuart-cls:
Should have tested this more before pushing...it works nicely except you can't actually use the markings any more! To be continued...

stuart-cls:
If `.curveplot.markings` is only used for preprocessor markings in this widget, we could do something like:

```
    def focusInEvent(self, event):
        super().focusInEvent(event)
        for m in self.widget().parent_widget.curveplot.markings:
            self.widget().parent_widget.curveplot.remove_marking(m)
        try: #active selection on preview
            self.widget().activateOptions()
        except AttributeError:
            pass
```

Except that doesn't work quite right when I chain two Cut preprocessors...

markotoplak:
I opted for your second suggestion and just added it in. For now it only deletes markings when some new preprocessor wants to add its own. Thanks!

PR #18, opened 2016-06-23
[owpreproc] Set normalization limits using gui markers

stuart-cls:
This PR uses the new gui markers to set the normalization limits (when desired).

Also includes some work to bring NormalizeEditor up to the level of CutEditor and some dataset sets to avoid crashing when passed an empty dataset (for example, from Cut set to 0.0-1.0)


markotoplak:
Sorry, I thought I have merged this already.

PR #19, opened 2016-07-05
Normalize: Implement normalize by attribute

### LINKED ISSUE #13: [owpreproc] Normalize by attribute
Issue to track remaining problem with #12 (Normalization Preprocessor)

Stuart:
> Another thing we need to do with this widget is normalize to a meta attribute, for example scaling by a weight percent to correct for KBr pellet thickness. I was unable to figure out a good way of filling a combo box with the meta/class attributes of the data, since the BaseEditor isn't aware of what data is connected to it as far as I can tell. Help?

Marko:
> I got your question, but I do not have a good answer now. It is also connected with smart filewise limits that would come in handy for default values.

stuart-cls:
This allows spectra to be normalized by some attribute, for example pellet weight percent.
Currently it only shows meta attributes as the normal attributes are all wavenumbers and overwhelm the dialog.
Closes #13

### DIFF
--- a/orangecontrib/infrared/widgets/owpreproc.py
+++ b/orangecontrib/infrared/widgets/owpreproc.py
@@ -30,6 +30,7 @@ from Orange.preprocess import Continuize, ProjectPCA, \
 from Orange.widgets import widget, gui, settings
 from Orange.widgets.utils.overlay import OverlayWidget
 from Orange.widgets.utils.sql import check_sql_input
+from Orange.widgets.utils.itemmodels import VariableListModel
 
 import Orange.widgets.data.owpreprocess as owpreprocess
 
@@ -673,11 +674,12 @@ class Normalize():
     # Normalization methods
     MinMax, Vector, Offset, Attribute = 0, 1, 2, 3
 
-    def __init__(self, method=MinMax, lower=float, upper=float, limits=0):
+    def __init__(self, method=MinMax, lower=float, upper=float, limits=0, attr=None):
         self.method = method
         self.lower = lower
         self.upper = upper
         self.limits = limits
+        self.attr = attr
 
     def __call__(self, data):
         x = getx(data)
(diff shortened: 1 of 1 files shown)


orangecontrib/infrared/widgets/owpreproc.py
```diff
@@ -704,8 +706,12 @@ def __call__(self, data):
             elif self.method == self.Offset:
                 data.X -= np.min(y_s, axis=1, keepdims=True)
             elif self.method == self.Attribute:
-                # Not implemented
-                pass
+                # attr normalization applies to entire spectrum, regardless of limits
+                # meta indices are -ve and start at -1
+                if self.attr not in (None, "None", ""):
```
markotoplak: Why doesn't just comparison with None suffice?

orangecontrib/infrared/widgets/owpreproc.py
```diff
@@ -730,6 +736,12 @@ def __init__(self, parent=None, **kwargs):
         self.lower = 0
         self.upper = 4000
         self.limits = 0
+        self.attrs = ['None']
```
markotoplak: I think this should be just None, not a string.

markotoplak:
I will accept this... We can address that "None" stuff later. If we worry about it now, we will have to deal with rebasing.

orangecontrib/infrared/widgets/owpreproc.py
stuart-cls: None is the default attribute, "None" is the string the widget sets as default, and "" is what the widget VariableListModel returns when there are no meta attributes. I agree this is messy but I liked having something in the combobox when there is no data/attribute to select. I'm not tied to it though.

orangecontrib/infrared/widgets/owpreproc.py
stuart-cls: That would be cleaner, yes. As I said above, I put it so that there was something in the combobox...

PR #29, opened 2016-07-29
Integrate: Match produced attribute name to user-defined window

stuart-cls:
Our users are having difficulty matching attributes between training and test data with the coerced value naming I was using before. This approach will also be more consistent between datasets and allows the user to apply a constant window regardless of data spacing.

### DIFF
--- a/orangecontrib/infrared/widgets/owpreproc.py
+++ b/orangecontrib/infrared/widgets/owpreproc.py
@@ -926,22 +926,17 @@ class Integrate():
         if len(x) > 0 and data.X.size > 0 and self.limits:
             x_sorter = np.argsort(x)
             all_limits = np.searchsorted(x, self.limits, sorter=x_sorter)
-            range_attrs = []
+            range_attrs = [Orange.data.ContinuousVariable.make(
+                            "{0} - {1}".format(limits[0], limits[1]))
+                            for limits in self.limits]
             newd = None
             for limits in all_limits:
-                try:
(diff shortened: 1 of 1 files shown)


markotoplak:
If I understand the problem correctly, there is only one long term solution here: automatic preprocessing of the testing data with the same preprocessors as used with the training data. Until now all infrared preprocessors create a new domain where features are defined just by names and we fill the value table with some precomputed numpy matrix. The recommended way we will have to move towards is to define domain of the preprocessed outputs as functions that can take a table (with not necessarily the same attributes) and convert it into a preprocessed table. 

This is already used in ordinary Orange preprocessors, which, instead of computing transformations explicitly, set compute_value parameters of features to a function. Then the transformed values are computed in the domain conversion process (this domain conversion process is then automatic for test data).  For now this kinds of transformations work only on single features which makes it too slow for the kind of data (and preprocessors) we are dealing with. We are planning to make transformations of whole domains accessible. I was studying the problem once but then went of to some other tasks, but this one is a must, really. Therefore this depends on changes in core.

By the way, what are the users doing with training and test data? Are they trying to do classification?

stuart-cls:
I think we are talking about the same thing, yes. I understand what you are saying and the required solution...


For the users, maybe it's not really train/test: they are making simple linear regression models based on integrated peak areas from standard samples and then using the resulting linear equation to predict concentrations in unknowns.

________________________________

PR #32, opened 2016-08-23
Omnic reader with PyMca

markotoplak:
Can read some files. Works properly with 150416-01 degman 10x10µm² 5x10 spc.map but not with dlg_cls_1.map (PyMca did not extract metainfo for that file).


borondics:
I just downloaded this update. Awesome! We will test with other *.map files.

PR #34, opened 2016-09-09
Opus reader

stuart-cls:
This PR will add the ability to load OPUS binary files. It depends on the opusFC module:

Issues to discuss before merging:
  1. I didn't rebase on the latest master as I noticed that #33 requires a version of Orange later than what our users can install (only 3.3.7 is available so far)
  2. I added opusFC to `install_requires` but I'm not sure about this: on the one hand, it means users installing orange-infrared automatically get OPUS support. But the downside is that users with problems with installing opusFC will be unable to install orange-infrared at all. I'm not sure there is a great solution to this.
  3. I haven't quite got MacOS .whls of opusFC up yet, so sorry Feri :)

I look forward to your comments and suggestions.

### DIFF
--- a/orangecontrib/infrared/data.py
+++ b/orangecontrib/infrared/data.py
@@ -123,10 +123,10 @@ class OmnicMapReader(FileFormat):
         pass #not implemented
 
 
-class OPUS2DReader(FileFormat):
-    """Reader for 2D OPUS files"""
-    EXTENSIONS = ('.0', '.1')
-    DESCRIPTION = 'OPUS 3D Spectra'
+class OPUSReader(FileFormat):
+    """Reader for OPUS files"""
+    EXTENSIONS = ('.0', '.1', '.2', '.3', '.4')
+    DESCRIPTION = 'OPUS Spectrum'
 
     @property
     def sheets(self):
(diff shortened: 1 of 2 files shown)


markotoplak:
I have tried it with my Opus 2D and 3D files. I like it a lot. I also compared its performance with my pure-python prototype from the end of April and your implementation is faster.

Basically the only complaint I would have is that map_x and map_y are not continuous variables, but this is something I can easily change myself.

a1. No problem.

a2. I think is good to require it now so that problems appear earlier. We could think about extras_require in the future.

a3. Nice work for linux! I saw 4 distinct packages on PyPI which should cover most of the users. It worked perfectly on my system.

stuart-cls:
Glad to hear it worked well on your setup. 

I'm surprised that map_x and map_y are not continuous, on line 166/167 they are defined as ContinuousVariable.make()

For the linux, I used the new manylinux project which provides docker images to build on a CentOS 5 system as a minimum base (https://github.com/pypa/manylinux) and can build for a bunch of python versions for both 32 and 64 bit. I should have time this week to get the MacOS ones up as well.

markotoplak:
Hmm, if they are Continuous, then there is a problem with sorting in Orange Table widget.... Need to check that.

PR #35, opened 2016-09-12
Change README.md to reflect that setuptools doesn't support wheels

stuart-cls:
See:
[1] https://github.com/pypa/setuptools/issues/78
[2] https://github.com/biolab/orange3-example-addon

Does this bug cause a problem with the installation from the Add-ons menu in Orange?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,11 @@ under the section Infrared.
 For developers
 --------------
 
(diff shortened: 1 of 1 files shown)


markotoplak:
Thanks! This is not connected with installation from the menu.

PR #43, opened 2016-10-20
Absorbance / Transmittance preprocessors

stuart-cls:
This version only exposes conversion from AB -> TR and TR -> AB but the goal is to use for calculation from single-channel data as well.

I thought I would push this first to make sure I'm using the new preprocessor framework properly. And tests!

### DIFF
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -327,3 +327,99 @@ class Interpolate(Preprocess):
         domain = Orange.data.Domain(atts, data.domain.class_vars,
                                     data.domain.metas)
         return data.from_table(domain, data)
+
+class AbsorbanceFeature(SharedComputeValue):
+
+    def __init__(self, feature, commonfn):
+        super().__init__(commonfn)
+        self.feature = feature
+
+    def compute(self, data, shared_data):
+        return shared_data[:, self.feature]
+
+class _AbsorbanceCommon:
+
+    def __init__(self, ref):
+        self.ref = ref
+
+    def __call__(self, data):
+        if self.ref:
+            # Calculate from single-channel data
+            absd = self.ref.X / data.X
+            np.log10(absd, absd)
+        else:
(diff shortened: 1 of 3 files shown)


markotoplak:
I like your tests with roundtrip!

You are using the preprocessor framework for transforming multiple features at the same time properly. But as this transformation is feature independent, the old approach that would handle domains per-attribute would work just as well and be less complex (you would not need the *Common classes).

There are still some things we will have to handle, for example checking whether ref and data have the same domain and decide what to do if not.

I think these are good commits, therefore I am merging. We can modify later if needed.

stuart-cls:
Thanks for the feedback. For the domain transformation which is feature-independent, should I have just put the calculation logic in a per-attribute function and called that from `Orange.data.ContinuousVariable(compute_value=fn())`? Would that make the transformation slower? (i.e. instead of using the numpy `=/` and `log10` array functions, essentially doing the calculation one attribute at a time like a `for` loop)

I think matching domains is going to be a recurring problem, and we should have a consistent policy for handling that. My suggestion would be:
  1. If domains are substantially similar, silently merge  
      where "substantially similar" is
      - x values match within rounding errors
      - ignore data regions outside smallest x limits (i.e. 400-1000 cm<sup>-1</sup> + 500-1000 cm<sup>-1</sup> -> 500-1000 cm<sup>-1</sup>)
  2. If domains are not substantially similar, throw an exception and force the user to manually match the data domains, using a separate widget. This is to prevent "surprise" interpolation / data manipulation.

I think this is basically how `infarared.widgets.owfiles.load_data()` already works, so it could be generalized to implement this policy?

markotoplak:
Stuart, about single transformation, yes, you are right, that would be fine. In your current case the transfomation would not be slower.

But please leave these as they are, because domain matching will need the whole domain. (I only figured that out later) :)

I like your proposal. I thought of implementing something similar tomorrow. I think a possible solution would be to build this kind transformer into the domain when opening spectral files.

PR #52, opened 2017-03-01
OPUSReader: OpusFC 1.1.0 support (multiregion and time-resolved 3D files)

stuart-cls:
This PR adds support for OpusFC 1.1.0 including:
 - multiregion 3D map files
 - time-resolved 3D files
 
Other changes:
 - Direct data structure matching using return Object type (instead of guessing)
 - Simplification of parameter reading code, including a fix to a previously unused code path.

@markotoplak it would be great if we could get a small release with this change for our mapping file users.

### DIFF
--- a/orangecontrib/infrared/data.py
+++ b/orangecontrib/infrared/data.py
@@ -244,43 +244,33 @@ class OPUSReader(FileFormat):
         else:
             raise ValueError("Empty or unsupported opusFC DataReturn object: " + type(data))
 
-        try:
-            stime = data.parameters['SRT']
-        except KeyError:
-            pass # TODO notify user?
-        else:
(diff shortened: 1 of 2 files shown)


markotoplak:
Great, thanks!

I tested it with some of the old OPUS files and it seems to be working well.

Could you also, in a separate pull request, add some tests that read OPUS files and get some values out of then? It would be nice to have a file for each reader variant. The files should be as small as possible, perhaps spanning only a few wavelengths or points with images. If they are small, they can be added into datasets folder directly, if not, we could test them as in TestNea.

stuart-cls:
Yes, I would have included tests normally (I have a series that I run on the opusFC library itself) but I don't have a representative data file that I can share publicly at the moment. I'll file an issue to track this and once the beamline is operating again (May) I will add tests.

PR #54, opened 2017-03-02
Resampling Plotted Curves

borondics:
Added resampling capability in Curves to be able to see more than the initial random selection

### DIFF
--- a/orangecontrib/infrared/widgets/owcurves.py
+++ b/orangecontrib/infrared/widgets/owcurves.py
@@ -328,6 +328,7 @@ class SelectRegion(pg.LinearRegionItem):
 
 
 class CurvePlot(QWidget, OWComponent):
+    sample_seed = 314
     label_title = Setting("")
     label_xaxis = Setting("")
     label_yaxis = Setting("")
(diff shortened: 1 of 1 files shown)


stuart-cls:
@borondics 
Feri, the problem with this approach is if I share a worksheet with you, we're no longer guaranteed to be looking at the same N sample spectra. I think deterministic, reproducible data presentation is critical for goals of reproducible, audit-able data analysis.

Thoughts?

borondics:
Hi Stuart,

You are right about the comparison, but we actually had this request from
quite a few people during the workshop in ELETTRA this week. Nevertheless
it would be great if we could add a standard view as well to satisfy the
comparison needs. How do you think we could do it best?

F

markotoplak:
I agree with @stuart-cls . It has to repeatable and also it has to be saved with the schema (and has to reset to the usual - Random(0) view - for unseen data).

I would also prefer code reformatting commits to be separate from commits that make functional changes.

stuart-cls:
Perhaps it would be helpful to nail down what the real root request is: do your ( @borondics ) users really want to be able to see a different random sampling of spectra or do they actually want to see a specific sub-set but are using random sampling as a work-around?

borondics:
Ideally they want to see all spectra, but Marko is very right about
downsampling it, since nobody can really see differences even when you
display 100 of them (although this argument works for comparisons as well).
Cycling through can give you an idea though about the distribution of
spectra.

I am working on something, if I can manage to finish it today I'll push the
new code.

borondics:
OK, I added a "reset" function that always shows the same set of curves, while the "resample" function cycles through them.

PR #57, opened 2017-03-10
Resample curves

markotoplak:
This is code by @borondics from PR #54 with added state saving and minor changes. What do you think?


borondics:
Looks OK to me, although I have to add that an initial seed of Pi*100 is much more exciting than 0...

stuart-cls:
This looks good to me and addresses my earlier concerns wrt reproducibility. Thanks!

PR #58, opened 2017-03-10
[RFC] Fix vector normalize

stuart-cls:
Looking for some feedback, not quite ready to merge.

### Rationale
Current vector normalization routine does not return the correct value.

### Changes
Replaced custom implementation with `sklearn.preprocessing.normalize`, since it
(a) handles sparse data, (b) is faster and (c) minimizes unnecessary data copies.

In addition, "OPUS-style" vector normalization includes mean-centering the spectra
_before_ vector normalization, which is not what most users will expect, so I broke that out. Mean-centering is also useful for some data treatment workflows.

### Interested in feedback on 
  1. Naming of the different normalization routines
  2. Refactoring the normalization routines into separate functions. Specifically,
      is there an example best-practice I could follow, there are so many different
      approaches in our/Orange codebase. I was thinking of following the Integrate preprocessor.
  3. `Normalize.Vector` and `Normalize.MeanVector` fail `test_no_samples` since 
      `sklean.preprocessing.normalize` can't accept inputs with shape (0, n)

Thanks!

### DIFF
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -272,22 +272,11 @@ class _NormalizeCommon:
         else:
             y_s = data.X
 
-        if self.method == Normalize.MinMax:
-            data.X /= nanmax(np.abs(y_s), axis=1).reshape((-1,1))
-        elif self.method == Normalize.MeanOffset:
-            # zero offset correction applies to entire spectrum, regardless of limits
-            y_offsets = nanmean(data.X, axis=1).reshape((-1,1))
-            data.X -= y_offsets
-        elif self.method == Normalize.Vector:
+        if self.method == Normalize.Vector:
             data.X = sknormalize(data.X, norm='l2', axis=1)
-        elif self.method == Normalize.MeanVector:
-            # zero offset correction applies to entire spectrum, regardless of limits
-            # TODO refactor methods into separate functions
-            y_offsets = nanmean(data.X, axis=1).reshape((-1,1))
-            data.X -= y_offsets
-            data.X = sknormalize(data.X, norm='l2', axis=1)
-        elif self.method == Normalize.Offset:
-            data.X -= nanmin(y_s, axis=1).reshape((-1,1))
+        elif self.method == Normalize.Area:
+            # Not implemented yet
+            pass
         elif self.method == Normalize.Attribute:
             # attr normalization applies to entire spectrum, regardless of limits
             # meta indices are -ve and start at -1
@@ -300,9 +289,9 @@ class _NormalizeCommon:
 
 class Normalize(Preprocess):
     # Normalization methods
-    MinMax, Vector, Offset, Attribute, MeanVector, MeanOffset = 0, 1, 2, 3, 4, 5
+    Vector, Area, Attribute = 0, 1, 2
 
-    def __init__(self, method=MinMax, lower=float, upper=float, limits=0, attr=None):
+    def __init__(self, method=Vector, lower=float, upper=float, limits=0, attr=None):
         self.method = method
         self.lower = lower
         self.upper = upper
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -259,9 +259,7 @@ class _NormalizeCommon:
         if data.domain != self.domain:
             data = data.from_table(self.domain, data)
 
-        x = getx(data)
-
-        if data.X.shape[0] is 0:
+        if data.X.shape[0] == 0:
             return data.X
         data = data.copy()
 
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -8,6 +8,7 @@ from scipy.ndimage import gaussian_filter1d
 from scipy.spatial.qhull import ConvexHull, QhullError
 from scipy.signal import savgol_filter
 from bottleneck import nanmax, nanmin, nansum, nanmean
+from sklearn.preprocessing import normalize as sknormalize
 
 from orangecontrib.infrared.data import getx
 
@@ -18,7 +19,7 @@ def is_monotonic(a):
 
 
 class SelectColumn(SharedComputeValue):
-    
+
     def __init__(self, feature, commonfn):
         super().__init__(commonfn)
         self.feature = feature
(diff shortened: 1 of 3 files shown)


markotoplak:
Great!

1. @borondics , could get Christoph involved with this? He had some ideas during our workshop in Trieste.
2. I think a class per routine would be the best. Then there is less confusion of parameters. Of course all of them could use the same superclass if this makes the code more elegant.
3. Then we should handle this edge case with a simple if.

clsandt:
Hello Stuart and Marko.
My observation on the Normalization during the training was about the same problem than Stuart is mentioning : Opus also mean centers the spectra when normalizing which creates shifts in baseline position. This is counter-intuitive and requires an additional correction to remove the shift or it may introduce errors in the subsequent data analysis process. This mean centering step should be removed.
The standard practice is to - first baseline correct the spectra -second normalize the baseline corrected spectra. 
Another observation is that the offset correction which is found in Infrared orange Normalization preprocessor (Opus style) is not a normalization 'per se' for spectroscopy. It should be in the baseline correction options.
 
Regarding the naming of the normalization routine maybe we should differ from the Opus convention. Depending on the software suit it can be named "Area  normalization" or "Unit vector normalization" and can be processed slightly differently : 

Area normalization would normalize all the spectra in the dataset to a common Area Under Curve for a given spectral range with the equation Xnorm=X/sum(Xi). It doesn't work if the spectra have positive and negative values which is always the case with derivative spectra.
That's why Unit vector normalization is often preferred: it normalizes spectra to unity Xnorm=X/rootsquare(sum(Xi²)) and works even when positive and negative peaks are present
Both are doing the same in the end but the Unit vector normalization has less risk of introducing errors.

However, there is an additional complication: it is not always desirable to normalize the whole spectrum on the complete spectral range. You may want to normalize the spectrum using the surface of a particular peak. Opus has a lot of options on how to do that. Infrared Orange doesn't manage that.

Thus I would propose to have two normalization options: one (why not called Vector normalization) that would normalize the whole spectral range, one second option (why not called area normalization) that would ask for a specific peak/spectral range and then normalize the whole spectrum relative to this peak.

stuart-cls:
Christoph,

Thank you for these suggestions, I think they are great. I especially agree on your ideas to reorganize into logical sections of baseline correction and normalization, no need to slavishly follow OPUS.

I will work on this and request more feedback in the future.

Note that it is possible to normalize to an area using the Attribute normalization function but it is quite difficult to use: I think your suggestion is superior.

clsandt:
Hi Stuart
thanks for your comment and your programming skills.
Even if we don't fully follow Opus, it still has one of the largest variety of options for area measurement  (8, I think), we should get inspiration from that.

stuart-cls:
I've implemented what we discussed:

 - There is now a unit vector normalization, area normalization (which simply pulls in the integrators from Integrate() and so will be extended as we add more types) and the attribute normalization. 
 - Offset/baseline subtraction routines are removed, I will re-add them to a generalized SubtractBaseline preprocessor
 - Cleaned up some of the UI

@markotoplak The routines are now only 1-2 lines each so I didn't do the refactoring discussed above. I did fix the `test_no_samples` failure but now `TestConversion.test_slightly_different.domain` is failing on the `Normalize(method=Vector)` preprocessor and I can't figure out why.

clsandt:
So quick!!

markotoplak:
I added a few lines in https://github.com/markotoplak/orange-infrared/tree/fix-vector-normalize-marko to catch possible exceptions that got hidden in the Orange's testing functions.

Using the code above I can see that the tests fails because there are problems with unknowns. That tests crashes because the new vector normalization does not handle nan columns.

markotoplak:
As the number of methods has decreases - which is good - did you consider what would happen if a schema with older settings (with higher method IDs) with was loaded? There might be a crash - I did not check it.

clsandt:
Thanks for solving all of this.
Since there seem to be an integrating function already coded, would it be possible to use it to create a widget or preprocessor to measure the integrated area under the curve of a peak in a spectrum? That's not a multivariate approach but that's something we, spectroscopists do all the time ;)

stuart-cls:
Good catch on the unused getx() and ==. I did handle old method IDs in 5b4fdd8144bd8383d4a43c5eff5627e2b16c9b2e (line 736).

@clsandt See Integrate in the preprocessor widget. It works best if you put the Preview thing above the Integrate step so you can see what you're doing.

I'm a bit stumped on the test failure. If I replace sknormalize with

    norm_dataX = np.linalg.norm(data.X, axis=1, keepdims=True)
    data.X /= norm_dataX

it gives the same result and doesn't trigger the Nan error, but of course the Nan problem itself doesn't go away. I still get aucnow = [ 0.5 ] and aucorig = [ 0.965698 ].

If I replace the `Normalize(method=Normalize.Vector)` in the test_preprocessor.PREPROCESSORS list with `Normalize(method=Normalize.Area, int_method=Integrate.PeakMax, lower=0, upper=10000)` I get the same test failure, except aucorig = [ 0.998541 ]. So it seems that we have a general problem handling this column of Nans.

stuart-cls:
> Hm, solving the unknown problem was not that trival. The bad AUC remained because linalg.norm of something containing NaN was NaN even though there is no error. I am merging this and adding a commit which aims to fix it. Please check whether it is OK.

I'm not sure this is the correct fix. In your test of the NaN-handling, we have
````
[ 2, 2, 2, 2 ] , norm = 4
    -> [ 0.5, 0.5, 0.5, 0.5]
[ 2, 2, 2, NaN ] -> [ 2, 2, 2, 0 ], norm = 3.46 
    -> [ 0.57, 0.57, 0.57, 0 ] * sqrt(3/4) 
    -> [ 0.5, 0.5, 0.5, 0 ]
    -> [ 0.5, 0.5, 0.5, NaN ] 
````
but there are two problems with this. One, the correct vector norm of [ 2, 2, 2, 0 ] or [ 2, 2, 2 ] is 3.46, not 4. 
Two, if you replace your test array with [ 2, 2, 2, 3 ] then the norm is 4.58 and correcting [ 2, 2, 2, NaN ] back to 0.5 won't match anyway.

So the question is: should we return the array normalized to the vector norm of each spectrum, regardless of the array size, or do we include this correction which essentially calculates the norm using the mean as an extrapolation value?

stuart-cls:
Separately, adding 

    Normalize(method=Normalize.Area, int_method=Integrate.PeakMax, lower=1200, upper=1300),
    Normalize(method=Normalize.Area, int_method=Integrate.PeakMax, lower=0, upper=10000),

to `PREPROCESSORS` in the tests shows that only the second one fails. The domain conversion produces an all-NaN array which I can't replicate by calling either Normalize.Area or Integrate.PeakMax with the same limits directly on the collagen dataset.

markotoplak:
Yes, exactly, this is the question. I do not like pretending that unknowns
are zero (which this solves), but agree that a linear interpolation with
edge fill  could be better.

markotoplak:
@stuart-cls Can I get the same results as with the old kind of Vector normalization with current preprocessors? If I understood correctly I would need to mean-centre manually. How do I do it now?

PR #64, opened 2017-04-25
OWCurves: showing_sample should be called after update_view()

stuart-cls:
self.curveplot.sample_indices isn't populated until update_view() is called. 

Also I don't think the other warning works either, if I feed `titanic.dat` to curves it doesn't trigger.


markotoplak:
Thanks!

PR #66, opened 2017-04-26
[RFC] Average spectra preprocessor

stuart-cls:
Preprocessor to calculate the mean of either the entire group of spectra or grouped by discrete variable. Can also output the standard deviation.

This is basically stolen from the curves widget, it wasn't really setup for re-use directly.

@markotoplak the change to nanmean was required to pass the propagate unknowns test, note that if you feed the dataset from that test[1] into OWCurves and choose Average view, it has the same problem.

    [1] 
    data = self.collagen.copy()
    # one unknown in line
    for i in range(200):
        data.X[i, i] = np.nan

### DIFF
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -719,3 +719,58 @@ class Transmittance(Preprocess):
         domain = Orange.data.Domain(
                     newattrs, data.domain.class_vars, data.domain.metas)
         return data.from_table(domain, data)
+
+
+def _split_by_value(data, avar):
+    # from owcurves.CurvePlot._split_by_color_value()
+    rd = {}
+    if isinstance(avar, str):
+        rd[None] = np.full((len(data.X),), True, dtype=bool)
+    else:
+        cvd = Orange.data.Table(Orange.data.Domain([avar]), data)
+        for v in range(len(avar.values)):
+            v1 = np.in1d(cvd.X, v)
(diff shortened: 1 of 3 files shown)


markotoplak:
This is good, it is only a question whether this belongs in the Preprocess widget. All the other proprocessors return a single line on the output for a single line on the input. So I am not sure about it. 

Any different ideas? It could also be a new widget. Also, there is already a widget that computes all this stuff and can also split according to discrete variables but does not output it (Box Plot) - maybe fix that one?

The other thing is that one test fails (test_slightly_different_domain). I am not sure whether packaging it differently (like other Preprocessors that can handle domain conversion) would help as it is very special. If we keep it as a Preprocessor we could just skip it in that test.

stuart-cls:
I didn't consider Box Plot but I like it, it already has a nice visualization and calculates the values we need. Perhaps we (you, I guess) could add another output ("Statistics" maybe?) to Box Plot and that would serve the need we have.

For the output structure, I suppose we could have [Variable mean] [Variable Std Dev] [Variable - whatever] columns instead of rows? I wasn't sure about that when I was making the preprocessor, I just chose rows arbitrarily.

I have had troubles with test_slightly_different_domain before, it fails intermittently on this branch for me so I thought it was something on my setup.

@scott-cls any thoughts?

stuart-cls:
Closing in favour of #219

PR #68, opened 2017-05-04
Improved exception handling for the specific HERMES HDF5 reader

borondics:
OK, I am trying to test both the file handling and moved up the x_locs on the top in case that gives an exception.


markotoplak:
Is there perhaps any special signature inside these files? A string which could make us sure that it is Hermes if we find it?

borondics:
There is actually!!! I'll include it in an update in a few minutes/hours.

markotoplak:
If you are at it, if x_locks and y_locks are optional and the rest is always in, then the try should only contain these parts.

PR #75, opened 2017-06-05
Added saving capability to the top part of Hyperspectra

borondics:


borondics:
I would still like to have to modifications: 
- 300 dpi png resolution (I think it is in Orange, so couldn't/didn't change anything)
- save only the area of the picture, not the window

PR #76, opened 2017-06-23
[RFC][WIP] Spectral calculator widget

stuart-cls:
Here is a quite rough draft of a Spectral Calculator widget I started working on a while ago. I just quickly updated it and tried to remove a few anachronisms from our earlier development style. 

I'm interested in mostly big picture feedback: is this how we want the calculator to work? These three example calculations are ones that I have needed but no doubt there are others to be added (Reflectance, off the top of my head) and perhaps a different interaction model.

Todo
-----

- [ ] Reflectance
- [ ] Custom calculations

### DIFF
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -839,3 +839,47 @@ class Transmittance(Preprocess):
         domain = Orange.data.Domain(
                     newattrs, data.domain.class_vars, data.domain.metas)
         return data.from_table(domain, data)
+
+class SubtractFeature(SelectColumn):
+    pass
+
+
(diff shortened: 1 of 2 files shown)


markotoplak:
Aha, I have tried this PR and it is the calculator for the whole spectra, which is good and need, of course.

This would be a different widget, but anyway: I was thinking of a different kind of calculator - one that could, for example, compute a ratio between two integrals. A kind of simpler Feature constructor widget which would work only with continuous features.

borondics:
I think the integral ratio calculation should be part of the Integrator widget...

stuart-cls:
We don't need a special widget. See #198.

PR #83, opened 2017-07-18
OpusReader: Fix Start Time meta variable loading

stuart-cls:
Small mistake that prevented Start Time from being properly assigned as a TimeVariable.


markotoplak:
Great! Do you perhaps have a collection of files that could serve for automatic testing? So that such things would not happen happen?

Preferably files made as small as possible but with all the features that are supported.

stuart-cls:
Good point! I’ll try to work on adding tests, right now my opusFC code tests are on large files but I should be able to make some smaller ones.

PR #94, opened 2017-10-23
Added new files for documentation and completed the Spectra widget help

borondics:
_En route_ towards better and more complete documentation.


markotoplak:
It seems that you forgot to actually commit the changed files.

borondics:
Ah, good catch - I had the *.md files in the commit list instead of the *.rst ones. I fixed that in the last two commits.

PR #95, opened 2017-10-24
Refactored file names to match widget names

borondics:
- Matched file names and widget names 
- Changed the name of the "Files" widget to "Multifile" to avoid confusion with the basic Orange "File" widget name.


markotoplak:
I agree with Files -> Multifile.

And with the following:
owcurves -> owspectra

But I would be shorter with the other (I would not use _spectra at the end)
owpreproc -> owpreprocess
owmapbuilder -> owreshape

Does owfft really need renaming? I do not think there need to be strict 1:1 correspondance with widget names. Also, here we need to check for backwards compatibility of saved schemas. And tests fail now. :)

I think I should do this rename, let's just decide how should we name stuff.

borondics:
OK, we can leave "_spectra" out and also have FFT unchanged

I did the refactoring from Pycharm - should have handled it.

But go ahead and make these changes!

f

markotoplak:
Replaced with #98

PR #98, opened 2017-10-25
Rename widgets

markotoplak:
This has to be tested with some old saved workflows - they should still load with all settings.

When this is merged, older Infrared versions will not be able to open newly created workflows.


markotoplak:
@borondics @stuart-cls Could you review try this?

PR #117, opened 2017-12-01
Add file readers for Agilent (Resolutions Pro) image and mosaic files

stuart-cls:
Does what it says on the tin.

Credit to Alex Henderson for the MATLAB code this is based on.
https://bitbucket.org/AlexHenderson/agilent-file-formats

### DIFF
--- a/orangecontrib/infrared/agilent.py
+++ b/orangecontrib/infrared/agilent.py
@@ -1,0 +1,297 @@
+__version__ = "0.1"
+from pathlib import Path
+import struct
+
+import numpy as np
+
+DEBUG = False
+
+def _check_files(filename, exts):
+    """
+    takes filename string and list of extensions, checks that they all exist and
+    returns a Path
+    """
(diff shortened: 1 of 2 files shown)


borondics:
We can already do this via the ENVI export in Agilent. It works for mosaics too...
Nevertheless, it would be even cooler not to have to export the header file.
How about the video images in the file - can it grab that?

stuart-cls:
I didn't implement visual image export yet, mainly because we don't have a good place to put them in Orange. Let's discuss: #118

markotoplak:
Great!

Can you also add some (small) test data sets and tests that make sure they are read correctly?

If you can not produce something small you can send large files to me and we can handle it in the same was that "dust" is tested.

stuart-cls:
Hi Alex,
 (@AlexHenderson )
Thank you for the MATLAB code, it was very illuminating! 

I'm aware of the function of the .drd files and removing the .drd and .seq test is on my to-do list. Your suggestion of compressing them is great, I hadn't thought of that. 

Best,
Stuart

PR #122, opened 2017-12-20
Rename widget category: Infrared -> Spectroscopy

markotoplak:
Any objections? @stuart-cls, @GioB81

### DIFF
--- a/README.pypi
+++ b/README.pypi
@@ -1,4 +1,4 @@
-Orange Infrared
+Spectral Orange
 ===============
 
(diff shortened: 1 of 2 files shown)


scott-cls:
I think this is a great move.

GioB81:
Totally agree, I'd split in Spectroscopy and Imaging ( or chemical
Imaging)...

Il 20 Dic 2017 14:41, "scott-cls" <notifications@github.com> ha scritto:

markotoplak:
@GioB81 , we will split Imaging where there is more to the Imaging. :)

PR #124, opened 2017-12-21
Hyper and Spectra: multiple group selection

markotoplak:
Similar selection of multiple groups as in Scatter plot.

Difference between selected groups is not yet shown in Spectra because I do not have a good idea how should it look.


stuart-cls:
This is super cool! 👍 👍 👍

PR #128, opened 2017-12-24
Unify transformations in one preprocessor

borondics:
With more and more conversions coming (and not only IR, I hope) this will be necessary to keep the main drop-down menu uncluttered and also allow adding conversions easily.

### DIFF
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -283,6 +283,56 @@ class RubberbandBaseline(Preprocess):
         return data.from_table(domain, data)
 
 
+class LinearBaselineFeature(SelectColumn):
+    pass
+
+
+class _LinearBaselineCommon:
+
+    def __init__(self, peak_dir, sub, domain):
+        self.peak_dir = peak_dir
+        self.sub = sub
+        self.domain = domain
+
+    def __call__(self, data):
+        if data.domain != self.domain:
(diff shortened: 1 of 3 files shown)


borondics:
There is no test for the actual calculation though. Should we handle this dynamically or hard-code each transformation?

markotoplak:
If our goal is not to clutter the widget I agree with joining transformations in the widget.

But I think that in the scripting part (preprocess.py) it is clearer if each kind of transformation is in its own class. If there are lots of common computation we can make a superclass that does the repeated stuff.

Of course, as you say, additional tests that actually test the computation would be useful.

stuart-cls:
I agree with @markotoplak that it doesn't make sense to add complexity to the scripting interface. Please keep the preprocessors the way they are.

If you want to combine the graphical interface, I am fine with that.

borondics:
@markotoplak @stuart-cls, I also like the way the GUI and the calculations are separated in principle, but then we need to solve the issue of varying data ranges and the way the GUI learns about the data to properly initialize the GUI elements. I am sure I can hack something together, but maybe it can be done properly as well. 

How should this be done in your opinion?

stuart-cls:
@borondics did you see my response in #131 ? Initializing the GUI with sample data is already possible (and done for many Editors)

borondics:
OK, I'll check it out!

PR #132, opened 2017-12-28
Shift spectra

borondics:
Add a preprocessor for spectrum shift


markotoplak:
Cleaned up variant in #135

PR #134, opened 2018-01-02
Agilent file tests

stuart-cls:
Agilent file format reader sample data and tests.

@markotoplak let me know if they are too large. Each set (image, mosaic) totals under 100 kB I think.


borondics:
Stuart,

Could you upload the corresponding .hdr (envi) files too? Then we can test
your reader and the one that uses the envi bridge.

Feri

markotoplak:
Thanks, this is great. Of course, if you add the .hdr files that would be great also.

stuart-cls:
@borondics great idea! I will do that.

PR #135, opened 2018-01-03
 Add spectrum shift preprocessor

markotoplak:
Cleaned up #132 

@borondics , can you check if it still works as it should?

### DIFF
--- a/orangecontrib/infrared/preprocess.py
+++ b/orangecontrib/infrared/preprocess.py
@@ -932,8 +932,8 @@ class CurveShiftFeature(SelectColumn):
 
 class _CurveShiftCommon:
 
-    def __init__(self, sd, domain):
-        self.sd = sd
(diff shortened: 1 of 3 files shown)


stuart-cls:
This seems to work, but I wonder why it is not part of baseline correction as "Offset" or similar? Especially considering all the interest in reducing preprocessors in the GUI in #128 . Maybe I'm missing something?

markotoplak:
@stuart-cls That is a good point. I am merging this one for now, but yes, we can and should reorganize if we see fit.

borondics:
I think spectral shift can be an operation by itself without baseline
correction, although I made this to be able to shift the baseline corrected
spectra to 1 (100%) for transmission data.

stuart-cls:
It's semantics really, but to me shifting the baseline of the spectrum is a kind of baseline correction. Anyway, not a big deal.

PR #147, opened 2018-01-24
Preprocess widget: add EMSC

markotoplak:
I added simple EMSC to the "Preprocess Spectra" widget.

To do EMSC, we need a reference spectrum. Therefore I needed to add "Reference" input to the "Preprocess Spectra".

Are there any other preprocessors that would need something similar? I remember @clsandt once suggesting subtraction that would need something similar. Also, @stuart-cls made some preprocessors that use reference (Transmittance and Absorbance, but these were not passed from the widget).

If we add the "Reference" input (the alternative is to make a new widget that would be essentially the same but with different inputs), which name for the input would make sense? @borondics ?

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -824,3 +824,7 @@ def getx(data):
     except:
         pass
     return x
+
+
+def spectra_mean(X):
+    return np.nanmean(X, axis=0, dtype=np.float64)
--- a/orangecontrib/spectroscopy/preprocess/emsc.py
+++ b/orangecontrib/spectroscopy/preprocess/emsc.py
@@ -38,11 +38,12 @@ class _EMSC(CommonDomainOrderUnknowns):
             M.append(wavenumbersSquared)
         if self.use_b:
             M.append(ref_X)
-        M = np.vstack(M).T
+        M = np.vstack(M).T if M else None  # edge case: no parameters selected
 
         newspectra = np.zeros(X.shape)
         for i, rawspectrum in enumerate(X):
-            m = np.linalg.lstsq(M, rawspectrum)[0]
+            if M is not None:
+                m = np.linalg.lstsq(M, rawspectrum)[0]
             corrected = rawspectrum
             n = 0
             if self.use_a:
--- a/orangecontrib/spectroscopy/preprocess/emsc.py
+++ b/orangecontrib/spectroscopy/preprocess/emsc.py
@@ -2,7 +2,7 @@ import Orange
 import numpy as np
 from Orange.preprocess.preprocess import Preprocess
 
-from orangecontrib.spectroscopy.data import getx
+from orangecontrib.spectroscopy.data import getx, spectra_mean
 from orangecontrib.spectroscopy.preprocess.utils import SelectColumn, CommonDomainOrderUnknowns, interp1d_with_unknowns_numpy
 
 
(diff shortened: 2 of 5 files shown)


stuart-cls:
This is basically the problem I was trying to solve with the spectral calculator in #76 . At the time, we had discussed moving calculations that required more than one input into a separate widget, but I'm not stuck on that idea. 

The main concern I can see is it might be confusing to show/understand which preprocessor that takes a reference (Absorbance/Transmittance/Subtraction/EMSC/etc) should use which reference input. For example, if a user wanted to do a subtraction and an EMSC. Although maybe the answer is "don't do that" :)

**To answer your actual question**, yes: Absorbance, Transmittance (and the Subtraction one in #76 ) should make use of the reference input.

"Reference" would be my preferred terminology.

Thanks!

clsandt:
Hello
It seems to me that Orange may already have what you need to manage this reference problem.
Some widgets can output selected/non-selected data or training/test datasets. It should be possible to use this widget configuration.  Would it be possible/suitable to have a widget that input the whole dataset and output one channel for the reference spectrum and another channel for the rest of the spectra that need to be preprocessed? Or that the EMSC/preprocess widget use the ‘Data’ widget and import all the data by default, regardless whether they are selected otr not but treat the selected data as the reference and the unselected data as the data to be corrected?

Reg

C

De : Stuart Read [mailto:notifications@github.com]
Envoyé : mercredi 24 janvier 2018 16:16
À : markotoplak/orange-infrared
Cc : SANDT Christophe; Mention
Objet : Re: [markotoplak/orange-infrared] Preprocess widget: add EMSC (#147)


This is basically the problem I was trying to solve with the spectral calculator in #76<https://github.com/markotoplak/orange-infrared/pull/76> . At the time, we had discussed moving calculations that required more than one input into a separate widget, but I'm not stuck on that idea.

The main concern I can see is it might be confusing to show/understand which preprocessor that takes a reference (Absorbance/Transmittance/Subtraction/EMSC/etc) should use which reference input. For example, if a user wanted to do a subtraction and an EMSC. Although maybe the answer is "don't do that" :)

To answer your actual question, yes: Absorbance, Transmittance (and the Subtraction one in #76<https://github.com/markotoplak/orange-infrared/pull/76> ) should make use of the reference input.

"Reference" would be my preferred terminology.

Thanks!

stuart-cls:
I found a bug: if you de-select all terms the emsc preprocessor crashes since the `M` list (line 34-43) is empty when you try to `np.vstack`.

stuart-cls:
I like the preview of the reference curve, just a couple comments:
 - It would be nice to tell the user that if they select the "Scaling" term without a reference input, the software will automatically use the mean of the main dataset.
 - Relatedly, I think we should always show the reference being used, even if it is just the mean 
 - The reference curve shows up only when pressing "Play" (as with most preprocessors) but then remains when "Play" is deselected.
 - The reference curve is still shown when "Scaling" is not selected, not sure if that matters but could be confusing

---

For the Absorption, etc we rarely (never?) would want the reference to be a mean of the entire data set. Rather, you would either want a single reference applied to the entire dataset, or possibly a dataset of the same size broadcast across the array. 

I think it would be best to let this PR settle and get merged, and then I can update the other preprocessors to use the new reference input in a separate PR.

markotoplak:
@stuart-cls, thank you for your review

I disabled automatically taking mean of data as a reference, because as you say, it is confusing. I also added some short reference info to the preprocessor settings. Regarding reference and scaling, reference is always used in EMSC to obtain coefficients.

I opened #150 for non-disappearing reference.

stuart-cls:
Looks great

PR #166, opened 2018-04-10
[ENH] Hyperspectra: precise input of color limits

markotoplak:
Text fields instead of sliders for selecting color limits. If an user did not specify anything, infer limits from data.

What do you think, @stuart-cls? Should we even keep the sliders?


stuart-cls:
This is great! I started down this road myself but got stuck on the levels vs threshold question: your solution of just showing both and treating them differently wrt `schema_only` is good.

I like how the values are shown but on first use I was surprised that the values in the text boxes didn't update with the sliders. Of course this makes sense when you look at the code...

I think we should keep the sliders for doing general data exploration. The general workflow I can see happening is:

 * View new map, play with sliders to visualize best 
 * Find / otherwise determine appropriate levels for a project/experiment
 * Set levels manually (therefore with threshold sliders at 0, 1)

The confusion I foresee happening is the mixing of the threshold Settings (sliders) and the limits Settings(schema_only) (text boxes). Especially since threshold will carry over from other analyses.

This is where I got stuck yesterday: how to have linked sliders/text boxes that both set limits (not threshold) which doesn't break when loading in a new dataset with different min()/max() values.

One last question, why did you use Decimal instead of float()? Some datasets have very small/large numbers (images based on normalized peak areas or 2nd derivative peak heights, for example).

markotoplak:
I used Decimal for number input, because it can represent all floats exactly (I do not lose any precision) and offers me better control of what is shown when converted to string (I would not get possible spurious additions of something .00000000132132 when converting to string, as conversion from string -> float can be inexact).

stuart-cls:
Re: Decimal, very interesting. I misunderstood the function of the Decimal type, it seems very useful.

I like the latest version. The bonus dynamic range on the colours is really nice for images where the relevant variation is small!

I wonder if we should reset the threshold sliders when the user inputs a limit. That way if I input 50 as my lower limit, I can be sure the lowest colour == 50.

PR #167, opened 2018-04-18
Speedup image reading

markotoplak:
Removed explicit loops from _spectra_from_image.

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -115,34 +115,24 @@ def _spectra_from_image(X, features, x_locs, y_locs):
     Create a spectral format (returned by SpectralFileFormat.read_spectra)
     from 3D image organized [ rows, columns, wavelengths ]
     """
-    spectra = np.zeros((X.shape[0]*X.shape[1], X.shape[2]), dtype=np.float32)
-    metadata = []
+    X = np.asarray(X)
+    x_locs = np.asarray(x_locs)
+    y_locs = np.asarray(y_locs)
 
-    cs = 0
-    for ir, row in enumerate(X):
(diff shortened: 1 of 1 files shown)


stuart-cls:
Very nice! I measured a 20-40% speedup for small-medium images using this code, which is very good considering reading from disk into memory dominates the time.

Is there still a way for these "new-style" readers to include additional metadata beyond `x_locs` and `y_locs`? I remember I had done it once but I can't find that code.

markotoplak:
I updated pycharm today and was trying out its profiling feature - and noticed some unexpectedly slow behavior.

I would try to keep this function so that it just constructs a data table with x_locs and y_locs and outputs a spectra array. It is how its usually used.

Afterwards, the outputs are used to construct a table that actually contains the spectral data. Before that (or after, it does not matter much, the speed difference should be negligible), additional metas could be added with something like:

```
newmetacolumn = StringVariable("blabla")
metas = data.domain.metas + (newmetacolumn,)
newdomain = Domain(data.domain.attributes, data.domain.class_vars, metas)
newtable = data.transform(domain)
newtable[:, newmetacolumn] = values_for_that_meta_feature
```

The above code makes "space" in the domain and then fills it with some values (last line).

markotoplak:
And interestingly, loops themselves were not the slowest part in the old code: these images do not have that many pixels anyway. What was slow was copying of spectra from the array `X` to the array `spectra` (one copy operation for each pixel).

stuart-cls:
Ah, right, the final Table is constructed by `SpectralFileFormat.read` . Thanks for the code.

I agree 💯 to keep this function as simple as possible.

Interesting indeed! I think we do pretty well to avoid unnecessary copy() but it is definitely something to keep in mind.

PR #182, opened 2018-05-23
[ENH] EMSC supports bad spectra input

JohanneSolheim:
Added functionality for bad spectra input in EMSC.


markotoplak:
@sumrtime, could you also check whether this changes are OK?

PR #185, opened 2018-05-31
Remove workaround for memory use in pre-3.11 Orange

markotoplak:


markotoplak:
Remove workaround intruduced in #157

PR #189, opened 2018-06-11
Pylint fixes

stuart-cls:
gitignore: ignore .pyline_cache generated by check_pylint_diff 

check_pylint_diff: handle system pylint installed as "pylint3"

affects at least Debian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ __pycache__
 
 # Coverage reports
 htmlcov/*
+
+# check_pylint_diff
+.pylint_cache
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks! It works for me as it uses either pylint and pylint3 commands.

It is just interesting that no one from Orange team was complaining about this before. It seems that we are all working in a virtual python environments or with python3 as default system python. :)

stuart-cls:
Yes, I think Debian is special for keeping the "3" on things. I was using site-packages as much as possible in my venv which is why my pylint was called `pylint3`. Anyway, glad the `pylint`/`pylint3` worked as designed.

PR #192, opened 2018-06-21
A utility that can relate two settings: gui.connect_settings

markotoplak:
Someone programming a preprocessor requested a way to have movable lines that are relative to some value. Here is a short utility to do it with an example.


stuart-cls:
This is very nice!

Issue #149, opened 2018-01-25
Update spectral transformation preprocessors


stuart-cls:
So I don't forget, once #147 (add EMSC / reference input to preprocessors) is done, the Absorbance/Transmittance/etc spectral transformation preprocessors should be updated to 

- [ ] make use of the new reference input
- [ ] simplify the GUI presentation of these preprocessors (like #128 but without changing the underlying preprocessor code)

PR #198, opened 2018-06-29
[ENH] Unify spectral transforms preprocessor editor

stuart-cls:
Here is my attempt to put #76 #149 #128 to bed. I used Feri's GUI from #128 as a starting point.

Outstanding:
- [x] @markotoplak I think I did the settings handling correctly, but I'd like a second set of eyes
- [x] `CommonDomainRef` (427c79d) ~~doesn't seem to work with Auto Interpolate the way I would expect (see figure below)~~ I was using Auto Interpolate wrong again, it goes on the target dataset
- [x] I'm also not sure how to properly test the reference domain conversion functionality
- [x] I was surprised that the error fixed in c772c61 didn't trigger a test failure (it does in 46633f1)
- [ ] many/many reference crashes in preview window since data[preview] != ref anymore
- [ ] no user feedback when they do something that makes no sense (currently just returns an empty Table) 
- [x] 46633f1 adds reference domain conversion tests, but the domain-equality assertion fails (commented out as I don't understand why)

Non-goals:
 - Handle references which are a subset of the data (like a reference FPA tile)

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_preprocess.py
+++ b/orangecontrib/spectroscopy/tests/test_preprocess.py
@@ -81,6 +81,8 @@ def add_different_reference(class_, reference_arg_name, reference, *args, **kwar
 
 PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(EMSC, "reference", SMALL_COLLAGEN[0:1]))
 
+for p in [Absorbance, Transmittance]:
+    PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(p, "ref", SMALL_COLLAGEN[0:1]))
 
 # Preprocessors that use groups of input samples to infer
 # internal parameters.
--- a/orangecontrib/spectroscopy/tests/test_preprocess.py
+++ b/orangecontrib/spectroscopy/tests/test_preprocess.py
@@ -42,7 +42,7 @@ PREPROCESSORS_INDEPENDENT_SAMPLES = [
 def make_edges_nan(data):
     s = data.copy()
     s[:, 0:3] = np.nan
-    s[:, s.X.shape[1]-1] = np.nan
+    s[:, s.X.shape[1]-3:] = np.nan
     return s
 
 
@@ -73,6 +73,7 @@ def add_different_reference(class_, reference_arg_name, reference, *args, **kwar
     modified = [reference,
                 shuffle_attr(reference),
                 make_edges_nan(reference),
+                shuffle_attr(make_edges_nan(reference)),
                 make_middle_nan(reference)]
     for d in modified:
         kwargs[reference_arg_name] = d
--- a/orangecontrib/spectroscopy/tests/test_preprocess.py
+++ b/orangecontrib/spectroscopy/tests/test_preprocess.py
@@ -82,7 +82,10 @@ def add_different_reference(class_, reference_arg_name, reference, *args, **kwar
 PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(EMSC, "reference", SMALL_COLLAGEN[0:1]))
 
 for p in [Absorbance, Transmittance]:
+    # single reference
     PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(p, "ref", SMALL_COLLAGEN[0:1]))
+    # multi reference (many:many)
+    PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(p, "ref", SMALL_COLLAGEN))
 
 # Preprocessors that use groups of input samples to infer
 # internal parameters.
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -12,6 +12,7 @@ from orangecontrib.spectroscopy.data import getx
 
 from orangecontrib.spectroscopy.preprocess.integrate import Integrate
 from orangecontrib.spectroscopy.preprocess.emsc import EMSC
+from orangecontrib.spectroscopy.preprocess.transform import Absorbance, Transmittance
 from orangecontrib.spectroscopy.preprocess.utils import SelectColumn, CommonDomain, CommonDomainOrder, \
     CommonDomainOrderUnknowns, nan_extend_edges_and_interpolate, remove_whole_nan_ys, interp1d_with_unknowns_numpy, \
     interp1d_with_unknowns_scipy, interp1d_wo_unknowns_scipy, edge_baseline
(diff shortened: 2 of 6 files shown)


stuart-cls:
PS Happy Canada Day 🍁 🇨🇦

stuart-cls:
I made a test which catches the many/many preview problem described above. However, I'm not sure there is a good solution: the `createinstance()` method of the `SpectralTransformEditor` has no way of knowing whether the instance will be used for preview data or the full data set, and so there's no way to use the appropriate reference subset.

orangecontrib/spectroscopy/tests/test_preprocess.py
```diff
@@ -127,6 +156,70 @@ def test_roundtrip(self):
...
+        """Test domain transformation of reference input"""
+        data = Absorbance(ref=RSC_JUICE_SMALL)(SSC_JUICE_AUTO)
+        np.testing.assert_equal(getx(data), getx(SSC_JUICE))
+        # self.assertEqual(data.domain, SSC_JUICE_AUTO.domain)
+        iref = RSC_JUICE_SMALL.from_table(SSC_JUICE_AUTO.domain, RSC_JUICE_SMALL)
+        nt = Absorbance(ref=iref)(SSC_JUICE_AUTO)
+        np.testing.assert_allclose(data.X, nt.X)
+        # self.assertEqual(data.domain, nt.domain)
+
+    def test_multi_ref(self):
```
markotoplak: What exactly did you want to test with these tests?

markotoplak:
Thank you for this pull request! Sorry that I took so long to go through it.

I think you handled settings transition correctly and you even made tests! Also, a good call for opusFC independence.

I do have some questions though:

- How does the subtraction fit with the other two transforms that change spectra types? I am probably missing some context, because to me it (naively) seems as something that would warrant its own preprocessor.

- Why did you make individual reference transformation tests instead of using something similar to `PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(EMSC, "reference", SMALL_COLLAGEN[0:1]))` line and then rely on general preprocessor test to catch incorrect matching  (of course you would have to fake a collagen reference somehow)? I am not saying it can be done though. :)

- Could you explain what did you want to do with different kind of references (many:many, many:1, many:subset reference)? Do I understand correctly, that a possible use cases would be that you could either have a reference of a single row

Unfortunately, I do not know yet how to produce error messages nicely. For EMSC preprocessor I was just showing a text in a label...

The domains (in reference domain tests) have the same labels, but they are not the same. The transformed one can also transform some separate testing data so it has additional data on the transformation. So, if I add these lines to your tests:

     print(repr(data.domain[0]))
     print(repr(SSC_JUICE_AUTO.domain[0]))

Then they print

    ContinuousVariable(name='4000.1161', number_of_decimals=3, compute_value=<orangecontrib.spectroscopy.preprocess.transform.AbsorbanceFeature object at 0x7f0e935dee80>)
    ContinuousVariable(name='4000.1161', number_of_decimals=3, compute_value=<orangecontrib.spectroscopy.preprocess.InterpolatedFeature object at 0x7f0e92f22400>)

That's why we struggle with transformations through making new features that have `compute_value` defined. See https://docs.orange.biolab.si/3/data-mining-library/reference/data.variable.html#derived-variables

Hmmm, I think I see a problem with multi references on test data (if I understand them correctly).

stuart-cls:
> How does the subtraction fit with the other two transforms that change spectra types? I am probably missing some context, because to me it (naively) seems as something that would warrant its own preprocessor.

It doesn't, really. I was looking for something simple to test using the reference input mechanism, we need a subtraction calculator, and I was convinced by Feri's argument to limit the number of preprocessor editors (GUI). I would be fine breaking it out graphically. Or did you mean logically separate as well?

> Why did you make individual reference transformation tests instead of using something similar to PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_different_reference(EMSC, "reference", SMALL_COLLAGEN[0:1])) line and then rely on general preprocessor test to catch incorrect matching (of course you would have to fake a collagen reference somehow)? I am not saying it can be done though. :)

I think the preprocessors were passing the general test but not working for the desired interpolation functionality, so I was trying to test interpolating incompatible references and domain transform in one go. I'm not fixed on this approach though, the tests could definitely use some simplification.

So to be clear about what I was really trying to test: the dataset I was playing with had a reference with a slightly larger energy window than the sample, and I want to test the case where you put the sample set (with auto-interpolated capability) on the input and the non-matching reference on the reference.

> Could you explain what did you want to do with different kind of references (many:many, many:1, many:subset reference)? Do I understand correctly, that a possible use cases would be that you could either have a reference of a single row

So:
- for the **many:1** case, you have a set of sample single beams (say from a pt-by-pt map) and a single reference single beam and you want to calculate absorbance or whatever
- for the **many:many** case you have a set of sample single beams (say from a set of measurements) and a equal-sized and ordered set of reference single beams and you want to calculate the absorbance for each as appropriate. This would also be the case for a single-tile FPA sample/reference single beam.
- for the **many:subset** case, the case I'm thinking of is a mosaic FPA sample single beams and you want to reference to a single-tile reference FPA set. This is a bit more tricky so I decided not to handle this yet (see Non-goals).

Your tip about using `repr()` to get the `compute_value` of the domain is really helpful!

markotoplak:
Thank you for your work on this pull request.

As you point out, multi-reference is problematic because it does not fit into the preprocessing pipeline (it can not preprocess new data). For now, I can only think of this:

- in the preprocessing widget, only allow a single reference (either a single spectrum or an average of multiple input spectra)
- have a separate widget that can also handle multi-reference

Did you perhaps think of anything better?

stuart-cls:
Yes, I think the same thing. Especially in the context of other simple spectral calculations (subtraction, etc) that don't really fit into the preprocessor model. So my plan going forward is to clean up this PR to do the simple thing, and then resurrect the spectral calculator widget idea to cover cases that don't make sense as preprocessors.

stuart-cls:
Okay, I handled multi-spectral tables on reference by simply using the first spectrum in the table and informing the user.

There are still two test failures I mentioned above, any comments? Otherwise ready to go.

stuart-cls:
>     * `unknown_no_propagate` test fails because it expects that the preprocessor doesn't introduce NaNs. But in the case where we've applied `make_edges_nan()` and `make_middle_nan()` to the reference using `add_different_reference(), the reference has NaNs that the sample doesn't and the result reflects this. It seems to me that it is more accurate to admit the calculated result is unknown if _either_ the sample or the reference is unknown for a given feature.
> 
>     * `slightly_different_domain` tests fails for the `make_edges_nan(reference)` case when the test data has been shifted by 1 wavenumber and interpolated. I'm not sure why.

@markotoplak, anyone else,
Any thoughts on these two test failures? For the first one, how do you handle this in EMSC?

Otherwise I'm done with this (much-reduced scope) PR. I can rebase it when ready if you'd like.

markotoplak:
@stuart-cls, thank you, I'll try to handle the tests.

markotoplak:
Stuart, I added a commit that changed how reference data is used: it interpolates it to the other data before using it (and somewhat sensibly interpolates possible missing values). This makes it pass the tests. Thank you for your contribution!

stuart-cls:
Thanks Marko. I was hesitant to silently interpolate but I guess it is the best option.

markotoplak:
You are right, a warning should come up to the widget somehow... If you remember a good way to do it, I am all for it.

PR #200, opened 2018-08-15
Changed the text on the Menu button from View to Menu in Spectra and …

borondics:
…Hyperspectra


borondics:
Seems like something is wrong with the Travis configurations. 
Questions: 
- why is Orange set to 3.11
- why don't we test for Python 3.6 and 3.7 as well?

markotoplak:
The tests are OK. The minimum version of Orange we support is 3.11.

When we drop Python 3.4, we will test for newer Pythons also. It is just that we have to wait more if we test more...

PR #202, opened 2018-08-15
Changed initialization of the limit lines in Preprocess Spectra/Cut

borondics:
…tra/Cut

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpreprocess.py
+++ b/orangecontrib/spectroscopy/widgets/owpreprocess.py
@@ -346,11 +346,16 @@ class CutEditor(BaseEditorOrange):
     def set_preview_data(self, data):
         x = getx(data)
         if len(x):
-            self._lowlime.set_default(min(x))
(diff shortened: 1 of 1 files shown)


stuart-cls:
Hi Feri!

Can you explain why we want this and how it changes the behaviour? 

Depending on your answer, I wonder if there are other places we want this logic and if we could consolidate it in a function.

borondics:
Yes, previously the lines went to the middle of the range on top of one
another. This way the initialization is clear and (I think) conveniently
spaced at 10% range far from the data edges.

stuart-cls:
Oh I see, your issue is just in the case of inverted Cut. I like the 10% full range thing!

borondics:
I changed both cuts...

stuart-cls:
Sure, but normal Cut() already set the limits to min()/max()

borondics:
Something happened with the print precision of the Low limit in the cut preprocessor at initialization...

image

PR #206, opened 2018-08-27
[ENH] XAS normalization, EXAFS extraction

oroue:
A prototype version of XAS.

Do not merge yet.


markotoplak:
This passes basic tests. I've also linted the code a bit so that lint does not complain too much.

While there are still things to be improved with the exception handling, I think this can be merged.

PR #208, opened 2018-08-30
owpreprocess: reworked Savitzky-Golay editor

markotoplak:
The settings entered are now always valid and are executed as they are.
The preview is shown directly after editing any of the settings.


stuart-cls:
Very nice! This has been bugging me for a while too :)

One comment, since we are doing a widget replacement and settings upgrade anyway, is this a good time to rename from `orangecontrib.infrared.savitzkygolay` to `orangecontrib.spectroscopy.savitzkygolay`? Obviously not at all important.

PR #211, opened 2018-09-04
Movable line labels that overlap less

markotoplak:


borondics:
Looks better! 
However:
- they overlap with each other (this is not a major issue, because we rarely put them so close)
- the Low-High label doesn't swap if they become the opposite...

PR #212, opened 2018-09-04
Preprocess Spectra/Cut: round initial values

markotoplak:
Trying to solve a problem @borondics mentioned in #202.


borondics:
Thanks!

PR #219, opened 2018-09-27
Added averaging widget

lsammut:
Created widget to average spectra

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owaverage.py
+++ b/orangecontrib/spectroscopy/tests/test_owaverage.py
@@ -1,0 +1,45 @@
+import numpy as np
+import Orange
+from Orange.widgets.tests.base import WidgetTest
+from orangecontrib.spectroscopy.widgets.owaverage import OWAverage
+
+
+class TestOWReshape(WidgetTest):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.collagen = Orange.data.Table("collagen.csv")
+
+    def setUp(self):
+        self.widget = self.create_widget(OWAverage)
+
+    def test_load_unload(self):
+        # just to load the widget (it has no inputs)
+        pass
+
+    def test_no_data_warning(self):
+        self.assertTrue(self.widget.Warning.nodata.is_shown())
+        self.send_signal("Data", self.collagen)
+        self.assertFalse(self.widget.Warning.nodata.is_shown())
+
+    def test_average(self):
+        self.send_signal("Data", self.collagen)
+        out = self.get_output("Averages")
+        self.assertTrue(out.X.shape[0] == 1)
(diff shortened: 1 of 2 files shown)


borondics:
@lsammut this is a great start! Do you think you could add functionality to check for classes and provide class averages as an option? Many users are requesting this. The collagen dataset should be a good test.

I am not sure why the checks fail @markotoplak?

markotoplak:
@borondics, it was a library problem, now solved in #220

@lsammut, thank you. As you see, the linter complains. :)

stuart-cls:
Great! This is a very popular request but I've gotten stuck on the feedback of my implementation a couple of times. 

In https://github.com/Quasars/orange-spectroscopy/pull/66 you can see my first attempt, but it doesn't work as a preprocessor. However you will find useful code in there to add averaging by a discrete variable.

One challenge is how to handle meta-variables. The code I've been providing to our users currently looks like this: https://gist.github.com/stuart-cls/59cb17e1833d41e9e13f8c3f8751ff65

The main problem (as I now understand it) with these approaches is that running it as a preprocessor was a bad idea.

Hope that helps!

stuart-cls:
@markotoplak I should also mention that I started adding a statistics output to BoxPlot as you suggested. I have a prototype but it seems overly complicated and fragile due to the internal data structures of the BoxPlot widget. I can send it to you if you'd like.

borondics:
Could we keep track of things people start somehow better? It already
happened to me that I started something and someone suddenly pushed the
same thing. I am no Github guru, so please come up with something. I see
one can assign tasks, etc...

stuart-cls:
What I try to do is have an issue or WIP PR to track what I'm working on. It's true the averaging calculation sat untouched for a long time waiting for a way forward.

See also https://github.com/Quasars/orange-spectroscopy/blob/master/CONTRIBUTING.md#communication

markotoplak:
@stuart-cls, thank you for your averaging code. 

Your code works preprocessor-like: a new domain is build that knows how to transform and then it is applied to the table. That certainly works, but do we need it? What is the possible use case? I could not think of any (you can not build classifiers on averaged data anyway). Have you or your users found an use case where you needed it?

stuart-cls:
@markotoplak No we don't need to do it that way, it was a dead end as you point out. My understanding of preprocessors is better than when I wrote that and I 100% agree the approach here is better.

@lsammut Thanks for this code, I'm sorry for derailing your nice PR :) 

I'll close the other one to remove any confusion.

lsammut:
Hi all, I hope this widget is useful, at least as a starting point.
@markotoplak Thank you for taking the time to work through this with me.
@borondics I am new to orange, but if I have some time I can look into including the functionality you mention. 
@stuart-cls Thank you for your comments and suggestions. Perhaps, some of the functionality in your code may be useful for this widget.

PR #221, opened 2018-10-04
Change spectra layout

borondics:
The saving functionality was changing the graph layout before the save and changing it back afterward. I took this out and made the top and right axes permanent with no values displayed next to the ticks. I also added some margins (we could make this dynamic).


markotoplak:
Thanks! I removed your accidental commits. I also removed the line that set the margins, because it could be problematic with longer axis values.

borondics:
Thanks for the merge. We will need some kind of margins though because the
last number gets cut off in the bottom right corner. There is probably a
better way of doing it than what I tried to hard-code.

Issue #190, opened 2018-06-21
Dependency on a binary package: opusFC


yurivict:
https://github.com/Quasars/orange-spectroscopy/blob/master/setup.py#L134

PyPI website doesn't provide the source tarball https://pypi.org/project/opusFC/#files
There is no meaningful upstream URL there either.
So the package can't be built, and ```orange-spectroscopy``` can't be installed through the package.

scott-cls:
Due to sensitive information in the source code for this project, a source tarball WILL NOT be provided.

We will look into the problem you are having.  Thanks.

stuart-cls:
Let's continue this in the opusFC-dist issue. Thanks for your report.

markotoplak:
Reopening, as I think this issue is valid. I believe it is wrong that we depend on a binary package (a user can have a platform where there is no binary for or prefers not to use binaries altogether).

I propose removing opusFC from dependencies. The interested users could either:
1. Install opusFC themselves, perhaps the easiest option would be to use Orange's add-ons dialog.
2. When we release it (plan: August), use a pre-packaged version of the Orange + Orange Spectroscopy + opusFC + whatever we want. 

To support (1) we could rewrite file readers so that they would still offer to open Opus files, but if a user tried to open such file, we could ask him to install the opusFC library.

I do not think removing a dependency is a problem, as most of less computer savvy users will probably use (2) anyway.

yurivict:
There should be the open source alternative for such a common task as reading some file format.
The FreeBSD ports can't use binary python packages.

stuart-cls:
@markotoplak This is fine with me, as long as we can provide non-expert users with a simple way to install. 

The big question, who will do this?

PS this is basically a duplicate of #51

yurivict:
> Had we provided the source code the format itself will be exposed.

The veracity of this statements depends on the way how the file format is parsed.
Imagine a flex/bison parser of some file format. Providing the generated parsing tables doesn't actually expose the format.

-----

If in addition to just installing a package some custom extra-actions would be required, like installing some custom binary package, accepting a license, etc, this would reduce the motivation to use it by a lot.

borondics:
Thanks for the comment! Guys, what do you think?

PR #225, opened 2018-10-28
Remove opusFC from requirements

markotoplak:
Remove a binary-only module from the requirements. Opening an Opus file produces an error that instructs the user which library to install.

Fixes gh-190. Merge when Quasar packages are available.


markotoplak:
Yes, I agree that the error is a bit hidden. Hopefully all users will begin to use Quasar soon and it won't be a problem for them.

Should we open a warning dialog? But the problem here is that the code does not know whether is called from the widget or "just" from a script. I do not want to add additional checks into the File widget. Do you perhaps see a solution?

stuart-cls:
No, you're right it's better to use the existing Error infrastructure. Learning to look for those little error boxes is part of the fun of learning Orange ;)

PR #230, opened 2018-11-07
Add datasets/ to MANIFEST.in

markotoplak:
Need to mention datasets manually in MANIFEST.in if I build the package on my Macbook with setuptools 38.4.0. 

I do not know why we need it there though. If I now rebuild old versions (0.3.6) that are fine on PyPi, the resulting package is missing the datasets.


stuart-cls:
This might help:
https://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distribute

although it's quite old it's been updated somewhat. Seems to be related to some changes in handling the `include_package_data=True` setting. See the second and third answers.

Note the comment that `MANIFEST.in` only works for sdist, but not bdist. That might cause problems with bdist_conda.

Issue #231, opened 2018-11-07
Average Spectra: group by variable


stuart-cls:
See #219 for context. 

@lsammut, I'm planning on working on this unless you still want to?

lsammut:
> See #219 for context.
> 
> @lsammut, I'm planning on working on this unless you still want to?

Hi @stuart-cls, I'm happy for you continue working on this. Unfortunately I'm not sure how much time I'll be able to contribute towards it in the next few months.

PR #233, opened 2018-11-08
Average by group

stuart-cls:
Implements #231 

- [x] Add a few more other_var methods and matching GUI/tests
- [x] documentation :)


orangecontrib/spectroscopy/widgets/owaverage.py
```diff
@@ -67,10 +67,35 @@ def set_data(self, dataset):
         self.commit()
 
     @staticmethod
-    def average_table(table):
+    def average_table(table, other_vars="None"):
```
markotoplak: Wouldn't it be enough to just do "Mean" and not complicate with options. This one keeps the most information. What do you think?

stuart-cls: Yes, great suggestion. It's easy to get carried away...

stuart-cls:
I think this is ready to go. Two potential gotchas:

1. #234 might be a problem with the output OWAverage is generating, although I don't see how.
2. I couldn't quite get the documentation to work (when I press F1). Maybe we can add a small blurb somewhere or at least  a link to how Orange does it?

markotoplak:
Thank you, I like what you did and how you did it. 

The bug you found is independent of this PR: if I save and open the averaged data I still get the error.

You need to build the documentation to see it:

    cd doc
    make htmlhelp

PR #245, opened 2018-11-20
owpreprocess: highlight pairs of spectra 

markotoplak:
In the previews, highlight both the original and the processed sprectrum together.

I was showing effects of preprocessing to students but it was hard to see which spectra belonged together.


stuart-cls:
This is very nice, thank you.

PR #246, opened 2018-11-20
travis: use Python 3.6 and do not test on PyQt4 anymore

### LINKED ISSUE #227: Python 3.4 support
https://github.com/biolab/orange3/issues/3341 and https://github.com/biolab/orange3/pull/3343 is exciting, 

I think we should follow suit and drop at least 3.4, especially before releasing Quasar.

@markotoplak thoughts?

markotoplak:
Fixes #227


markotoplak:
There is a failing test because of changes in biolab/orange3#3403. Will fix that in another PR.

PR #247, opened 2018-11-21
 Fix PCADenoising to work with refactored Orange projection 

markotoplak:
Fix the reason #246 is still failing


stuart-cls:
Hi @markotoplak , I guess this is a backwards-compatible refactor?

markotoplak:
Yes. Otherwise the older tests would not pass. We intentionally test on older Orange releases too.

PR #248, opened 2018-11-21
Remove automatic interpolation with some readers

markotoplak:
It was inconsistent to automatically interpolate only two formats
Also, concatenating such files was impossible,
which led to problems if multiple of these files were read.

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -57,8 +57,7 @@ class DatReader(FileFormat):
     def read(self):
         tbl = np.loadtxt(self.filename, ndmin=2)
         domvals = tbl.T[0]  # first column is attribute name
-        from orangecontrib.spectroscopy.preprocess import features_with_interpolation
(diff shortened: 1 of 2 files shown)


stuart-cls:
What are the practical consequences of this? Files with different energy spacings/ranges must be opened in different file widgets?

I also wonder what the remaining `TestReaders.test_autointerpolate` is really for if we no longer autointerpolate on file read.

stuart-cls:
I forgot to say 💯 for improving our consistency :)

markotoplak:
Before, if you had one of this auto-interpolate files as training data, the testing data could be whatever (other wavelengths) and they would automatically work. But then everything that reads spectra should have it, which is in practice impossible, because spectra can also be the ordinary Orange files. And introduces some problems with widgets such as Concatenate.

Because a user can achieve the same effect manually I just removed it.

I left that test in because it still tests a certain helper function (which could perhaps also be removed).

stuart-cls:
Okay, I see. I'm happy to see the silent interpolation removed in any case. Thanks!

Issue #238, opened 2018-11-14
Average spectra: handling of unknown values of Group-by


markotoplak:
Currently, no average is computed for the unknown value of group-by feature. Could we perhaps also do averages of those?

Issue #237, opened 2018-11-13
AverageSpectra: crashes on StringVariable


stuart-cls:
Missed testing StringVariable (and possible TimeVariable) in #231 .

stuart-cls:
For reference, this is actually due to `numpy.lib.nanfunctions._replace_nan` just guessing and returning a boolean array mask for object arrays because object arrays do not support `isnan` ([numpy-gh-9009](https://github.com/numpy/numpy/issues/9009)).

The resulting error shows up in a ContinuousVariable stored in a meta object array, which we can fix by casting to np.float64 (as all ContinousVariables are).

PR #249, opened 2018-11-22
OWAverage: Fix average group handling of metas and unknowns

stuart-cls:
A couple of fixes to work around issues with numpy / Orange handling of object arrays (which the meta arrays are whenever a `StringVariable` is present).

Implemented handling of unknown group_var values and group_var values with no members.

Related #237, #238


stuart-cls:
Thanks to lint complaining I didn't actually test the output of the last test, I found (and fixed) a nasty bug where `average_table()` would set the first row of the _input_ dataset to Unknown when calculating a global average (since .metas was sometimes a view).

Should be good to go now.

markotoplak:
Wow, thank you. Kudos for finding the copying bug - that would certainly bite us later on.

PR #257, opened 2018-11-30
[FIX] AgilentReaders: use correct file extensions

### LINKED ISSUE #194: Agilent module: Don't search for seq/drd interferogram tiles on load.
As suggested by @alexhenderson in #117 , we don't need to have the `.seq` or `.drd` files which only contain interferogram data present to load the agilent processed (`.dat` / `.dmd`) data blocks.

stuart-cls:
Now the reader can load files without the interferogram tiles (*.seq / *.drd) being present (Fixes #194)

@markotoplak , can you check that I implemented the meta-reader for *.dat files properly?

Also is there anyway to hide the meta-reader in the File widget list? It's kind of ugly and distracting.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 # Ref https://docs.python.org/2/distutils/sourcedist.html#commands
 recursive-include orangecontrib *.ows icons/* datasets/*
 recursive-include orangecontrib/spectroscopy/datasets matlab/*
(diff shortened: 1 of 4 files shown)


markotoplak:
It is good that you figured out the meta reader. There is no way to hide it now, but yes, this shows that Orange could use some additional validation mechanism apart from the filename.

PR #259, opened 2018-12-11
SpectralReader: do not create intermediate tables

markotoplak:
This saves memory.


orangecontrib/spectroscopy/data.py
```diff
@@ -40,11 +40,12 @@ def read(self):
             domain = Orange.data.Domain(features, None)
             return Orange.data.Table(domain, data)
         else:
-            domain = Orange.data.Domain(features,
-                                        class_vars=additional_table.domain.class_vars,
-                                        metas=additional_table.domain.metas)
-            ret_data = additional_table.transform(domain)
-            ret_data.X = data
+
```
stuart-cls: This seems like an unnecessary line break?

PR #260, opened 2018-12-11
owpreprocess: make saved preprocessors schema only

markotoplak:
I think preprocessors should not fill from before when putting a new widget on canvas.


stuart-cls:
I agree, now that we have the new Transform widget it is easy to reuse existing preprocessor chains.

Issue #261, opened 2018-12-11
OWFFT: Support laser wavenumber and UPD metadata


stuart-cls:
Follows up on #243 support

PR #262, opened 2018-12-11
[ENH] Agilent read interferogram data (seq and drd)

### LINKED ISSUE #243: Read .seq Agilent file
We need to read .seq files to do FFT

stuart-cls:
Add support for reading agilent interferogram data files, including metadata required to calculate FFT properly (See #261)

Closes #243

### DIFF
--- a/orangecontrib/spectroscopy/agilent.py
+++ b/orangecontrib/spectroscopy/agilent.py
@@ -1,4 +1,4 @@
-__version__ = "0.1.2"
+__version__ = "0.3.0"
 from pathlib import Path
 import struct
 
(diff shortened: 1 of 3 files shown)


orangecontrib/spectroscopy/agilent.py
```diff
@@ -98,8 +98,9 @@ def _get_prop_d(dat, param):
         return val
 
     def _get_prop_str(dat, param):
```
markotoplak: I would have named this internal helpers differently: I would not use the underscore. Also, "get" can often seems a bit superfluous.

orangecontrib/spectroscopy/agilent.py
```diff
@@ -245,62 +353,186 @@ class agilentMosaic(DataObject):
     https://bitbucket.org/AlexHenderson/agilent-file-formats
     """
 
+    def __init__(self, filename, MAT=False):
+        super().__init__(filename, MAT)
+        self._get_data()
+
+    def _get_data(self):
```
markotoplak: For, the name _get_data was a bit confusing and was looking for what it actually returns. Perhaps a name such as _prepare_data or _load_data or similar would be more appropriate next time.

markotoplak:
I like that you are handling tiled images... Let's see where this gets to.

stuart-cls:
Thanks Marko, I agree the internal method naming is not the best in the agilent module. I'll make a note on that.

If you would like to try the tile-by-tile file loader, it is the only widget in the [prototypes](https://github.com/Quasars/orange-spectroscopy-prototypes) repo right now. There is an example .ows included showing how to use it. Some of our users have been putting it through its paces for a few months now without issue.

PR #267, opened 2018-12-13
owspectra: use less memory for computing averages

markotoplak:
Averages temporarily allocated lots of memory, both for selection and for the nanmean/nanstd. That computation was split column-wise into smaller chunks.


markotoplak:
Yes, it would be also good for OWAverage. 

But also many other widgets have similar problems. Like integrals (which freezes HyperSpectra), but those can be split line-wise...

Issue #252, opened 2018-11-22
FFT: calculation tests


stuart-cls:
Need to complete tests of actually FFT output.
I've started this but it was holding up #187 

 - Sanity check before averaging forward/backward spectra after FFT
 - Other

stuart-cls:
Closed in #282

PR #268, opened 2018-12-19
[WIP] Tests for irfft calcuations (agilent data)

stuart-cls:
Some irfft calculation tests (#252) which compare the results to the stored single channel and absorbance data in our existing test data. There is one new dataset to provide a background file to compare the entire IFG -> SC -> AB calculation.

Currently the single channels match nicely only if they are normalized: there is a scaling factor that I haven't been able to track down. Note that this has no effect on the final AB calculation as long as both sample and reference are calculated the same way (which they ought to be).

This (WIP) PR builds on #262 and #263.


stuart-cls:
Oops, didn't mean to close, and now I can't re-open on the rebased branch :P I'll open again.

PR #273, opened 2019-01-10
[FIX] Preprocessors in the Preprocess Spectra widget also preprocess the reference

markotoplak:
#### Issue

@achimkohler reported the following error with the next schema

image003

The upper result looks like: EMSC pre-process spectra are flipped: I think raw spectra are averaged and used as reference, while the averaged derivated spectra should be used

image004

The lower stream results in (output from Preprocess Spectra (2)): Her you can see that the EMSC preprocessed spectra are not flipped compared to the derivated spectra:

 image005

#### Fix

I think that preprocessing reference data with the same preprocessors as we process input data makes sense. If someone wants old behavior, they can still have it by using 2 preprocess widget instead of one.

The older workflows should still work the old way but give a warning.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpreprocess.py
+++ b/orangecontrib/spectroscopy/widgets/owpreprocess.py
@@ -1648,7 +1648,7 @@ class SpectralPreprocess(OWWidget):
             self.curveplot_after.hide()
 
     def _initialize(self):
-        for i, pp_def in enumerate(self.PREPROCESSORS):
+        for pp_def in self.PREPROCESSORS:
             description = pp_def.description
             if description.icon:
                 icon = QIcon(description.icon)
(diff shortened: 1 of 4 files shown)


markotoplak:
@stuart-cls, if you have time, I would be grateful you take a look at this.

stuart-cls:
Sure, I can look at this tomorrow.

orangecontrib/spectroscopy/widgets/owpreprocess.py
```diff
@@ -1478,6 +1484,11 @@ class Outputs:
     class Error(OWWidget.Error):
         applying = Msg("Error applying preprocessors.")
 
+    class Warning(OWWidget.Warning):
+        reference_compat = Msg("Reference is not processed for compatibility with the loaded"
```
stuart-cls: Need a space at the end here

markotoplak:
@stuart-cls, thank you for the comments. I fixed all except (4). 

Reference should not be shown at all if the upper preview window was not the one you edit the settings for. But this is a separate problem.

PR #275, opened 2019-01-15
GPL3+ license text

markotoplak:


stuart-cls:
A+!

I would note that the license text you add says:

> GNU Affero General Public License as
published by the Free Software Foundation, either version 3 of the
License, or (at your option) any later version.

which is the AGPLv3+, not the GPLv3+. They are similar but not the same.

markotoplak:
Thanks, I was sloppy. I changed it to GPL. Which one do you prefer? I do not think it matters for us...

stuart-cls:
No preference for me.

PR #279, opened 2019-01-23
[FIX] agilent: Handle lower-cased .dmt filename

stuart-cls:
In common agilent datasets, the base filename may be mixed case, but the `.dmt` file is always lower-cased. When we switched to selecting `.dmt` files in #257 , I didn't account for this properly as my test machine is case-insensitive (Windows).


stuart-cls:
Tests now properly catch the bug (87b0a9f) and pass.

PR #280, opened 2019-01-25
[ENH] STXM stack aligner

markotoplak:
Moved from Quasars/orange-spectroscopy-prototypes#3


markotoplak:
For this to work we need another big dependency: scikit-image. It would increase installers by 25MB.

Also, @stuart-cls, what happens if you try to install this library with pip on Windows? Conda would probably work fine, I guess.

borondics:
Can we just take that function from scikit-image that is used?

markotoplak:
@borondics, I am quite surprised to say, that, yes, in this case we could actually take just that function (it is pure python, no other dependencies than numpy or scipy). Thanks for the suggestion!

markotoplak:
Stuart, thank you!

Issue #264, opened 2018-12-12
OWFFT: Keep metadata on transform


stuart-cls:
Most importantly, `map_x` and `map_y`

PR #281, opened 2019-01-28
[FIX] FFT widget: keep metas

markotoplak:
The meta data (X and Y image positions) were lost after FFT.


stuart-cls:
🙌  This is fantastic! I especially like merging `build_spec_table` and `SpectralFileFormat.read()`.

Thanks for fixing this. Fixes #264

Issue #265, opened 2018-12-12
OWFFT: zero-filling value not consistent between widget and library


stuart-cls:
Basically, `OWFFT.zff` is an _index of a combo box_ (0, 1, 2, 3) where the index is `2**n` of the value that should be sent to IRFFT. Currently this is sent directly as `IRFFT(zff=self.zff)` which is wrong.

@markotoplak The simple fix is to send `IRFFT(zff=2**self.zff)`, which is what I did so far as it requires no settings migration, etc. Alternative is to make `self.zff` equal to the actual zero-filling-factor, which is more logical and consistent but requires settings migration and GUI changes.

What do you recommend?

markotoplak:
I think the simple fix is fine, but we should document it in the widget. I will just add the comment and accept your PR.

PR #282, opened 2019-01-28
[NOMERGE][WIP] Tests for irfft calcuations (agilent data)

stuart-cls:
Some irfft calculation tests (#252) which compare the results to the stored single channel and absorbance data in our existing test data. There is one new dataset to provide a background file to compare the entire IFG -> SC -> AB calculation.

Currently the single channels match nicely only if they are normalized: there is a scaling factor that I haven't been able to track down. Note that this has no effect on the final AB calculation as long as both sample and reference are calculated the same way (which they ought to be).


stuart-cls:
Clones #268 which I closed by mistake...

stuart-cls:
Still contains a temporary fix for #265, so don't merge.

markotoplak:
@stuart-cls, were planning to add anything else to this PR? If not, feel free to merge.

stuart-cls:
It's not perfect but it's more useful than nothing!

Issue #285, opened 2019-02-04
owpreproc: No longer outputs PreprocessorList consistently


stuart-cls:
Since the 0.4.0 release, OWPreprocess no longer outputs a PreprocessorList consistently, which I relied on for the tile-by-tile loader.

See https://github.com/Quasars/orange-spectroscopy-prototypes/commit/c3879329b4a967b518ca2b45a18eb689e162aa4f for a test that works with 0.3.13 but not with 0.4.0. 

@markotoplak I can't quite follow everything this widget is doing and how it decides to run `apply()` or not?

markotoplak:
Sorry I caused you problems, but could you rather check for the Proprocess instead of the PreprocessorList? Like this: 

```
diff --git a/orangecontrib/protospec/tests/test_tile_reader.py b/orangecontrib/protospec/tests/test_tile_reader.py
index 444cf90..e489ac3 100644
--- a/orangecontrib/protospec/tests/test_tile_reader.py
+++ b/orangecontrib/protospec/tests/test_tile_reader.py
@@ -3,7 +3,7 @@ import unittest
 
 import Orange
 from Orange.data.io import FileFormat
-from Orange.preprocess.preprocess import PreprocessorList
+from Orange.preprocess.preprocess import PreprocessorList, Preprocess
 from Orange.widgets.tests.base import WidgetTest
 
 from orangecontrib.spectroscopy import get_sample_datasets_dir
@@ -73,8 +73,6 @@ class TestTileReaderWidget(WidgetTest):
         pp_out = self.get_output("Preprocessor", widget=self.preproc_widget)
         self.send_signal("Preprocessor", pp_out, widget=self.widget)
         self.assertEqual(self.widget.preprocessor, pp_out)
-        pp_from_model = self.preproc_widget._create_preprocessor(
-            self.preproc_widget.preprocessormodel.item(0), None)
-        # v0.3.13
-        # self.assertIsInstance(self.widget.preprocessor, type(self.preproc_widget.buildpreproc()))
-        self.assertIsInstance(self.widget.preprocessor, type(pp_from_model))
+        #pp_from_model = self.preproc_widget._create_preprocessor(
+        #    self.preproc_widget.preprocessormodel.item(0), None)
+        self.assertIsInstance(self.widget.preprocessor, Preprocess)
diff --git a/orangecontrib/protospec/widgets/owtilefile.py b/orangecontrib/protospec/widgets/owtilefile.py
[...]

PR #287, opened 2019-02-05
[FIX] Preprocess widget: always output preprocessor (also on no input data)

markotoplak:
Also some minor maintenance and comments.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpreprocess.py
+++ b/orangecontrib/spectroscopy/widgets/owpreprocess.py
@@ -3,22 +3,23 @@ import sys
 from collections import Iterable
 
 import numpy as np
+import pyqtgraph as pg
 
 import Orange.data
-import Orange.widgets.data.owpreprocess as owpreprocess
-import pyqtgraph as pg
 from Orange import preprocess
 from Orange.data import ContinuousVariable
 from Orange.widgets import gui, settings
 from Orange.widgets.settings import SettingsHandler
 from Orange.widgets.widget import OWWidget, Msg, OWComponent, Input, Output
+from Orange.widgets.data.utils.preprocess import SequenceFlow, Controller, \
+    StandardItemModel
 from Orange.widgets.data.owpreprocess import (
-    Controller, StandardItemModel,
     PreprocessAction, Description, icon_path, DescriptionRole, ParametersRole, BaseEditor, blocked
 )
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.utils.sql import check_sql_input
 from Orange.widgets.utils.overlay import OverlayWidget
+from Orange.widgets.utils.colorpalette import DefaultColorBrewerPalette
 
 from AnyQt.QtCore import (
     Qt, QObject, QEvent, QSize, QMimeData, QTimer, QBasicTimer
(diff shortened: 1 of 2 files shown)


orangecontrib/spectroscopy/widgets/owpreprocess.py
```diff
@@ -1707,7 +1707,8 @@ def _initialize(self):
             # will be triggered by LayoutRequest event on the `flow_view`)
             self.__update_size_constraint()
 
-        self.apply()
+        # call apply() to output the preprocessor even if there is no input data
+        self.unconditional_commit()
```
stuart-cls: See my reply in #285, but even calling unconditional_commit() won't be sufficient to output the preprocessor even if there is no input data.

orangecontrib/spectroscopy/widgets/owpreprocess.py
```diff
@@ -1780,7 +1781,7 @@ def set_data(self, data=None):
 
     def handleNewSignals(self):
         self.show_preview(True)
-        self.apply()
+        self.unconditional_commit()
```
stuart-cls: Do we really want to by-pass the auto-commit logic when we connect new data? Users expect if that box is unchecked that no calculations take place (important for large datasets).

orangecontrib/spectroscopy/widgets/owpreprocess.py
markotoplak: That change was not meant to solve your problem. It was just a description for future us of what is being done, but it did not change anything really, except calling `apply()` once instead of twice on load.

orangecontrib/spectroscopy/widgets/owpreprocess.py
markotoplak: It is how (other) Orange widgets currently work. We try to encourage users that want to prevent propagation at new data by pausing a connection (right click it and you have an option for that).

Until recently, I though that this works differently, but last time I was playing with loading -> preprocess -> ft (or reverse) -> hyperspectra and was happy that it works such as is. But I do not know which would be better really... The current implementation is good if you want to use a workflow to compare effects on different inputs, while what you propose could be better if you want user to modify prebuilt workflows for their data.

markotoplak:
I changed `apply()` to always output reference. Now it also sends None if there are no preprocessors so that the connecting line can be shown correctly.

orangecontrib/spectroscopy/widgets/owpreprocess.py
stuart-cls: I'll admit I'm a bit confused: I thought it worked like this:

1. Auto-commit on: any change to data input / widget settings is immediately propagated to the outputs
2. Auto-commit off: any change to data input is reflected in widget previewing, but the heavy-lifting / full-data calculations (i.e. Preprocessing, clustering, FFT, etc) is postponed until the user presses the "Commit" button.

I think 2. is superior to freezing the connection because you can still see the preview/info boxes. It is inferior because you can't tell from the canvas which widgets are not calculating automatically.

> The current implementation is good if you want to use a workflow to compare effects on different inputs, while what you propose could be better if you want user to modify prebuilt workflows for their data.

Hopefully we can find a way to support both? I thought this is what auto-commit check box is for.

markotoplak: What you are saying makes sense. This widget really is quite special because seeing a preview before doing something compute heavy is a must.

Did it ever work as your are saying and I recently changed it for the worse?

stuart-cls:
This is great, thank you for all your work on it.

orangecontrib/spectroscopy/widgets/owpreprocess.py
stuart-cls: I don't recall if this particular widget worked that way :) I will do some investigating and open a new issue if necessary to track it.

PR #288, opened 2019-02-06
[ENH] Normalize by reference

markotoplak:


markotoplak:
@stuart-cls, I though of adding this into the Normalize preprocessor in the Preprocess widget. Do you think it fits there?

stuart-cls:
Is this something people ask for? I'm all for adding things that help users discover the calculation they want. 

The reason I ask is that this is the identical calculation to Transmittance.

markotoplak:
@stuart-cls, hmm, I did not know it was identical to something we have. :)

borondics:
@stuart-cls @markotoplak in principle it is identical (mathematically) but do we have the infrastructure to do it? The Spectral Transformation preprocessor can only convert between the A and T representations, but not actually calculate either from single channel spectra. Or am I missing a feature?

stuart-cls:
>  Like using a cluster of an image as a BG

@borondics Do you mean something like cluster -> average cluster -> normalize by that average spectrum?

@markotoplak Perhaps we could re-use the calculation code (to avoid bugs due to subtly different implementations)? If your new code is more robust, we can use that. I think it is important to make calculations discoverable, a user might not think of using a transmittance calculation to ratio two spectra if they are not experts.

I like the other changes in here to abstract things.

@borondics Just saw your latest reply. If you set the input on the transform to Single Channel, it calculates transmittance from Single Channels (i.e. data.X/ref.X)

borondics:
@stuart-cls Yes, something like that workflow. 

I see now how the transmission calculation works. Another push for writing documentation... 

I think it would also make sense to add it to the Normalize preprocessor...

What do you think?

stuart-cls:
I agree to both documentation ;) and adding to the Normalize for discoverability.

I'd still like to combine the actual calculation code to avoid duplication and because it looks like this new code is more resilient to wrong-shaped references. @markotoplak can you add this new protections to the Transmittance and Absorbance transforms as part of this PR?

borondics:
How about removing this feature from the spectral transformation and adding it to Normalization?
I think there we are mixing features...

stuart-cls:
The original idea (which I think you started in #128) was to unify all spectral transforms in a single preprocessor...

With what you propose, a user wanting to calculate Transmittance has to know to Normalize by Reference but a user wanting to calculate Absorbance has to Normalize then Transform. I don't see the problem with having both interfaces using the same calculation code?

I do agree that the transform user interface as it is could be better :)

borondics:
Yes, it is true that you would need to combine two preprocessors to get A from two SC datasets, i.e. Normalize by reference and then Transform to A, but I think this is more transparent than just adding a reference input to the widget and then using the Spectral Transformation from SC to A.

stuart-cls:
I understand and agree with adding it to Normalize.

I just don't agree that removing it from the transforms would be any more transparent. If a user wants to normalize to a reference, they will look there. If a user wants to calculate a Transmittance, they will look for that word.

In any case, we'll have the functionality so :)

borondics:
OK, let's leave it in there and add it to Normalization

markotoplak:
I added Normalize by Reference to the Preprocess Spectra widget. This code uses the error mechanism from #300 (needs to be merged first) to complain about wrong reference inputs.

Also, I started splitting preprocessors into separate files. I also moved Integrate into a separate file, but that was just because I needed it to avoid a cyclic import.

PR #289, opened 2019-02-06
[RFC] owpreprocess: do not commit on data change if autocommit is off

markotoplak:
This goes against the general consensus in Orange, which is to commit/apply for new data (see biolab/orange3#3455). 

The ordinary workaround to disable commit on new data (freeze a connection) prevents previews. Thus, we could make an exception Preprocess. To show this on the canvas, widget now issues a warning.

From @stuart-cls (on #287):
> I thought it worked like this:
>
> 1. Auto-commit on: any change to data input / widget settings is immediately propagated to the outputs
> 2. Auto-commit off: any change to data input is reflected in widget previewing, but the heavy-lifting / full-data calculations (i.e. Preprocessing, clustering, FFT, etc) is postponed until the user presses the "Commit" button.
>
> I think 2. is superior to freezing the connection because you can still see the preview/info boxes. It is inferior because you can't tell from the canvas which widgets are not calculating automatically.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpreprocess.py
+++ b/orangecontrib/spectroscopy/widgets/owpreprocess.py
@@ -1501,6 +1501,7 @@ class SpectralPreprocess(OWWidget):
         reference_compat = Msg("Reference is not processed for compatibility with the loaded "
                                "workflow. New instances of this widget will also process "
                                "the reference input.")
+        uncommited_changes = Msg("There are uncommited changes.")
 
     def __init__(self):
         super().__init__()
@@ -1771,7 +1772,7 @@ class SpectralPreprocess(OWWidget):
 
     def on_modelchanged(self):
         self.show_preview()
-        self.commit()
+        self.invalidate()
 
     @Inputs.data
     @check_sql_input
@@ -1781,7 +1782,7 @@ class SpectralPreprocess(OWWidget):
 
     def handleNewSignals(self):
         self.show_preview(True)
-        self.unconditional_commit()
+        self.invalidate()
 
     def add_preprocessor(self, action):
         item = QStandardItem()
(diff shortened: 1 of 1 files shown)


stuart-cls:
Thank you for the link to 3455, I admit a lot of my understanding was from the wording in the Python script widget as well. I'm afraid I've been misinforming all my users for a long time about what the point of the auto-commit checkbox is for!

That said, I do think we need to break convention for this special widget. Perhaps we should use a different naming convention to distinguish from the "commit" nomenclature? "Freeze on new data"?

markotoplak:
I'd avoid adding a new checkbox. Do you have any suggestions for the button text (in both states)? Also, if we do it with warning, as in this PR, what do you suggest it would say?

stuart-cls:
What about "Process" when button is enabled / auto is unchecked and "Process Automatically"? This matches the paradigm in a number of Widgets but implies the processing part of Preprocessing?

I like the warning, it could say "New data / preprocessors have not been calculated." ? Hmm that's a hard one to get the wording for...

For the FFT widget, I called it "Calculate [Automatically]", operating under my misunderstanding of the point of that check box/button.

stuart-cls:
I've been thinking more about this over the last few days, as I process the new mental model about how "(auto)commit" is designed.

In fact, very few (any?) of the preprocessors are much different than other widgets in Orange in that you can set all their parameters without any data present. It's true the Preview is useful for sanity checking operations, but that's also true of PCA / k-means clustering which are also computationally intensive. If anything, the Integrate Spectra widget is more sensitive to previewing the data than Preprocess Spectra.

So instead I propose shelving this idea for a while, and really commit (ha) to the data flow model consistently (#291). If after a while we find this unworkable, we can revisit this special case.

Thoughts?

markotoplak:
We can wait a bit with this... If we had it for so long it can stay in a bit longer. :)

PR #295, opened 2019-02-21
Added test for IOError the Matlab reader to increase coverage

borondics:
data.py is an important part of the code but has very low coverage for tests...

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_readers.py
+++ b/orangecontrib/spectroscopy/tests/test_readers.py
@@ -1,4 +1,5 @@
 import unittest
+from unittest.mock import patch
 
 import numpy as np
 import Orange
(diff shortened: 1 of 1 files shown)


stuart-cls:
I'm all for coverage improvements, but can you explain what this test actually does as well? 

I'm wondering if it could be generalized to all readers or if it only helps the matlab reader.

borondics:
@stuart-cls, it tests whether the `IOError` is raised upon the file is not returned for any reason. As usual, I am trying to learn something, so @markotoplak was the mind behind it and he says this is actually not the best way of doing it, we should make a file that raises the exception instead of forcing the `IOError` through the `patch`, but we could not do that, so defaulted for this. In principle, we should test these cases and I hope that for the other readers we can do it properly.

PR #297, opened 2019-02-22
Omnic map reader tests

borondics:


markotoplak:
Thank you, but I see two problems:
1.) This is not a small map (it is more that 1 MB, for tests we would like something much smaller, like 100kb).
2.) Tests fail, because you forgot to change the file name. :)

The name would be trivial to fix, but the map is too big. I suggest you to open a new branch from master and try again (the big file would stay in history unless your either rebased it out or started anew).

borondics:
Oh. The file name is easy, I'll fix that but we couldn't make a smaller map
for some reason. This one has only 4 pixels.

stuart-cls:
@borondics For the agilent reader tests, I kept the map smaller but also made the spectra light through a combination of binning / very low resolution acquisition / narrow spectral window saved. Hope that helps.

markotoplak:
This compresses to 300kb so it is fine for me.

Issue #30, opened 2016-07-29
Integrate: Warn user when data is outside of spectral window


stuart-cls:
@markotoplak , 

In #29 I changed the integration window to be user-defined. I'd like to pop up one of those warning boxes to tell the user if their defined window contains values outside the available dataset, but I don't think I can do that from the Integrate() class directly. Suggestions?

PR #298, opened 2019-02-26
[ENH] Preprocessor warnings

markotoplak:
Preprocessors that display warnings need to inherit BaseEditorOrange. Now, BaseEditorOrange creates a layout with a (hidden) message bar, therefore preprocessor editors need to add controls to .controlArea, which lies above the message bar.

This PR also shows how it issue warnings in the CutEditor. Something like this could be used for #30.


markotoplak:
Individual preprocessors' message bars are not shown with pre-3.20 Orange. I'll fix it.

Issue #302, opened 2019-03-01
Consistent and safe divide-by-zero handling


stuart-cls:
In several places we have `/` or `/=` operators (at least Normalize, NormalizeReference, Absorbance, Transmittance).

Currently, we have:
`0 / 0 == Nan` -> Okay, as Orange can handle NaNs
`1 / 0 == inf` -> Not okay, as Orange / OWHyper chokes on (+/-) infs
`-1 / 0 == -inf` -> ditto

We could do something like `X[np.isinf(X)] = np.nan` after each divide or in a helper function? 

Or we could not re-invent the wheel and use this from scikit-learn:

```
def _handle_zeros_in_scale(scale, copy=True):
    ''' Makes sure that whenever scale is zero, we handle it correctly.
    This happens in most scalers when we have constant features.'''

    # if we are fitting on 1D arrays, scale might be a scalar
    if np.isscalar(scale):
        if scale == .0:
            scale = 1.
        return scale
    elif isinstance(scale, np.ndarray):
        if copy:
            # New array to avoid side-effects
            scale = scale.copy()
        scale[scale == 0.0] = 1.0
return scale
```

from https://github.com/scikit-learn/scikit-learn/blob/7b136e92acf49d46251479b75c88cba632de1937/sklearn/preprocessing/data.py#L70

This **changes the behaviour to keep val / 0 as val** but makes Normalize.Vector and the rest consistent, since Vector uses the sci-kit code already.

@markotoplak Thoughts?
Anyone else have strong opinions on divide-by-zero handling?

markotoplak:
I dislike sklearn's approach that quietly ignores the operation. I'd prefer to keep such values NaN. Do you see any advantages for this approach?

I like your utility function idea. And for consistency, I would reimplement Normalize.Vector. :)

stuart-cls:
I agree, it is unexpected to get an un-scaled value in the middle of a scaled plot. 

Let's go with:

- [ ] helper function for consistent behaviour which returns NaN for all divzero cases
- [ ] re-implement Normalize.Vector to match

as you suggest.

If anyone (myself included) starts working on this, just assign to yourself :)

stuart-cls:
To clarify the status of this issue: `Normalize.Vector` still uses sklearn's code and silently ignores the normalization so it should be re-implemented to not do that.

PR #305, opened 2019-03-05
Test and handle (-) infs in preprocessors

stuart-cls:
Part of #302 

- [x] Adding `replace_infs` around `np.log10()` made `divide_no_infs` redundant, so can probably be removed.
- [x] `test_unknown_no_propagate` fails now, since we introduce NaNs.
- [x] test_slightly_different_domain fails

Changed attribute name for Absorbance and Transmittance to "reference" to standardize naming (no longer needed in the PR but I left it in)

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/utils.py
+++ b/orangecontrib/spectroscopy/preprocess/utils.py
@@ -215,3 +215,10 @@ def interp1d_wo_unknowns_scipy(x, ys, points, kind="linear"):
 def edge_baseline(x, y):
     i = np.array([0, -1])
     return interp1d(x[i], y[:, i], axis=1)(x) if len(x) else 0
+
+
+def replace_infs(array):
+    """ Replaces inf and -inf with nan.
+    This should be used anywhere a divide-by-zero can happen (/, np.log10, etc)"""
+    array[np.isinf(array)] = np.nan
+    return array
--- a/orangecontrib/spectroscopy/preprocess/utils.py
+++ b/orangecontrib/spectroscopy/preprocess/utils.py
@@ -59,9 +59,9 @@ class CommonDomain:
 
 class CommonDomainRef(CommonDomain):
     """CommonDomain which also ensures reference domain transformation"""
-    def __init__(self, ref, domain):
+    def __init__(self, reference, domain):
         super().__init__(domain)
-        self.ref = ref
+        self.reference = reference
 
     def interpolate_extend_to(self, interpolate, wavenumbers):
         """
--- a/orangecontrib/spectroscopy/tests/test_preprocess.py
+++ b/orangecontrib/spectroscopy/tests/test_preprocess.py
@@ -93,7 +93,7 @@ def add_edge_case_data_parameter(class_, data_arg_name, data_to_modify, *args, *
 
 for p in [Absorbance, Transmittance]:
     # single reference
-    PREPROCESSORS_INDEPENDENT_SAMPLES += list(add_edge_case_data_parameter(p, "ref", SMALL_COLLAGEN[0:1]))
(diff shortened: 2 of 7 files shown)


orangecontrib/spectroscopy/preprocess/utils.py
```diff
@@ -215,3 +215,20 @@ def interp1d_wo_unknowns_scipy(x, ys, points, kind="linear"):
...
     i = np.array([0, -1])
     return interp1d(x[i], y[:, i], axis=1)(x) if len(x) else 0
+
+
+def replace_infs(array):
+    """ Replaces inf and -inf with nan """
+    array[np.isinf(array)] = np.nan
+    return array
+
+def divide_no_inf(numerator, denominator, in_place=False):
```
markotoplak: You added this function to suppress the warning, right? If warning suppression was not needed, I would prefer to call replace_infs directly.

stuart-cls: Actually, at first there was no replace_infs(), but then I noticed np.log10() with zeros in the reference caused the same problem (thanks to the test). 

I'll just simplify this and remove divide_no_inf entirely.

orangecontrib/spectroscopy/tests/test_preprocess.py
```diff
@@ -343,9 +343,14 @@ def test_unknown_no_propagate(self):
         for i in range(min(len(data), len(data.domain.attributes))):
             data.X[i, i] = np.nan
         for proc in PREPROCESSORS:
+            if hasattr(proc, "reference"):
+                max_nans = 1 + data.X.size - np.count_nonzero(proc.reference)
```
markotoplak: A bug. For Absorbance this gives max_nans of 42729, which are of course never found.

The problem seemed strange. The data used for the tests and the input main reference all have absolutely no zeros. Then, I saw that the problematic data is caused by `add_zeros`. :)

Solving this makes quite a mess, forces the same name just for the test to work (I do not not mind the same name though). And sometimes, references could be interpolated and thus their length could change, so the formula for expected number of NaNs is unclear.

Because it is hard to write I criterium, I would try attaching a flag like `skip_unknown_not_propagate` to the preprocessor (not reference!) in `add_edge_case_data_parameter` for `add_zeros` and just skip preprocessors with this flag.

Furthermore, `test_no_infs` should also have `data` with zeros, because they can also cause infs if the situation is reversed. You should also have data when a whole row (or, maybe, also a whole column) is a zero, as summing over it is common. But we can never catch everything like this...

stuart-cls: Good catch!

Thanks for the suggestions. I'll admit I don't know how to attach the `skip_unknown_no_propagate` flag as you suggest.

I'm also unclear how the data (numerator) having zeros can result in an inf?

markotoplak: Isn't data also denumerator in some Absorbace? ref/data.X?

markotoplak: For `add_edge_case_data_parameter`

```
    ...
    for i, d in enumerate(modified):
        kwargs[data_arg_name] = d
        p = class_(*args, **kwargs)
        if i == 5:
            p.skip_something = True
        yield p
```

And then you can check for it in the test with `hasattr`

stuart-cls: Thank you.

> Isn't data also denumerator in some Absorbace? ref/data.X?

Right :)

stuart-cls:
@markotoplak I think I've got it now. Would you mind taking one last look? 

Thanks for all your help on this one.

markotoplak:
Thank you!

PR #306, opened 2019-03-06
[ENH] OWPreprocess: show up to 100 preview spectra

markotoplak:
I think this makes sense because now users can see which corresponding pairs of spectra.


borondics:
I agree to this. How do we control the initially displayed number of spectra (besides maybe the Settings saved)? I think we should also set it to a higher number, maybe ~25?

markotoplak:
@borondics, it is a setting, which mean that a new widget will get the value of the last closed one.

I changed the default to 25.

Issue #301, opened 2019-03-01
Preprocess widgets lose saved preprocessors on upgrade from 3.19.0/0.3.13 -> 3.20/0.4.1


stuart-cls:
I'm still trying to narrow this one down, but if anyone knows why this is happening...

A workflow saved in Orange 3.19.0 (unknown orange-spectroscopy version) does not restore the peak analysis regions in Orange 3.20.1/spec 0.4.1 or in Quasar 0.3.0 (Orange 3.20.0)/spec 0.4.1. It works fine in Orange 3.18.0/spec 0.4.1, so it's not exactly a bug in orange-spec but it's our problem now....

Related issues/possible causes so far:

1. OWPreprocess and OWIntegrate have diverged on how they build the preprocessor chain
   - I don't think this is related but may cause weird bugs down the road
2. Orange 3.19.0/spec 0.3.13 -> Orange 3.20.1/spec 0.4.1 loses preprocessor settings
3. orange-spec 0.3.13 fails on Orange 3.20: can we force an update? This is confusing for users

stuart-cls:
Okay I think this might be related to #287 and friends. The code in `OWPreprocess.apply()` and `OWIntegrate.buildpreproc()` have diverged, even though they are nominally doing the same thing. I'll try to come up with a test to confirm this.

stuart-cls:
Narrowed down a version matrix to reproduce this:
Orange 3.19.0 / spec 0.3.13 -> Orange 3.20.1 / spec 0.4.1

Happens for both OWIntegrate and OWPreprocess.

I was going to test 3.20.1/0.3.13 -> 3.20.1/0.4.1 but over half the widgets fail to load due to 
`ImportError: cannot import name 'get_next_name' from 'Orange.widgets.utils.annotated_data'`

stuart-cls:
Okay I guess this is #276 ! The version soup and it working on 3.18.0/0.4.1 for some reason sent me down the wrong path.

I still think item (1) above should be consolidated and item (3) is profoundly umimpressive to users but the main thing is done here I guess.

markotoplak:
Hmm, I will look into (1). 

[...]

PR #307, opened 2019-03-06
[FIX] OWIntegrate: fix non-meta output

markotoplak:
I created that bug when I also started processing reference in #287. @stuart-cls noticed that something was off in #301.

It was another case of "what is not tested does not exist", so I added some tests.


stuart-cls:
This is nice. I was hoping we could consolidate the `create_output` code even more, but I see that the special casing for references and handling metas is too intertwined. 

I do wonder if the new `_build_preprocessor` method really adds anything to the readability or if it just makes it harder to compare OWIntegrate and OWPreprocess?

> It was another case of "what is not tested does not exist", so I added some tests.

We should make this our slogan ;)

markotoplak:
You are correct: `_build_preprocessor` adds nothing useful. I just left it like this because it was already a function which I just renamed.

PR #308, opened 2019-03-06
CurvePlot: fix select_at_least_1

markotoplak:
I introduced this bug with the thick/thin selected curve optimization.


markotoplak:
I noticed the bug when I tried to decrease the number of preview curves in Integrate.

PR #312, opened 2019-03-08
[FIX] Fix rubberband (qhull) ValueError on all-NaN sample row

stuart-cls:
Includes test.

I didn't touch the "FIXME notify user" on this `Except` block because updating/improving the baseline subtraction editor is a rabbit-hole I don't have time for right now. Plus it's Friday afternoon :)


markotoplak:
Thank you!

PR #313, opened 2019-03-11
Select column reader

markotoplak:
Reader that can read .txt files and can select the column inside, such as the files by ROCK beamline at Soleil.


markotoplak:
This PR will cause that .txt files will be opened with this reader (instead of Nea reader) by default. The users can still choose the Nea reader explicitly though.

Both are a tab-separated format and I did not figure a nice way to separate them.

PR #316, opened 2019-03-12
[FIX] Subtract DC value from IFG

borondics:
I think simply subtracting the ifg.mean() should be OK as we can assume that the IFG is good quality, which will mean that the points at the DC level will be overwhelmingly more than the centerburst ones, which should be symmetric around the DC anyway.


stuart-cls:
I agree. I'll just go ahead and merge this.

PR #319, opened 2019-03-21
[FIX] owpreprocess: fix segfault on widget removal

markotoplak:
Preprocess Spectra widget segfaulted if it was removed while preview info labels were still shown. A quick way to crash it was adding a widget (connected to the data) and removing it immediately.


borondics:
Nice catch! I tested with the master and this PR and it works fine now. Thanks!

PR #327, opened 2019-04-10
[ENH] Spectral Series widget

markotoplak:
This widget show all wavenumbers for a single chosen axis (or data table index).

For collagen it does this:
Screen Shot 2019-04-10 at 11 41 07


markotoplak:
Hmm, now that it is done, I see that this widget is very similar to the Heat Map widget, only transposed.

Screen Shot 2019-04-10 at 15 05 18

In the Heat Map widget you can not set the variable for order; here you set the X. But Heat Map, on the other hand, also shows classes and can do clustering. Also, Line Scan uses the same color schemas as HyperSpectra.

What do you think? Should we add this widget? Here in Brazil they wished for something like this.

borondics:
I think we should add it - it is an important feature for nanoIR...

However, should we try to extend Hyperspectra with line plotting
functionality instead?

I would also add a panel on the bottom and the right for line plots
corresponding to the clicked position (the spectrum at that point on the
bottom and the variation of the feature on the right throughout the
dataset).

This kind of feature would be useful for other datasets (time dependent,
temperature dependent, pressure dependent, etc. and-so-on and-so-on...:))

markotoplak:
@borondics, I do not think it fits that well in Hyperspectra. It shows a different content (does not integrate, it shows all wavenumbers).

Also, implementation-wise it would just complicate stuff.

borondics:
OK, but then let's call it something like "Spectral Series" and still keep
the options open for those bottom and side panels. As I said in the
previous post, it will be great for many things.

stuart-cls:
This is great! I've been dreaming about this functionality for a while as well. I agree with Feri that two line plot panels on the bottom and side for context would really add to the usefulness of this widget.

markotoplak:
I renamed it to "Spectral Series".

I am open to bottom / right panes + many possible improvements, but I am not going to implement them now.

PR #328, opened 2019-04-10
[ENH] New reader for neaspec files

1rafaelmayer:
Neaspec new files did not work.


orangecontrib/spectroscopy/data.py
```diff
@@ -955,6 +955,86 @@ def channel_type(a):
                                          metas=np.asarray(final_metas, dtype=object))
             return X, final_data, meta_data
 
+    def read_v2(self):
+
+        # Find line in which data begins ########################### Max header lines = 100 #######################
+        count = 0
+        with open(self.filename, "r") as f:
+            for _ in range(100):
```
markotoplak: Why the limit of 100? I know it works for the current files, but you could replace this with a while loop and it would be unlimited.

PR #331, opened 2019-04-26
Refactor hypercube

stuart-cls:
I found the hypercube-building code in owalignstack very useful, and I'd been using this refactored version in some scripts for a while. Since I need a hypercube for the binning widget, I thought I'd break it out for all to use.

Thoughts?


markotoplak:
Thanks! If you found it useful, then it is useful.

stuart-cls:
👍 Thanks for the lint

PR #332, opened 2019-04-26
[WIP][ENH] Block-view Binning Widget for hyperspectral images

stuart-cls:
Simple block-view binning widget. This first version hard-codes assumptions that the hyperspectral data of interest is a hyperspectral cube from imaging data and that binning blocks are square.

In future should be n-dim or at least support 1-3 binning dimensions.

owbin

Uses code in #331 

- [ ] Change widget name to "Binning"
- [ ] Add icon
- [ ] Documentation
- [ ] Replace nan hypercube error with nanmean and warning
- [ ] generalize block shape?

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owbin.py
+++ b/orangecontrib/spectroscopy/tests/test_owbin.py
@@ -1,0 +1,49 @@
+import numpy as np
+import Orange
+from Orange.widgets.tests.base import WidgetTest
+from orangecontrib.spectroscopy.widgets.owbin import OWBin
+
+
+class TestOWBin(WidgetTest):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.mosaic = Orange.data.Table("agilent/5_mosaic_agg1024.dmt")
+
+    def setUp(self):
+        self.widget = self.create_widget(OWBin)
+
+    def test_load_unload(self):
+        # just to load the widget (it has no inputs)
+        pass
+
+    def test_bin(self):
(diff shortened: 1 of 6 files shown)


markotoplak:
Thank you for this. Good work!

What you say if we change the name to "Binning"? I would like to avoid to use "Data" to often. Do you have an icon to suggest? :) You can also make one - no need to be fancy.

It is necessary for this widget to error when we have nan values in the hypercube? Perhaps we could relax that limit a bit and use nanmean and put up a warning if we ever did a mean where numbers and nans were mixed.

Also, in the future, we could maybe relax the block size assumption. But I am not sure how difficult would it be...

stuart-cls:
Thanks for looking at it!

I'm fine with the name change, naming things is hard. I can also try to add an icon.

Originally I left the nan error because I was thinking of just simple hyperspectral cubes from image array detectors, but I agree: nanmean would be a perfectly reasonable outcome. 

I actually already found the current blockshape assumption too limiting for a different application (binning along arbitrary continuous axes) and started generalizing it in another branch. Do you think I should just incorporate that in or do it as a separate PR. My only concern with two-PRs is that the bin_sqrt setting will obviously change.

stuart-cls:
Actually, about the name: most of our widgets are verbs like Interpolate or Integrate: to follow that convention the name should be "Bin". But on the other hand, that is very short!

Naming things is hard.

borondics:
Great functionality! Thanks, @stuart-cls!

I'll add an icon just to contribute a bit :)

PR #334, opened 2019-05-03
[FIX] emsc: do not call lstsq with NaNs

markotoplak:
It crashes tests on my Mac.


markotoplak:
Closed because #335 solves this in a more general way.

PR #337, opened 2019-05-07
[ENH] Hypercube: separate nan handling

stuart-cls:
This PR 

- moves the NanInsideHypercube exception back out of the get_hypercube utility function
- changes the behaviour of the OWBin widget to produce nanmean binned data and warn the user instead of erroring


markotoplak:
Thank you!

PR #341, opened 2019-05-10
[ENH] Linear baseline where a user can specify baseline (zero) points

markotoplak:


borondics:
Looks good, thanks!

I have two requests: 
- Could we add text boxes that show the position of the line on the graph next to the line? This would be useful for other preprocessors as well and then eventually for making nice annotated figures in the future. 
- Can we add non linear baselines? We don't need to go crazy on this, I think 2nd and 3rd order polynomial should be enough. If we make that general we can always enable more orders...

markotoplak:
@borondics, I will not do any of your requests in this PR.

PR #343, opened 2019-05-15
[ENH] OWBin: n-dimensional binning and non-square block shape

stuart-cls:
This adds support to OWBin for n-dimensional binning and non-square block shapes.

* `MAX_DIMENSIONS` is arbitrarily set to 5 to slightly simplify the GUI
* `utils.get_hypercube` has been generalized, but I kept `get_hypercube()` for simplicity / backwards compat
* I did **not** do a settings migration as the first version of OWBin has not actually been released
* ~I added a special case for 2-dimensional datasets: the code reverses the attr order (assuming the user sets Axis 0 to "x" and Axis 1 to "y") to maintain the row-major image table order. This is tested in `test_hypercube_roundtrip`~

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owbin.py
+++ b/orangecontrib/spectroscopy/tests/test_owbin.py
@@ -41,7 +41,7 @@ class TestOWBin(WidgetTest):
         np.testing.assert_equal(len(m.X), len(self.mosaic.X) / 2**2)
 
     def test_nonsquare_bin(self):
-        self.widget.bin_shape = (2, 4)
+        self.widget.bin_shape = (4, 2)
         self.widget._init_bins
         self.send_signal(OWBin.Inputs.data, self.mosaic)
         m = self.get_output(OWBin.Outputs.bindata)
@@ -59,8 +59,12 @@ class TestOWBin(WidgetTest):
         self.send_signal(OWBin.Inputs.data, self.mosaic)
         m = self.get_output(OWBin.Outputs.bindata)
         np.testing.assert_equal(self.mosaic.X, m.X)
+        np.testing.assert_equal(self.mosaic[:, "map_x"].metas, m[:, "map_x"].metas)
+        np.testing.assert_equal(self.mosaic[:, "map_y"].metas, m[:, "map_y"].metas)
         # TODO this should be true
-        np.testing.assert_equal(self.mosaic, m)
+        # Issue is the array is row-ordered (y, x) but the metas are (x, y) ordered
+        # and after binning the metas are reversed
+        # np.testing.assert_equal(self.mosaic, m)
 
     def test_invalid_bin(self):
         self.widget.bin_shape = (3, 3)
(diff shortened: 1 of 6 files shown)


markotoplak:
Thank you. I love the individual bin sizes.

I am confused about one thing: why did you decide to threat 2D data differently? Does the order of axes even matter? I thought it does, but then I tried it, and I get the same results if I change the order. And this then made sense, because the same order is used to go both ways. :) Am I missing something?

The current interface is fine. I see two other possibilities of structuring it:
1) Always show 5 fields, which could also be None - if None, that dimension is ignored.
2) Start with a single dimension and an "Add button". 

Then I have a small bug: setting two axes to the same value makes a crash.

stuart-cls:
> why did you decide to threat 2D data differently? Does the order of axes even matter?

When I was testing the hypercube roundtrip in fd51a20 I found that since `data._spectra_from_image()` assumes the order is [row, column, energy], it was rewriting the array unnecessarily (and failing the strict roundtrip test). So it was more of a memory layout / speed efficiency optimization than anything. I could remove the special case and just not worry about it?

About the interface, I think I'll leave it for now. I want to add some more user feedback about the array shape and resulting step size per dimension, so when I do that I can simplify the interface as you suggest.

I can't reproduce the bug you describe, what dataset are you using?

markotoplak:
> When I was testing the hypercube roundtrip in [fd51a20](https://github.com/Quasars/orange-spectroscopy/commit/fd51a20490092f859564d01ac0da49b7649d923f) I found that since `data._spectra_from_image()` assumes the order is [row, column, energy], it was rewriting the array unnecessarily (and failing the strict roundtrip test). So it was more of a memory layout / speed efficiency optimization than anything. I could remove the special case and just not worry about it?

Ah, it was for speed! But I do not understand why would a test fail if the arrays are rearranged a few times.

> I can't reproduce the bug you describe, what dataset are you using?

I was using the liver image from the Datasets widget.

stuart-cls:
Perhaps "premature optimization of an edge case" (binning of 1) would be a better description :) 

I directly compared the `.X` array before and after the hypercube roundtrip test. Since the original dataset was flattened (y. x) and the returned dataset flattened (x, y), the `.X` array had a different layout after the roundtrip.

I think I will remove the special casing in OWBin as the actual use case of the widget requires writing a new array anyway.

Regarding the bug, I see that it is actually OWHyper that crashes (I was first testing the OWBin alone). Probably the output in the case of (map_x, map_x) binning makes no sense (it certainly makes no practical sense) so I'll just change it not to do that.

markotoplak:
Sorry, it it is a problem in Hyper, do not try to fix it in Bin. It will have to be fixed in Hyper.

stuart-cls:
I've removed the special casing for 2-dimensions and updated the tests to match.

It was actually `test_no_bin` that started me down this road, but since this is an edge case (using a binning widget without actually binning anything) I'm fine with the array rearrangement.

Should be good to go now.

markotoplak:
Thank you!

PR #366, opened 2019-08-12
[ENH] Spectra, Hyper: selection also with mouse drag

markotoplak:
Also allow selection with mouse drags.


borondics:
Looks good to me.

PR #371, opened 2019-09-19
File reader for the Gwyddion Simple Field (.gsf) files

borondics:
This is a new file reader that will open possibilities to process AFM related data in the format of the Gwyddion software. The addition is to open to NeaSpec users.


markotoplak:
@borondics, do you have any small tests files? Can you write some simple tests?

PR #372, opened 2019-09-19
[FIX] GSF file reader data indexing

borondics:
- simplified the code so that it uses the same functions as other readers
- fixed the data indexing to be consistent with Gwyddion and NeaSpec orientation
- added tests for checking pixel values

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -797,15 +797,15 @@ class SPAReader(FileFormat, SpectralFileFormat):
         return domvals, data, None
 
 
-class GSFReader(FileFormat):
+class GSFReader(FileFormat, SpectralFileFormat):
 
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks. Travis in complaining because you are missing 3 spaces, but I will not bug you for that.

markotoplak:
Because you were so careful to implement correct data indexing you should verify it with tests. Like test of x and y map positions are fine.

borondics:
This I think is verified by the pixel value. If the indexing gets screwed
up, there is very little chance that the new values will be correct, no?

markotoplak:
Yes, indexing is verified by the pixes value, but the corresponding x and y positions are not. If someone screwed up positioning, these tests would not catch them.

borondics:
OK, I added the tests for the coordinates.

PR #387, opened 2019-12-04
[ENH] EMSC, me-EMSC: smoothed weighting

markotoplak:
- [x] flexible smoothed/non-smoothed weighting in the "library"
- [x] add smoothing parameter to the UI of preprocessors
- [x] show weighting function the preview


markotoplak:
@JohanneSolheim, could you test this a bit? You can see how the smoothing parameter relates to kappa from your previous implementation from in test_me_emsc.

markotoplak:
The weight construction, where I specified two intervals, one that is not smoothed and one that is tanh smoothed.

Screen Shot 2019-12-19 at 15 12 07


@JohanneSolheim , could you play with this a bit to see if it makes any sense? If you need any help getting this branch running, contact me.

PR #391, opened 2019-12-12
Reader NeaSpec raw GSF files

levandoskije:
I Create some methods in the NeaSpec class and now we can work with raw NeaSnom data.
I run the test and my modification does not affect the interface.


orangecontrib/spectroscopy/data.py
```diff
@@ -1014,16 +1014,195 @@ def read_v2(self):
...
+                    f = i + px_z
+                    print('apos o final_metas')
+        print('formatou')
+        print('info', info)
+        return np.asarray(data_complete), info, final_metas
+
+    def _html_reader(self, path):
+        from pandas import read_html
+        try:
+            parameters = np.asarray(read_html(path)[0])
```
markotoplak: If so, we should consider adding lxml as a dependency. If reading is done by lxml, why do we also need pandas? Do we need both?

markotoplak:
Thank you. To test this a bit, I would need some examples (files from some experiments).

orangecontrib/spectroscopy/data.py
```diff
@@ -1014,16 +1014,148 @@ def read_v2(self):
...
+        except ImportError:
+            print('''Probably you need to install xlml\nTry\n "pip install lxml"\nor\n"conda install -c conda-forge lxml"''')
+            pass
+
+        return parameters
+
+    def _gsf_reader(self, path):
```
markotoplak: Please, never copy the same code to two places. We would need to have one function that both this reader and GSF reader would use. If the GSF does not allow reuse (code is not structured as such), we would need to change the GSF reader's code and then reuse it here.

orangecontrib/spectroscopy/data.py
levandoskije: I read html using pandas, but when a call the function `read_html` it need `lxml` already installed. I think it's like dependency of pandas.

orangecontrib/spectroscopy/data.py
levandoskije: I fix this, but I used to return a little different.
So I created a global variable to change the flow, this global can be useful for more cases.

orangecontrib/spectroscopy/data.py
```diff
@@ -17,6 +17,8 @@
 from .pymca5 import OmnicMap
 from .agilent import agilentImage, agilentMosaic, agilentImageIFG, agilentMosaicIFG
 
+SET_flow = False    # Use for manual control flow in some function inter Class
+                    # See example in class NeaSpec and GSFReader
```
levandoskije: @markotoplak  I created this global variable to use GSF reader, because I needed a return before the default return.
I think this approach can be used in many other cases.
But if you have any suggestion, teel me

orangecontrib/spectroscopy/data.py
markotoplak: Please do not do this. Using a global variable here makes the code prone to bugs. You should avoid using such programming structures at all costs for now. There are very few cases where global would be meaningful.

What you should do is to copy the contests of that function into a new one, which you would call from both GSF reader and you nea reader. For example, you could make a new function (and put it out of a class) from almost all of `read_spectra`, until `X = X.reshape((meta["YRes"], meta["XRes"]) + (1,))`. The input to that function would be a filename, and it would return X, XRr and YRr. Therefore, this new function would be usable in GSF, where `read_spectra` would now only have a few lines, and in your NEA reader.

orangecontrib/spectroscopy/data.py
markotoplak: I answered with a suggestion about making a new function. For now, please avoid using global.

orangecontrib/spectroscopy/data.py
```diff
@@ -837,16 +839,24 @@ def read_spectra(self):
...
+
             data = _spectra_from_image(X, [1], XRr, YRr)
 
             return data
 
 
 class NeaReader(FileFormat, SpectralFileFormat):
 
-    EXTENSIONS = (".nea", ".txt")
+    EXTENSIONS = (".nea", ".txt", '.gsf')
```
markotoplak: This will simplify your code: if would be better if the new v3 reader is not implemented in the same class. That would simplify the read_spectra() function here.

So please restructure your code so that you make a new class called NeaReaderGSF which only contains your new reading code and leave the old NeaReader as it was before your changes.

This is better because v2 and v3 NeaReaders do not share any functionality or extensions. Therefore separating the code into multiple classes makes it more maintainable and prevents possible bugs with selection of the correct reader. That will avoid the try...excepts read_spectra.

orangecontrib/spectroscopy/data.py
```diff
@@ -1014,16 +1024,118 @@ def read_v2(self):
...
+
+        averaging = int(info['Averaging'][1])
+        px_x = int(info['Pixel Area (X, Y, Z)'][1]) # 10
+        px_y = int(info['Pixel Area (X, Y, Z)'][2]) # 10
+        px_z = number_of_points = int(info['Pixel Area (X, Y, Z)'][3]) # 1024
+        description = info['Description'][1]
+
+        data_complete = []
+        final_metas = []
+        for y in range(0, px_y):
```
markotoplak: Just a note... This code is fine for now, but there are many opportunities for better memory and processing efficiency by using numpy matrices in a better way. It is probably possible to avoid most of the for loops. Don't work on this now, we can always improve it later.

orangecontrib/spectroscopy/data.py
```diff
@@ -1014,16 +1024,118 @@ def read_v2(self):
...
-        with open(self.filename, "rt") as f:
-            if f.read(2) == '# ':
-                version = 2
+        try:
+            with open(self.filename, "rt") as f:
+                if f.read(2) == '# ':
+                    version = 2
+        except:
+            pass
+        try:
```
markotoplak: Empty excepts always look suspicious. As I advised in another comment, if you make your new functionality in a separate class you will avoid them. Simpler structure = more stable problemas.

orangecontrib/spectroscopy/data.py
markotoplak: So. I checked. Pandas does not install `lxml` automatically, so we need to remember it to add to our project's requirements.

Did you perhaps look if there are any other good ways to read this table without involving so big dependencies? I usually look for what is the simplest tool for the job. That keeps the final installation size small. It is fine to use pandas here if you decide it is the only solution that makes sense, but please look for alternatives.

markotoplak:
Thank you for adding the test files! Do you have (or you can make on the machine) any test files that are smaller in size? This is important because they are included in all installations. Keeping them small makes download times faster for everyone.

orangecontrib/spectroscopy/data.py
levandoskije: I have looked for an alternative `html` reader as  an array, but all these involving the same dependencies `lxml`, so kept the same opening package.

levandoskije:
@markotoplak 
I'll close this Pull request, because it's easier to start a new one with these modifications.

PR #392, opened 2020-01-06
New class NeaReaderGSF

levandoskije:
I created a new class to read GSF files from NeaSnom.


markotoplak:
Cool, thanks! This looks very good.

The test files are great. Could you also add some automatic tests such as the ones you find in test_readers? Please add tests that both confirm that values and the meta data are read correctly.

Then, I'll look a bit into parsing that html file differently. It is so simple that it could be parsed manually. I would still like to avoid making our distribution even bigger with new dependencies.

levandoskije:
@markotoplak 
I added unit tests for values and meta data.

markotoplak:
@levandoskije, I rebased your PR to the latest master and fixed some minor things.

Please, next time, try to lint your code (the command that shows your errors is "python setup.py lint"). Normally, this is automatically executed by travis on every PR - I do not understand why it was not the case here. I suggest that you either properly configure your editor for Python or just use pycharm.

If lxml was missing, I changed your print to an exception, because the code crashed later on anyways and in this way an user will be able to see the error in the File widget.

I did not do any rewriting of the html reading. For now, the users will need to install lxml manually.

markotoplak:
@levandoskije, please go carefully through my changes so that you will know what to watch out for next time.

Thank you for the work that you did here! It seems to be working well.

As a next step we could start thinking about optimizing it, but to do so properly, you will need to start trying it out on really big files. The current version seems that it could be a bit slow or potentially wasteful with memory in some cases.

levandoskije:
@markotoplak, thank you for your help and attention!
I will be more careful next time.
How about the big file, I will test here and inform you about the results or bugs.

PR #393, opened 2020-01-13
NeaReaderGSF: avoid lxml dependency

markotoplak:
Avoid lxml dependency in NeaReaderGSF.

Also, the output is now a python list instead of a numpy array.


markotoplak:
@levandoskije, could you check if this is fine for you? I believe dropping a dependency will make it easier to install and support.

levandoskije:
HOH! It's great!
@markotoplak, I tested the new parsed and worked fine.
Thank you!

Issue #44, opened 2016-10-23
Progress bar to processors


borondics:
The k-means clustering in core Orange has a very nice progress bar that goes around the widget icon. Could we have the same for all our stuff? The reason is for large dataset analyses.

markotoplak:
Closing this because of #395.

Anyway, adding progress bars to widgets seems github-issue independent anyway. :)

PR #395, opened 2020-01-16
[ENH] More responsive Preprocess and Integrate widgets; progress bars

markotoplak:
Towards fixing #44

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       env: ORANGE="release"  RUN_PYLINT=true
 
     - python: '3.6'
-      env: ORANGE="3.11.0"
+      env: ORANGE="3.24.0"
 
     - python: '3.6'
       env: ORANGE="release"  UPLOAD_COVERAGE=true
@@ -25,8 +25,6 @@ cache:
 install:
     # Update pip to be able to install wheel dependencies
     - pip install -U pip wheel codecov
-    # Orange before 3.18.0 is imcompatible with scikit-learn 0.22+
-    - if [ $ORANGE = "3.11.0" ]; then pip install scikit-learn~=0.21.0; fi
     # 20180124 force PyQt 5.9.2 due to segfaults on new 5.10
     - pip install pyqt5==5.9.2
     - source $TRAVIS_BUILD_DIR/.travis/install_orange.sh
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - recommonmark
   run:
     - python
-    - orange3 >=3.11
+    - orange3 >=3.24
     - scipy >=0.14.0
     - spectral >=0.18
     - serverfiles >=0.2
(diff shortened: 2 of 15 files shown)


markotoplak:
With the current change the preprocess widget moves the computation into another thread, which means it does not block the user interface.

Currently, it is only interruptible between chain of preprocessors. For more we will need to add callback support to preprocessors. And then we might even clone #289 for ever.

@stuart-cls, I'll appreciate it a lot if you play around with this a bit. It is hard not to introduce bugs while doing this.

markotoplak:
Previews are also interruptible now. The widget should be more responsive in general, and should stay responsive even if working with slow preprocessors, such as me-EMSC.

@borondics and @stuart-cls, try to crash it, please.

borondics:
I think this is a great addition and also very important to be more interactive.

However, there are a few things that came to my mind while playing with it.
- It is not clear how to interrupt the calculation of the preprocessor (at least I couldn't see it).
- The user can make modifications to the preprocessor parameters, even delete them while the preprocessing is running.
- If a preprocessor is running it is still possible to add widgets to the Canvas, but they don't do anything. This defeats the purpose a little bit, although I see that the previously added and finished widgets are still opening and are somewhat responsive, but there can be conflicts I guess.

So far it didn't crash. :)

markotoplak:
@borondics, thanks for trying it out! If not crashing is a very good sign.

Yes, their is no interrupt button - currently, things are interrupted only implicitly when you give new input or change parameters. We could easily add a button though: it is one line of cure.

Now, computation can only be interrupted between different preprocessors; to make it denser, we will need to modify some Orange internals

After I fix tests this will be ready for merging. I'll come back to this after we make improvements in Orange. My next stop: PCA widget.

markotoplak:
@borondics, you expressed a concern about for example deleting a preprocessor (or changing parameters) while it is being computed. In such cases, the current computation gets invalidated as soon as possible (and absolutely never gets to the output of the widget). The new computation is started as soon as the old one managed to be interrupted. At least that was what I tried to do.

stuart-cls:
This is very nice :)


I found one weird thing so far while trying to break it with the "Integrate" widget:
(But it turned out to be very hard to reproduce and is probably a weird Windows edge case, nothing actually crashes)

1. With 150k row dataset connected and Commit Automatically deselected,
2. Add about 5 peak features, including a few peak areas
3. Add some visualizations but leave one peak area hidden
4. Select Commit Automatically
5. While the computation is running, select the visualization of the peak area
6. The widget GUI minimizes itself at some point and can't be restored until the calculation finishes

This was hard to reproduce, I had to restart Orange a few times to clear some cache or something. I also noticed that it triggers the "Not Responding" title bar when this happens.

Maybe a race between multiple preview visualizations and the main calculation?

markotoplak:
@stuart-cls, thanks for testing!

Currently, Table.transform() is limited to 1 concurrent thread (it calls Table.from_table which is surrounded by a Lock), because it uses some kind of global cache. Therefore, if integrals for a bit data set are being computed, preview will wait until that operation is finished.

We will need to remove that lock from Table.from_table. Still, I think this PR, while imperfect, is a step in the right direction.

stuart-cls:
> We will need to remove that lock from Table.from_table. Still, I think this PR, while imperfect, is a step in the right direction.

I completely agree! It works really well and is a big improvement. Thanks for your hard work on it.

PR #396, opened 2020-01-26
[FIX] NeaReaderGSF: Changed order of metas in NeaReaderGSF

levandoskije:
I realized that I made a mistake.
This change is necessary to view the X and Y axes in the hyperspectral data.
I took the opportunity to insert a key in the attributes, that key will be very important for the FFT - I will finish it soon.

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -1130,12 +1130,13 @@ class NeaReaderGSF(FileFormat, SpectralFileFormat):
         info = {}
         for row in parameters:
             info.update({row[0].strip(':'): row[1:]})
+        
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks! I think it would be better if the FFT you are making would not rely on the specific reader. I am not sure how you want it to work, but I imagine it could work with any data with paired amplitude -phase right?

I would say that inserting a key like this is "cheating", which will make your next widget less useful. Ideally, we need to think how to make widgets useful in as many ways as we can.

levandoskije:
Yes! You are right.
They have another option, I thought about making a checkbox for the user to choose. But I haven't done that yet.
[https://github.com/levandoskije/orange-spectroscopy/tree/complex_fft]
This is the branch I'm working on now, if you want to take a look.

PR #398, opened 2020-01-28
[ENH] Responsive Hyperspectra widget

markotoplak:
Computing integrals does not block the widget anymore.

Also, while working on this, I noticed that the image was needlessly recomputed when a new spectra was chosen for integral preview.

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owhyper.py
+++ b/orangecontrib/spectroscopy/tests/test_owhyper.py
@@ -5,9 +5,11 @@ from unittest.mock import patch
 import numpy as np
 
 from AnyQt.QtCore import QPointF, Qt
+from AnyQt.QtTest import QSignalSpy
 import Orange
 from Orange.widgets.tests.base import WidgetTest
 
+from orangecontrib.spectroscopy.widgets import owhyper
 from orangecontrib.spectroscopy.widgets.owhyper import \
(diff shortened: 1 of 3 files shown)


markotoplak:
@borondics, @stuart-cls, could you perhaps try it out a bit?

stuart-cls:
Very nice! This will make some of our users very happy...

It works exactly as advertised and I was unable to crash it so far.

I noticed a few things on a large (~1M row) dataset:
 - I like how the image disappears when you change the region of integration, although this might be slightly confusing as there is no progress bar indicating calculation is happening for long-running integrals
 - When "Show averages" is selected, initial display of the widget is delayed by the average spectra calculation (I assume, since once the window updated the average spectrum displayed but the map did not and the map calculation then started on a different thread)
 - Changing the region of interest mid-calculation doesn't cancel the calculation. I think you said this has something to do with Preprocess and a global cache lock in Orange.

But those are all bonus things compared to the main event!

markotoplak:
@stuart-cls, thanks!

- Yes, when the image is recomputing, there is no indication that anything is going on. Someday. :) I would prefer if the image didn't go black and would rather have some text on it such as "updating" or something...
- When Curveplot's get multithreaded you will not need to wait for the average spectra.
- Yup, changing mid-calculation does not cancel anything for now. It is in works. Global cache lock is gone in Orange master, and the next step is to make domain transfomations interruptable. Then it will work fine.

PR #399, opened 2020-01-29
[ENH] Responsive Spectra widget

markotoplak:
- compute averages in a separate thread

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owspectra.py
+++ b/orangecontrib/spectroscopy/tests/test_owspectra.py
@@ -2,7 +2,7 @@ import os
 from unittest.mock import Mock, patch
 
 from AnyQt.QtCore import QRectF, QPoint, Qt
-from AnyQt.QtTest import QTest
+from AnyQt.QtTest import QTest, QSignalSpy
 import numpy as np
 import pyqtgraph as pg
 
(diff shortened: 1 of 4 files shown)


markotoplak:
@stuart-cls, this also fixes an issue you mentioned yesterday. Now Hyperspectra is not unresponsive when you connect big data.

stuart-cls:
Very nice! I see the 200% usage you mention in the other PR as well. Works with multiple (hyper)spectra windows going at the same time too.

This does make the absence of the progress bar more problematic because now there is no visible indication that any data is connected or that anything is happening for quite a while for large datasets (assuming you aren't watching your system monitor obsessively ;) ...) Actually even the usual red dot on the input doesn't light up for hyperspectra.

Also I tried quitting while two were chugging away and it did take a while to stop the calculation threads.

markotoplak:
For images I was thinking about showing the partial image, so that you would see it fill up: a visual progress bar.

I do not know what to do about averages though.

markotoplak:
The red dot not lighting up is normal: the widget already processed its stuff: it already sent any selections to the output. So yes, we do need to show users that something is being done.

stuart-cls:
I think as long as there is some indication you successfully connected data to the widget that will help.

PR #400, opened 2020-01-29
Hyperspectra: users can interrupt image computation

markotoplak:
Disconnecting data now stops processing, changed settings are (almost) immediately applied.

Merge after #399

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_utils.py
+++ b/orangecontrib/spectroscopy/tests/test_utils.py
@@ -7,7 +7,7 @@ import Orange.data
 
 from orangecontrib.spectroscopy.data import _spectra_from_image, build_spec_table, getx
 from orangecontrib.spectroscopy.utils import get_hypercube, index_values, \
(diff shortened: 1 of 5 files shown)


borondics:
I think these are great and very important improvements. One question: how will the possibility of interruption be conveyed to the user? I think it would be good to implement some kind of visual clue (square "stop play" button on top of the widget?) that would make it obvious to even those who use Orange for the first time.

markotoplak:
Yes, something like that could work nicely. For now, users will only discover that they can still use the widget while it is computing. If they change anything, the widget will start to work with new data.

Did you notice that Hyperspectra now uses 200% of CPU for display? To notice it you have to try with some big data. I was testing with a 2GB file from Olga.

stuart-cls:
This is fantastic!

I agree that most users I see just try to change things anyway and now will be pleasantly surprised instead of unpleasantly disappointed that their computer is busy for a long time with queued up calculations!

PR #409, opened 2020-03-23
[FIX] Filter non-finite values from spectra plot

### LINKED ISSUE #408: CurvePlot: no display for spectra with NaNs when Qt >= 5.12.5
The latest installer for Quasar (0.7.1) includes a version bump of Qt to 5.12.5. This means we are now triggering https://github.com/pyqtgraph/pyqtgraph/issues/1057 on spectra containing NaNs.

@markotoplak I can take a stab at this but if
    (a) you prefer to do it
 or (b) you have at least a preferred entrance point
I'd be happy for your input. 

Currently I'm thinking of filtering the input data set in CurvePlot.set_data() for NaNs.

stuart-cls:
in owspectra: subclass PlotCurveItem to filter non-finite values from Qt

fixes #408


stuart-cls:
It looks like the build failed for two reasons unrelated to this PR: something changed in Orange master, and pylint didn't like me adding to the already-very-long `owspectra.py`

Issue #360, opened 2019-07-11
Multiple CPU calculations


borondics:
@stuart-cls @markotoplak do you think it would make sense to use the `multiprocessing` python module to split calculations like the FFT to all the CPUs in a machine? I was converting some data the other day and it took hours...

markotoplak:
It is possible. For calculation on data in memory, it might even be better to use threading as long as python's GIL is not a problem. Renaming this issue to "FFT should be faster" because there are some thing already that could be improved and as I understand, @stuart-cls is already working on is.

stuart-cls:
I am indeed! I have a prototype that trades memory usage for speed by letting the underlying fft routine handle the multiprocessing, next step is to tune the memory/speed tradeoff.

markotoplak:
@stuart-cls, if you work in batches of ~100 I think you will keep most of the speedup and still remain memory friendly. We can fine tune it later. :)

PR #413, opened 2020-04-02
[ENH] FFT multirow processing

stuart-cls:
Enhancements to the FFT processing as mentioned in https://github.com/Quasars/orange-spectroscopy/issues/360.
Currently showing speedups of 1-10x depending on datasize. Chunking into batches 100 as suggested by @markotoplak keeps the speed improvements but completely mitigates memory issues.

This code is only used for datasets with fixed, constant zpd values. The code I wrote to support variable zpd values from a peak search was too complex and fragile. Since this speedup is mostly useful for large imaging-array detector datasets, I think this is an acceptable tradeoff.

- [x] process multiple rows at a time
- [x] tests for multirow results and zpds

Not in scope: Fix the poor API boundary between irfft and owfft and add a preprocessor layer for use in scripts. That will be next.

### DIFF
--- a/orangecontrib/spectroscopy/irfft.py
+++ b/orangecontrib/spectroscopy/irfft.py
@@ -272,8 +272,7 @@ class IRFFT():
         elif self.phase_corr == PhaseCorrection.STORED:
             if self.phase is None:
                 raise ValueError("No stored phase provided.")
-            else:
-                return
+            return
         elif self.phase_corr == PhaseCorrection.MERTZ:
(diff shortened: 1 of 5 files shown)


stuart-cls:
I edited above: This code is only used for datasets with fixed, constant zpd values. The code I wrote to support variable zpd values from a peak search was too complex and fragile. Since this speedup is mostly useful for large imaging-array detector datasets, I think this is an acceptable tradeoff.

markotoplak:
@stuart-cls, looks cool. Thanks!

PR #417, opened 2020-04-17
PLS Implementation

realrk69:
The coefficients output has been fixed. PLS uses scikit learn and works only with regression. I can work to create PLS transformation widget if you guys see a need for it. 

I am creating some widgets to work with this:

1. Variable PLS Selection: Outputs wavelengths or features (data.X) containing most information with relevance to target (data.Y) along with an informative plot
2. Moving Window: Performs similar function but over specific windows of the entire wavelength range (slower)

Let me know if more information is needed. As an addition, I'm uploading the wheat kernel datasets so you can take test this and can be used by the community as a whole for testing and training.


markotoplak:
@realrk69, thank you, I will thoroughly test you PR in the next few days.

I quickly checked it and it seems great. Just a few quick notes.

Do we really need to add another data set? We try to keep data in the repo as tiny, because they make the repository/distribution larger. Perhaps some data set that are already in the repository could be used for testing. Or, another option, test data can be generated programatically, in tests.

Did you perhaps forget to commit tests? I remember you had some in you Orange PR.

realrk69:
You're right. There is honestly no need for that data. But in my experience, I found the collagen dataset to be very useful for spectral classification, but could not find some spectral data to perform regression  on. I'll remove it from the PR.

And yes, I forgot to commit the test 😐. Have added it.

realrk69:
I want to plot a graph inside the Variable PLS widget. What should I use, pyqtgraph or matplotlib?

Can I reuse OWSpectra widget for this? (embed OWSpectra widget inside this widget) Is there some other widget I could use?

realrk69:
Ok I figured out the plotting part. Was banging my head for nothing...

I need some info, how can we change the X of Orange.Table?
For example, I have a function which transforms data.X. I'm not very familiar with this format tbh.

`
new_data = self.data
output_of_function = self.function(new_data.X,new_data.Y)
new_data.X = output_of_function
self.Outputs.optimized_data(new_data)
`

This gives me an error:
`
IndexError: Index 17 is out of bounds for axis 1 with size 17
`
This error also happens when I import data in Correlation Widget in Data category of widgets.

Some pointers appreciated.

realrk69:
Got it, Table.from_table. The datasheet is indeed very helpful

markotoplak:
@realrk69, in #419 I build upon your code. Normally I would push my changes into this branch, but you worked directly on your master and I did no dare modifying it. Next time it might be better if you branched first before making any changes.

I did the following changes:
1. I squashed all your commits into a single one, mainly to remove the data you accidentally committed, because adding a file and removing in later on still keeps the file in project history forever. You remained the author of the code. It might be better, if you set you git user info properly, now you are visible as `realrk69 <40699981+realrk69@users.noreply.github.com>`.
2. I fixed your code so the tests are passing. There were small changes. Do not forget to run tests yourself before opening a PR. Also, if you scroll up, commits have a symbol (red cross or ideally, a green tick). That is our automated testing. If you click the symbol you can get more info (you can also click "Details" under "All checks have failed").
3. I simplified your code and wrote some tests. I also fixed some further lint complaints by our automated testing code.

Please, review my changes from #419 and, if you have any comments, write there.

PR #419, opened 2020-04-21
[ENH] Partial least squares regressor

markotoplak:
Replaces #417, only works for 1D Y variables.

### DIFF
--- a/orangecontrib/spectroscopy/models/modelling/pls.py
+++ b/orangecontrib/spectroscopy/models/modelling/pls.py
@@ -1,6 +1,0 @@
-from Orange.modelling import SklFitter
-from orangecontrib.spectroscopy.models.pls import PLSRegressionLearner
-
-
-class PLSRegressionLearner(SklFitter):
-    __fits__ = {'regression': PLSRegressionLearner}
--- a/orangecontrib/spectroscopy/models/modelling/pls.py
+++ b/orangecontrib/spectroscopy/models/modelling/pls.py
@@ -1,0 +1,6 @@
+from Orange.modelling import SklFitter
+from orangecontrib.spectroscopy.models.pls import PLSRegressionLearner
+
+
+class PLSRegressionLearner(SklFitter):
+    __fits__ = {'regression': PLSRegressionLearner}
--- a/orangecontrib/spectroscopy/models/pls.py
+++ b/orangecontrib/spectroscopy/models/pls.py
@@ -22,30 +22,18 @@ class _FeatureScorerMixin(LearnerScorer):
         return np.abs(model.coefficients), model.domain.attributes
 
 
-# Simple PLS Regression:
-
-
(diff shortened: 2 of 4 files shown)


realrk69:
Updated my commit email. Don't know if its updated correctly.
I understand the testing widget now. Thanks for the update. Let me know if something else has to be done here. I'll be creating the Variable Selection widget soon.
Do you recommend creating a fork rather than a branch?

realrk69:
Updated my commit email. Don't know if its updated correctly.
I understand the testing widget now. Thanks for the update. Let me know if something else has to be done here. I'll be creating the Variable Selection widget soon.
Do you recommend creating a fork rather than a branch?

realrk69:
Sure, this looks great. One last thing I wanted to implement was to set a limit to the max number of components to len(self.data.domain.attributes), since exceeding this value will result in an error too. Other than that, I think its ready to be merged.

I'll commit the other widgets by creating a new PR then.

realrk69:
When I try to create a branch, it says that I need write access to the repository. Should I just create a fork then?

borondics:
Hey @realrk69,

This doesn't really belong here in this stream, but once again, I am happy that you are contributing to Orange Spectroscopy / Quasar. Thanks! In case you are interested in discussing further, please drop me an email and we can connect there: borondics .at. gmail.

Cheers,
Feri

markotoplak:
@realrk69, create a branch in your fork (you have one already, otherwise you couldn't open a PR).

markotoplak:
@realrk69, and what did you mean to plot? In any case, it would be best to use pyqtgraph, because such graphs can be interactive and all of Orange is using it. But if plotting is not a central functionality of the widget, it might be better to just output data you want to plot and then use other widgets for plotting.

realrk69:
@markotoplak The Variable Selection widget outputs the attributes data.X which have most relevance to predicting targets data.Y. The plot should consist of the original data with these relevant data highlighted through colours. Do you think that this can be done with any existing widget?

@borondics Are you refering to the PR or to the comments about forking 😄? I'll get in touch  via mail soon.

borondics:
No, I was referring to my own message to discuss via email. :)

realrk69:
Here's how I'm displaying data using the pre-installed pyqtgraph OWSpectra widget:
```
def __init__(self):
    self.plot = OWSpectra()
    self.plot.set_data(self.data)
    self.plot.handleNewSignals()
    self.mainArea.layout().addWidget(self.plot)
```

Then to change the data input in the OWSpectra widget and show the plot:
```
    self.plot.set_data(self.new_data)
    self.plot.handleNewSignals()
```

The other way is to initialize an instance of pg.PlotCurveItem which I can customize. 
The function I want to access is [LinearRegionItem](http://www.pyqtgraph.org/documentation/graphicsItems/linearregionitem.html), which allows marking of horizontal or vertical regions.

Issue #416, opened 2020-04-15
s-SNOM FFT Processing


stuart-cls:
Continued from #415 
@levandoskije @raul-freitas @ok-nc @borondics

stuart-cls:
@levandoskije Is there some code I can look at for your processing? I'm trying to re-organize the FFT code to make it more maintainable and usable in scripts so now would be a great time to co-ordinate! I'm not that attached to what is there so if we can make something that is easy to read and correct I think it would be best to have a single widget for FFT.

levandoskije:
@stuart-cls  Yes there is! If you can follow the link ![Branch: complex_fft](https://github.com/levandoskije/orange-spectroscopy/tree/complex_fft).
It will be great to be able to do coordinated work with you.
Just a few small details:
- I'm not working on a new widget, I just added one more class to irfft.py;
- And I'm using data.attibutes to automate FFT, but that has made my class very specific.

stuart-cls:
@levandoskije Thanks! I will take a look at this and get back to you. New class is great, and I like the attributes idea, we had talked about doing that as well (#272) and I think more automation where appropriate is helpful.

levandoskije:
@stuart-cls, I write here to bring the information together and keep everyone informed.
I will divide the work #423 into two parts, as you suggested, the next Pull Request it will include:

 - Some tests
 - Check_metadata()
 - FFT as an inherited class

ok-nc:
Hey guys, great work so far! 

@stuart-cls, one quick note about peak search: there's only a couple commercial s-SNOM instruments, but none do any zpd bookkeeping (at least to my knowledge), nor do most home-built setups, as the user usually has full/arbitrary control of the reference arm scan range. So a peak search for the intfgm max is always necessary, though not ideal for large batch processing. Search off by default when zpd present in metadata is a good strategy.

stuart-cls:
@levandoskije Fantastic, thanks!

@ok-nc That's good to know. I thought I had read somewhere about determining the zpd with a fixed mirror but that might have been a one-off home-built thing. For the batch processing, the current code only lets you peak search if you do it row-by-row. However, for datasets with weak signals but common zpd, it's better to determine it once for the group and then set it manually to avoid zpd "noise".

raul-freitas:
[...]

PR #423, opened 2020-04-29
[ENH] s-SNOM Complex FFT Processing

levandoskije:
Enhancement for FFT Processing as mentioned by @stuart-cls in issue #416 .
A new method for complex FFT with GSF files, read by Reader #392 .

-  [x] Processing complex interferograms
-  [ ] Tests for complex processing results

### DIFF
--- a/orangecontrib/spectroscopy/irfft.py
+++ b/orangecontrib/spectroscopy/irfft.py
@@ -280,75 +280,33 @@ class IRFFT():
     
     def complex_fft(self, ifg, zpd=None, phase=None, info=None):
 
-        # Numeros de pontos do interferograma
         number_of_points = int(info['Pixel Area (X, Y, Z)'][3])
-        # Distance do interferograma
         scan_size = float(info['Interferometer Center/Distance'][2].replace(',', ''))
-        # Interferometro de Michelson
         OPD = scan_size*2
-        # Auto explicativo
         step_size = OPD/(number_of_points - 1)
         self.dx = step_size
 
-        # choose_data = ifg
-        # zf = 2**4
+        ifg -= np.mean(ifg)
 
-        # data_fft = np.fft.fft(choose_data, n=zf*number_of_points)
-        # magnitude = np.abs(data_fft)
-        # angle = np.angle(data_fft)
-
-        # Wmax = ((number_of_points-1)/(2*OPD))*10**4
-        # wavenumber = np.linspace(0, Wmax, int(len(data_fft)/2))
-
-        # self.spectrum = magnitude
-        # self.phase = angle
-        # self.wavenumbers = wavenumber
-        #####################################################################################################
-        # Stored phase
-        self.phase = phase
         # Stored ZPD
         if zpd is not None:
             self.zpd = zpd
         else:
             self.zpd = find_zpd(ifg, self.peak_search)
 
-        # Subtract DC value from interferogram
-        # ifg = ifg - ifg.mean()
+        ifg = apodize(ifg, self.zpd, self.apod_func)
+        ifg = zero_fill(ifg, self.zff)
+        # Rotate the Complete IFG so that the centerburst is at edges.
+        ifg = np.hstack((ifg[self.zpd:], ifg[0:self.zpd]))
 
-        # Calculate phase on interferogram of specified size 2*L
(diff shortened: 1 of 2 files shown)


levandoskije:
@markotoplak and @stuart-cls, I tried  make `git rebase master`, but I has some difficult and I am afraid broken the code.
Any suggestions?

stuart-cls:
Hi @levandoskije 
I've started looking at this today. Don't worry about updating it so it can be merged for now.

stuart-cls:
First, thanks for this contribution. I appreciate that you've worked hard to incorporate your calculations into the very convoluted mess that was already there.

In general, I would like to use this opportunity to improve the calculations / code simplicity for everyone. I understand that you've probably been using this code in your work already, and want the calculations to stay identical. So I think the first and most important step is to define some tests. I propose storing the spectral output of the current code run against the test data and checking any changes against that. But if you have a different test you would prefer, that's fine too.

I'm going to get into the details now, but what I'd like to do is split these changes up into separate commits/pieces. Here are the how I see the changes:

## 1. General structure
A. Setup calculation based on input data
  - This is where we look at the dataset and prepare the widget for calculation, but don't actually do the calculation. Any code that changes calculation parameters or GUI values must be done here. I think everything like this in this PR should go in OWFFT.check_metadata(). It can be restructured as required.

B. Prep Data 
  - This is where we go from Orange.Table to something IRFFT can process. I want to separate this out, but currently this is done in OWFFT.calculateFFT()
    - Currently this handles things like double-sweep data, stored phase, and multirow processing
    - This is where your amplitude+phase combination should go (as you've done)

C. FFT
   - The actual FFT. I'd like to make your complex version an inherited class instead of a different function, but the devil is in the details, which is most of what the rest of this is about.
 
## 2. dx (step size) calculation

 - Do this once, in check_metadata() so users can see the value before calculating
 - This should output the value in cm, not microns
 

## 3. Data Rearrangement (combine amplitude and phase)

A one-liner that does this:
````
    sh_data = self.data.X
    full_data = sh_data[::2] * np.exp(sh_data[1::2] * 1j)
````	
Also, I wonder if it makes more sense to change the arrangement of the data
to be less fragile: the current code requires the data to always be carefully ordered.
We already have the case of dual-sweep interferograms which are split and then processed, I propose building the table as np.hstack(amplitude, phase) and then combining in the calculation. This keeps the paradigm of one row, one sample. We can do this in a separate PR though.

## 4. Peak search / zero-fill

 - Why re-implement this? I'd rather use a single code path unless you have a strong reason not to.
 - The code should respect the user's zff setting. If you must, set the default to 4 in your check_metadata step.
 - Excuse my ignorance, but doesn't s-SNOM use an externally determined zpd value anyway? Peak search is always fragile for weak signals, I'm aiming to move away from it as default when a known value is already present in the metadata.
 - This is not really zero-filling so much as throwing away the low-res side of the interferogram. What the existing code does instead is rotate the array so that the zpd is at the edges, which also gives a flat phase response. Comment?
 
## 5. Apodization / windowing

 - Any particular reason to use the scipy implementation here? I've tried to keep this module numpy-only and we already have an implementation which respects assymetric interferograms (very common in FT-IR as well). If there is some problem with the implementation, we should fix it instead of having two code paths
 
## 6. np.fft.fft vs np.fft.rfft

 - This is obviously the main difference. The only comment here is that using n to zero-fill here instead of the existing code doesn't gaurentee a DFT-efficient array size, which is why I don't use that.

## 7. magnitude / angle

 - This is just a different way of outputting the complex result of the fft, so I'd rather fold it in with that code and make it general. Someone else might like magnitude/angle output for their FTIR data as well.

## 8. Wavenumber calculations

  - Why not just use np.fft.fftfreq(Nzff, self.dx)?

Again, thanks for your hard work so far. I would suggest that you take the current master and add your code with this feedback on to that manually. It will be much easier than trying to rebase and then rearrange. If this is too much work / you don't have time, I can do it.

If you really just want to get this in as close to your existing code as possible, please just focus on suggestions 1-3 and we can discuss unifying the code in for steps 4-7 in a separate PR.

levandoskije:
Hi, @stuart-cls,  thank you for the review!
I will close this PR and do another one with the suggestions above.
The discussion is here #416.

PR #425, opened 2020-05-01
Fixed errors with PLS

realrk69:
Errors:

1. Components would sometimes raise an error in which they cannot accept floats.

2. The max number of components (maxv) is equal to len(self.data.domain.attributes). Otherwise, it will raise error, invalid number of components. So this will prevent user from setting it to more than the number of features.


stuart-cls:
Thanks for your contribution!

I have only two small comments:

1. You may want to use the tools in `orangecontrib.spectroscopy.widgets.gui` to solve this problem to avoid code duplication
2. Please use the topic branch method of git, roughly:
````
(master) git pull  # to ensure you are on the upstream master, assuming your master branch
                   # tracks upstream and not your origin fork
(master) git checkout -b my-int-pls-fix
(my-in-pls-fix) git commit  # etc
(my-in-pls-fix) git push origin my-in-pls-fix
````
This ensures only your relevant changes are in the PR, and not other unrelated commits, and makes sure your changes are on the latest code.

See https://github.com/Quasars/orange-spectroscopy/blob/master/CONTRIBUTING.md and specificially the https://help.github.com/articles/fork-a-repo/ link.

realrk69:
@stuart-cls Hey, I have a couple of questions:

1. What is the function to define maxv in gui.spin externally? I can't seem to find it in any documentation. Can you point me to the method or function you're referring to?

2. How do I create a new branch in this repo? Should I delete the old fork and create a new branch here only? Or continue in that fork? I'm not familiar with Git.

stuart-cls:
1. Sorry, I didn't look close enough at your code, there's no gui.spin implementation, just lineedit. 

2. I think you'll find [github-fork-a-repo](https://help.github.com/en/github/getting-started-with-github/fork-a-repo) really explains the concepts nicely, with detailed examples. See especially "Keep your fork synced", [Step 3: Configure Git to sync your fork with the original Spoon-Knife repository](https://help.github.com/en/github/getting-started-with-github/fork-a-repo#step-3-configure-git-to-sync-your-fork-with-the-original-spoon-knife-repository)
 Briefly, in the same manner in which you created this https://github.com/realrk69/orange-spectroscopy:PLS branch, create a new one with only this fix based on Quasars/orange-spectroscopy:master.
And yes, keep your work in your fork.

Once you have a working branch you will need to point this PR at it:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

markotoplak:
Apart from what @stuart-cls suggested, this PR does not fix the bug completely and introduces a new bug:

- Unfixed bug: the PLS widget also outputs a Learner, which should work with any new data (which can have, for example, only one column ). So that problem needs to be also fixed in the PLS learner. 

- New bug: the PLS widget should also work without any data on input and should allow setting any number of components, because the Learner on its output could go to arbitrary data.

Also, it would be good to have tests for the fixed functionality.

markotoplak:
@realrk69, I think the repository mixup happened because you started your from your personal master branch and not from the master branch of the official repository.

Yes, git can be a pain to learn sometimes... I understand this is not easy when you start working with it. So, you should have at least two remotes registered: the official repository (quasars/orange-spectroscopy) and yours. Then, when you start working on a new feature, you start from the official repositories master branch and, before making a PR, upload into your repository.

If your are not that familiar with git, the easiest solution might be to simply copy the files you'd like to change somewhere safe, delete your repository, and start over.

realrk69:
@markotoplak I kinda understand now. I will try to fix the git asap.

Have fixed the new bug, instead of setting the max components on None data (no input) to 1, it now sets it to the default value defined in the gui.spin.

Also, could you elaborate a bit on the unfixed bug? I've tried different combinations (data with one feature column and one target variable column), and it works on Test and Score Widget and the Rank widget which both use scorer or learner. If I input data with only one column in test data, then it scales the input to match training data, which already generates a warning.

PR #426, opened 2020-05-01
PLS: Update setting max components and int-float bug

realrk69:
Hope this is the correct method of using git 🤦‍♂️ 

Max number of components without input data is now 100
@markotoplak @stuart-cls

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpls.py
+++ b/orangecontrib/spectroscopy/widgets/owpls.py
@@ -1,4 +1,5 @@
 from AnyQt.QtCore import Qt
+from AnyQt.QtWidgets import QSpinBox
 import scipy.sparse as sp
 
 from Orange.widgets import gui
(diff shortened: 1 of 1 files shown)


stuart-cls:
Git looks good :) Unfortunately, it seems the "parameter" test is failing from this change.

I'll leave the review of the bug issues raised in the last PR to @markotoplak

markotoplak:
Now, one bug is fixed.

The other bug is this. PLS was set to 50, data set has 12 features. PLS raises ValueError, which (thankfully) Test widget handles and displays.

Screenshot 2020-05-07 at 07 56 42

Data dimension should be checked before passing it to sklearn (that means after Orange's preprocessing) and number of components lowered accordingly. Perhaps see class PCA from Orange.projection.pca and implement something functionally similar.

Also, can you explain why did you add `spinType=QSpinBox` to your spin boxes? On my computer they already return ints previously.

realrk69:
I'll try to fix the other bug. I think I understood how to. Will update this.

I tried the widget with different input-output combinations and some particularly gave this error constantly. I'll try to reproduce the error and post a screenshot. But this fix resolved it. It showed that n_components value returned as 11.0 instead of 11, which after trying manually fitting a model gives the same error outside orange.

markotoplak:
@realrk69, I tried to fixed "the other bug" in #449, which also takes care of problems in the canvas. So I am closing this PR.

I do not understand you you had int/float problems. Are you sure this was not a remnant of saved files with your test versions or some stale widget settings? Do you still get problems on new files with cleared settings (Options->Reset widget settings)?

Issue #415, opened 2020-04-14
FFT Meta Issue


stuart-cls:
This issue is for communicating plans for the FFT module and tracking github issues. 
I'll try to keep it up to date.

**API Boundary**
In order to make the `irfft` module re-usable for scripts (even outside Orange), it has minimal dependencies and no knowledge of Orange Table data structures.
However, the actual implementation is overly complex and leaves too much of that complexity in the GUI OWFFT side.

- [ ] Move last bits of data wrangling out of OWFFT back into irfft
- [ ] Simplify irfft calculation code to make it more readable/clear, esp phase calculation
- [ ] Add preprocessor-style layer for use in Orange scripts. May or may not actually be a preprocessor but should accept an Orange.Table like a preprocessor

**GUI / User-friendliness**
Mostly small things, except responsive changes which will build on the API fixes above.

 - [ ] Make responsive / cancel-able as in #398 and friends
 - [ ] Load and use stored zpd values (#414)
 - [ ] Provide user feedback on zpd (#250)
 - [ ] Handle (and test) multi-spectra stored phase / zpd condition #(253)
 - [ ] load / use parameters from Table attributes instead of metadata (#272)
 - [ ] Documentation (#254)

**s-SNOM**
People would like to do FFT on s-SNOM data. There are just a few missing pieces:

 - [ ] Happ-Genzel apodization
 - [ ] Hann apodization
 - [ ] Subtraction phase correction
 - [ ] Example workflow

**Speed**
Some progress (5-10x) was done in #413 but true multi-core processing is still not implemented

 - [ ] Multiprocessing (#360)

levandoskije:
Hey, @stuart-cls , I'm working on a complex FFT to process NeaSNOM data with synchrotron radiation.

raul-freitas:
Hello @stuart-cls, @levandoskije  above is my student and we were about to contact you about our development on the FFT module for s-SNOM data. One of the main points is that s-SNOM is intrinsically complex so we need to apply a FFT instead of a Re-FFT. We already implemented the in the Multiple widget the reading of Amplitude and Phase of the s-SNOM data. Now, we have a quasi-finished version of the FFT widget for this class of data.

borondics:
Hi @raul-freitas, @levandoskije, and @stuart-cls,

Since we don't have the luxury of a few hundred programmers I think it would be great to coordinate the work everyone does not to duplicate code. This should however not prevent us to review and rewrite parts that are not up to requirements.

[...]

PR #429, opened 2020-05-04
[FIX] build_spec_table: respect additional_table.attributes

stuart-cls:
`NeaReaderGSF` added attributes to _additional_table_, but `build_spec_table` didn't include them in the final table output.

Includes test :)

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -1005,7 +1005,8 @@ def build_spec_table(domvals, data, additional_table=None):
                         class_vars=additional_table.domain.class_vars,
                         metas=additional_table.domain.metas)
         ret_data = Table.from_numpy(domain, X=data, Y=additional_table.Y,
-                                    metas=additional_table.metas)
+                                    metas=additional_table.metas,
+                                    attributes=additional_table.attributes)
         return ret_data
 
 
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -1130,11 +1130,15 @@ class NeaReaderGSF(FileFormat, SpectralFileFormat):
 
         info = {}
         for row in parameters:
-            info.update({row[0].strip(':'): row[1:]})
+            key = row[0].strip(':')
+            value = [v for v in row[1:] if len(v)]
+            if len(value) == 1:
+                value = value[0]
+            info.update({key: value})
         
         info.update({'Reader': 'NeaReaderGSF'}) # key used in confirmation for complex fft calculation
 
-        averaging = int(info['Averaging'][1])
+        averaging = int(info['Averaging'])
         px_x = int(info['Pixel Area (X, Y, Z)'][1])
         px_y = int(info['Pixel Area (X, Y, Z)'][2])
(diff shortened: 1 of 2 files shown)


stuart-cls:
Actually, hold off on this. All the tests pass but it crashes OWFile...

stuart-cls:
Okay, the issue is `OWFile` special-cases keys `Name` and `Description` and makes assumptions about their content. I'll add a fix for the NeaReaderGSF (the only FileFormat using this facility so far)

stuart-cls:
@levandoskije is there any particular reason NeaReaderGSF returns these type of fields as 
````
'Project': ['', 'Spectroscopy datasets']
'Description': ['', 'Ref NeaReaderGSF_test']
````
instead of 
````
'Project': 'Spectroscopy datasets'
'Description': 'Ref NeaReaderGSF_test'
````

**Comment by levandoskije** [comment]
@stuart-cls, in this two cases don't, but for example:
```
 'Scanner Center Position (X, Y)': ['[µm]', '65.31', '42.72', ''],               
 'Rotation': ['[°]', '0', '', ''],                       
 'Scan Area (X, Y, Z)': ['[µm]', '0.000', '0.000', '0.000'],                    
 'Pixel Area (X, Y, Z)': ['[px]', '1', '1', '1024'],                     
 'Interferometer Center/Distance': ['[µm]', '670.000', '1,000.000', ''],                   
 'Averaging': ['', '1', '', ''],                  
```
There is more than one information, so I need a list or a tuple.
Was that a question?

stuart-cls:
@levandoskije  Yes, thanks. For the 'Averaging' case, could that be "1" instead of a list? I'm thinking of stripping the list if there's only one value. If you'd rather I leave it, I'll just special case "Name" and "Description".

levandoskije:
@stuart-cls , I think this great!
But need to do some changes in FFT for working correctly.

borondics:
Guys, sorry to intercept this here. Could you also write documentation for the FFT widget and all the old and new features? Currently, it is not too informative I think... :)

stuart-cls:
@levandoskije One gotcha is "Averaging" is used to build the table itself. Are there files that have multiple values for this or is always a single number?

@borondics Sorry, but it's difficult to discuss multiple things. Please, let's try to stay on topic. Documentation of FFT is discussed in #254 and collected in #415

levandoskije:
@stuart-cls, In the last 9 months I worked with this data, I’ve never seen more than one number, but I can’t say it’s always like that.
@raul-freitas, I think it's just a single number. What do you think?

stuart-cls:
Okay, I did it. We'll wait for the okay before merging.

raul-freitas:
> @stuart-cls, In the last 9 months I worked with this data, I’ve never seen more than one number, but I can’t say it’s always like that.
> @raul-freitas, I think it's just a single number. What do you think?

The **Averaging** field is always a single quantity or single number. It represents the number of interferograms (scans or accumulations) acquired, no matter it is a single point spectra, linescan or hyper spectral map.

stuart-cls:
@raul-freitas Great, thanks for replying!

@markotoplak This is good to go unless you have any comments

markotoplak:
Thanks, @stuart-cls.

Issue #432, opened 2020-05-07
Min-max normalization


markotoplak:
We do not have it. People want it.

borondics:
I think the best would be to add a new "integral" into the "Area" normalization.

This would also mean that we need to rename the method in the Preprocess Spectra and also implement something new because it actually doesn't make sense to have the "X-value of..." options there.

![image](https://user-images.githubusercontent.com/6304794/81346160-1773c000-90ba-11ea-86fc-c0bb6d775bcd.png)

If the IntegrateFeature____ methods in integrate.py could have a flag where to include them it would be great. Maybe we could redo the Integration to be more general and call it Spectrum Calculations or something like that.

stuart-cls:
The "Area" normalization type is for normalizations that require setting a window. Maybe I missed the point, but isn't Min/Max just Min value set to 0, Max value set to 1 of the entire spectrum?

On your point about naming things, I very much agree the "Integrate" name is well past its best before date :) The name I've been floating around in my head is more like "Peak Analysis" to suggest calculations that occur in a specific window (or "peak" :) )

Generally I think it's best to keep the single list of Integrate/Peak Analysis calculations so that when they show up in Preprocess/Normalize/HyperSpectra/Integrate etc they are consistent.

borondics:
0-1 could be defined for either the entire spectrum or a specific region. I guess it is up to us but both make sense. The region could be also done by using a _Cut_ before.

I think "Peak Analysis" could be good if we used it only for that but even in this case we are using it for Normalization. 

For showing the whole list of those calculations (especially if we add others that would act on each spectrum separately) the issue is that sometimes they make no sense in a particular context, like for example the _Normalize / Area Normalization_ using _X-value of maximum from 0_

stuart-cls:
Okay I get it now. I've actually had requests to do the vector normalization over a window as well so I'll just move the limits to be more generally applied.

I agree with your other comments :)

markotoplak:
Done in #437

PR #435, opened 2020-05-08
Added Min-Max 0->1 normalization

borondics:
Implements #432 

I have a feeling that this is not very elegant, but it seems to work just fine. @markotoplak, what do you think?

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -312,6 +312,12 @@ class _NormalizeCommon(CommonDomain):
                                   limits=[[self.lower, self.upper]])(data)
             data.X /= norm_data.X
             replace_infs(data.X)
+        elif self.method == Normalize.MinMax:
+            min_data = np.min(data.X, axis=1)
+            max_data = np.max(data.X, axis=1)
+            norm_data = (data.X.T - min_data) / (max_data - min_data)
+            replace_infs(norm_data)
+            data.X = norm_data.T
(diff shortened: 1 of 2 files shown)


stuart-cls:
Ha I also thought Friday was a good time to do this :) Mine handles Nans but is other wise exactly the same https://github.com/stuart-cls/orange-spectroscopy/tree/minmax-normalization

stuart-cls:
I'm just going to open a PR to see if mine passes Tests, mostly for personal curiosity :)

orangecontrib/spectroscopy/preprocess/__init__.py
```diff
@@ -312,6 +312,12 @@ def transformed(self, data):
                                   limits=[[self.lower, self.upper]])(data)
             data.X /= norm_data.X
             replace_infs(data.X)
+        elif self.method == Normalize.MinMax:
+            min_data = np.min(data.X, axis=1)
+            max_data = np.max(data.X, axis=1)
+            norm_data = (data.X.T - min_data) / (max_data - min_data)
```
stuart-cls, line 318: Why all the transposing?

borondics: Otherwise, the sizes are not compatible.

stuart-cls: Oh I see. You can also pass `keepdims=True` to min/max for the same effect.

markotoplak:
Thank you both! :)

PR #436, opened 2020-05-08
Added SNV normalization

borondics:
At the request of users from Spain.

Standard Normal Variate is calculated by mean centering the spectra and subsequently dividing each with its standard deviation.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/preprocessors/normalize.py
+++ b/orangecontrib/spectroscopy/widgets/preprocessors/normalize.py
@@ -27,6 +27,7 @@ class NormalizeEditor(BaseEditorOrange):
         ("Min-Max Normalization", Normalize.MinMax),
         ("Area Normalization", Normalize.Area),
         ("Attribute Normalization", Normalize.Attribute),
+        ("Standard Normal Variate (SNV)", Normalize.SNV),
         ("Normalize by Reference", NORMALIZE_BY_REFERENCE)]
 
     def __init__(self, parent=None, **kwargs):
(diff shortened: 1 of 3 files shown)


borondics:
The button "pressed" in the test for the "Normalize by Reference" feature is hardcoded. First, it failed and when I changed it it passes. Can we handle this better in the test?

stuart-cls:
See my minmax implementation for a "solution" to the hardcoded button

borondics:
@stuart-cls if you think this is OK please merge it to the master.

stuart-cls:
Can you add a simple test of the calculation? Also there's a rogue "print" function in one of the files :)

markotoplak:
I have squashed all your commits into one. Next time use rebase instead of merge when updating your branch. Merges within branches make history needlessly complicated (and your merged twice).

orangecontrib/spectroscopy/widgets/preprocessors/normalize.py
```diff
@@ -27,6 +27,7 @@ class NormalizeEditor(BaseEditorOrange):
         ("Min-Max Normalization", Normalize.MinMax),
         ("Area Normalization", Normalize.Area),
         ("Attribute Normalization", Normalize.Attribute),
+        ("Standard Normal Variate (SNV)", Normalize.SNV),
```
stuart-cls, line 30: Is it redundant to say "(SNV)" here? I guess users will know (hopefully) what it is?

borondics:
@stuart-cls I wouldn't count on them knowing it. :)
I didn't know what it means before I looked it up.

stuart-cls:
I just meant having both the name and the acronym :)

stuart-cls:
Okay, good news. We save the Normalization method as the setting, not the radio button position (duh I guess, good job past us :) ) so no problem with the settings.

markotoplak:
@stuart-cls, today I also thought of the same settings problem and then spent an hour looking for how could our settings ever work before the changes. Then I saw that everything is fine, we just add ids to radio buttons. Seems like you went through a similar process. :)

borondics:
Well, we could drop the long name, but then we really have to start writing
documentation for it! :)

PR #437, opened 2020-05-08
[ENH] Normalize: add Min-max normalization

stuart-cls:
Part of #432

Doesn't implement ranges yet.

Clone of #435


borondics:
Ranges could be done by a simple Cut before the Normalize...

borondics:
I like your better than mine. :)

Didn't know about the keepdims switch...

PR #438, opened 2020-05-12
Normalize, MinMax: no shifting, only division by range

markotoplak:
That makes the range=1 and also works for derivatives, when can be negative. This should be similar behaviour as in Unscrambler.

Thanks @clsandt.


markotoplak:
@stuart-cls, what do you think?

stuart-cls:
I like this, more modular. User can always combine with baseline subtraction to get min=0. Good catch @clsandt

PR #439, opened 2020-05-13
Integrate: Add "data" output which always contains spectra + results in meta

### LINKED ISSUE #310: OWIntegrate: Split into two outputs?
Thinking about #307, I wonder if it would be a good idea to split the output into Spectra + Peak Parameters[1] and get rid of the "Output as metas" checkbox altogether. I think this is more consistent with what most Widgets do. 

The only thing I'm unsure of is if this would result in extra memory usage in the "Output as metas" case.

[1] ...would need a better name...

stuart-cls:
Fixes #310

Rationale: basically exactly the same as https://github.com/biolab/orange3/pull/4255, I often find myself needing both the total spectra + metas and just the reduced data, this saves setting up two widgets.

I decided to leave the "output_metas" checkbox alone because someone might be depending on the useful control of the output preprocessor chain (in TileFile, for example).

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owintegrate.py
+++ b/orangecontrib/spectroscopy/tests/test_owintegrate.py
@@ -91,6 +91,23 @@ class TestOWIntegrate(WidgetTest):
         self.assertEqual(1, len(preprocessed.domain.attributes))
         self.assertEqual(0, len(out_data.domain.metas))
 
+    def test_output_data_with_metas(self):
+        data = Orange.data.Table("iris.tab")
+        self.widget.output_metas = True
+        self.send_signal(self.widget.Inputs.data, data)
+        self.widget.add_preprocessor(self.widget.PREPROCESSORS[0])
+        self.widget.unconditional_commit()
+        out_tdata_t = self.get_output(self.widget.Outputs.preprocessed_data)
+        out_data_t = self.get_output(self.widget.Outputs.data)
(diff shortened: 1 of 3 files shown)


stuart-cls:
Re: tests, something is broken in master as well. Related to `_get_skl_params` from Orange?

markotoplak:
Tests broke on Orange with the new release of scikit-learn. We are planning to pin it to 0.22 for now.

markotoplak:
If you understand this PR:
- adds a new signal that always does what "Output as metas" used to do.
- keeps the old signal.

So if "Output as metas" is set, then they are equal, right?

What do you think about removing "Output as metas" if a new widget instance is added to the plot and only keeping it in a compatibility mode, when old workflows are loaded? Similarly to what was done in biolab/orange3#4810.

And "Data" should probably be the default output. 

If you agree but lack the time to do it, I can add that.

stuart-cls:
Yes, what you describe is correct and your suggestion would be my preferred user interface.

I didn't do it at first because I wasn't sure how to do the transition. I also later realized that consumers of the `PreprocessorList` output might be using that setting. That said, the most useful application of the `PreprocessorList` output is data reduction either in TileFile or before a model so I think (without anyway to be sure) that it would mostly be used with `PreprocessorListMoveMetas.move_metas=True`.

If you could add it I would appreciate it :)

stuart-cls:
As discussed in #638 , closing this. Can reopen if necessary later.

PR #443, opened 2020-05-20
[ENH] HyperSpectra: display visible image

### LINKED ISSUE #118: What to do with visible images
Has anyone done much thinking about this? A common request is to overlay the visible image with the infrared/other image but this requires at least:

  1. Some standard way of attaching the visible image to the Table/group of spectra
  2. GUI overlay, probably in HyperSpectra

This came up in #117

pisarik:
Hi!

Sorry, I didn't start discussion in issues, since we discussed it at meetings.

Here is the following changes:

- Save all visible (microscopic) images (bytes, title, position and pixel size) from opus file into `data.attributes: dict[List[dict]]`;
- Bytes unpacked into `np.ndarray` with pillow package just before displaying it in `imageplot` (to save memory);
- Visible image can be selected from combo box;
- User can change opacity of visible image;
- User can select composition mode (Normal, Overlay, Multiply, Difference);

I don't have a good lightweight file to put in tests for Opus. The one I have weights 150 MBs

Closes #118

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -775,14 +775,17 @@ class OPUSReader(FileFormat):
 
         visible_images = []
         for img in opusFC.getVisImages(self.filename):
-            visible_images.append({
-                'name': img['Title'],
-                'image_bytes': img['image'],
-                'pos_x': img['Pos. X'] * img['PixelSizeX'],
-                'pos_y': img['Pos. Y'] * img['PixelSizeY'],
-                'pixel_size_x': img['PixelSizeX'],
-                'pixel_size_y': img['PixelSizeY'],
-            })
+            try:
+                visible_images.append({
+                    'name': img['Title'],
+                    'image_bytes': img['image'],
+                    'pos_x': img['Pos. X'] * img['PixelSizeX'],
+                    'pos_y': img['Pos. Y'] * img['PixelSizeY'],
+                    'pixel_size_x': img['PixelSizeX'],
+                    'pixel_size_y': img['PixelSizeY'],
+                })
+            except KeyError:
+                pass
 
         domain = Orange.data.Domain(attrs, clses, metas)
 
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -773,6 +773,17 @@ class OPUSReader(FileFormat):
                 else:
                     meta_data = params
 
+        visible_images = []
+        for img in opusFC.getVisImages(self.filename):
+            visible_images.append({
+                'name': img['Title'],
+                'image_bytes': img['image'],
+                'pos_x': img['Pos. X'] * img['PixelSizeX'],
+                'pos_y': img['Pos. Y'] * img['PixelSizeY'],
+                'pixel_size_x': img['PixelSizeX'],
+                'pixel_size_y': img['PixelSizeY'],
+            })
+
         domain = Orange.data.Domain(attrs, clses, metas)
 
         meta_data = np.atleast_2d(meta_data)
@@ -780,6 +791,9 @@ class OPUSReader(FileFormat):
         table = Orange.data.Table.from_numpy(domain,
                                              y_data.astype(float, order='C'),
                                              metas=meta_data)
+
+        if visible_images:
+            table.attributes['visible_images'] = visible_images
 
         return table
 
--- a/orangecontrib/spectroscopy/tests/test_readers.py
+++ b/orangecontrib/spectroscopy/tests/test_readers.py
@@ -1,5 +1,7 @@
 import unittest
 from unittest.mock import patch
+from io import BytesIO
+from PIL import Image
 
 import numpy as np
 import Orange
@@ -68,6 +70,39 @@ class TestDat(unittest.TestCase):
             d1.save(fn)
             d2 = Orange.data.Table(fn)
             np.testing.assert_equal(d1.X, d2.X)
+
+
+@unittest.skipIf(opusFC is None, "opusFC module not installed")
+class TestOpusReader(unittest.TestCase):
+
+    def test_no_visible_image_read(self):
+        d = Orange.data.Table("opus/no_visible_images.0")
+
+        # visible_images is not a permanent key
+        self.assertNotIn("visible_images", d.attributes)
+
+    def test_one_visible_image_read(self):
+        d = Orange.data.Table("opus/one_visible_image.0")
+
+        self.assertIn("visible_images", d.attributes)
+        self.assertEqual(len(d.attributes["visible_images"]), 1)
+
+        img_info = d.attributes["visible_images"][0]
+        # decompress bytes only in widgets to reduce memory footprint
+        self.assertEqual(type(img_info["image_bytes"]), bytes)
+        self.assertEqual(img_info["name"], "Image 01")
+        self.assertAlmostEqual(img_info["pixel_size_x"], 0.90088498)
+        self.assertAlmostEqual(img_info["pixel_size_y"], 0.89284902)
+        self.assertAlmostEqual(img_info["pos_x"],
+                               43552.0 * img_info["pixel_size_x"])
+        self.assertAlmostEqual(img_info["pos_y"],
+                               20727.0 * img_info["pixel_size_y"])
+
+        # test image
+        with BytesIO(img_info["image_bytes"]) as f:
+            img = Image.open(f)
+            img = np.array(img)
+            self.assertEqual(img.shape, (538, 666, 3))
 
 
 class TestHermesHDF5Reader(unittest.TestCase):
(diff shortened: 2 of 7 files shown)


orangecontrib/spectroscopy/data.py
```diff
@@ -700,6 +700,17 @@ def read(self):
...
+        for img in opusFC.getVisImages(self.filename):
+            visible_images.append({
+                'name': img['Title'],
+                'image_bytes': img['image'],
+                'pos_x': img['Pos. X'] * img['PixelSizeX'],
+                'pos_y': img['Pos. Y'] * img['PixelSizeY'],
+                'pixel_size_x': img['PixelSizeX'],
+                'pixel_size_y': img['PixelSizeY'],
+            })
+
```
markotoplak, line 713: @stuart-cls, is this safe? Do images always contain these attributes or should we wrap in try ... except?

markotoplak:
Thanks! Seem like a quality code, I will check more in depth if the settings (saving, loading, changing data sets) are handled as in other widgets.

Please rebase to recent master. I have just updated travis to use correct versions of scikit-learn so that the tests will run.

Regarding huge data files and testing. Well, this is only a problem when testing file readers. When testing HyperSpectra, you can always add an image manually by setting appropriate .attributes, and this is for testing, I think, even better. So there is no need to read an opus file when testing HyperSpectra.

borondics:
Hi @pisarik, thanks for this contribution it will be very useful. Could you also upload somewhere (Dropbox or similar) a few FPA images that we could use for human testing. We will also get data from other instruments and that should make sure that everything works fine. Also, if you could add a very small FPA file to do testing, it could be useful... or @markotoplak, should that be for opusFC?

orangecontrib/spectroscopy/tests/test_readers.py
```diff
@@ -70,6 +72,37 @@ def test_roundtrip(self):
             np.testing.assert_equal(d1.X, d2.X)
 
 
+class TestOpusReader(unittest.TestCase):
```
markotoplak, line 75: Watch out with OPUS files, always add  `@unittest.skipIf(opusFC is None, "opusFC module not installed")` and also test without opusFC installed, because OpusFC is an optional library.

That is another reason to remove opus files from Hyper tests.

pisarik: Agree that it is unnecessary to use opus files to test HyperSpectra. Since I store image bytes (jpeg or bmp) in attributes it is a little bit difficult to come up with dummy image bytes.

I can mock `Image.open` everywhere as it done in `test_select_another_visible_image`. Do you think it is the way to do it?

pisarik:
Hi @borondics! I've just sent to your email a link with the FPA file.

I asked to make a special measurement to obtain lightweight dummy Opus file for tests. But at this time I don't have it.

I also have a code to extract visible images from Agilent files. But I need to have more test files, since I hardcoded few parameters that I couldn't find in Agilent file. It perfectly works for the files that I have, but I would like to check it with Agilent files from other sources. Do you have Agilent FPA Image files?  It will greatly help me to extract visible image more reliably.

orangecontrib/spectroscopy/tests/test_readers.py
markotoplak: Well, I'd prefer not to mock it Image.open, I would rather mock the data. You could perhaps add a small image to the repo and then read that one as bytes. Or, if the image is really small (few pixels would be enough anyway), you could base64 encode it and just leave it in the test .py file.

borondics:
Not sure if this is the reason, but HyperSpectra is broken for me using https://github.com/pisarik/orange-spectroscopy/tree/owhyper-show-visible-image and the FPA file @pisarik sent.

Here is the error message:
```
---------------------------- ValueError Exception -----------------------------
  File "/Users/borondics/Documents/code/Python/orange-canvas-core/orangecanvas/scheme/widgetmanager.py", line 378, in __process_init_queue
    self.ensure_created(node)
  File "/Users/borondics/Documents/code/Python/orange-canvas-core/orangecanvas/scheme/widgetmanager.py", line 326, in ensure_created
    self.__add_widget_for_node(node)
  File "/Users/borondics/Documents/code/Python/orange-canvas-core/orangecanvas/scheme/widgetmanager.py", line 242, in __add_widget_for_node
    w = self.create_widget_for_node(node)
  File "/Users/borondics/Documents/code/Python/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 296, in create_widget_for_node
    widget = self.create_widget_instance(node)
  File "/Users/borondics/Documents/code/Python/orange-widget-base/orangewidget/workflow/widgetsscheme.py", line 409, in create_widget_instance
    widget.__init__()
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owhyper.py", line 919, in __init__
    self.setup_visible_image_controls()
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owhyper.py", line 962, in setup_visible_image_controls
    callback=self.update_visible_image)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/gui.py", line 476, in comboBox
    return gui_comboBox(*args, **kwargs)
  File "/Users/borondics/Documents/code/Python/orange-widget-base/orangewidget/gui.py", line 1498, in comboBox
    callfront.action(cindex)
  File "/Users/borondics/Documents/code/Python/orange-widget-base/orangewidget/gui.py", line 1938, in action
    raise ValueError("Combo box does not contain item " + repr(value))
ValueError: Combo box does not contain item -1

```

pisarik:
- Rebased onto master
- TestOpusReaders: added `skipIf`
- TestVisibleImage: replaced `opus/one_visible_image.0` with mocked images (base64 strings in the file)

pisarik:
> Not sure if this is the reason, but HyperSpectra is broken for me using https://github.com/pisarik/orange-spectroscopy/tree/owhyper-show-visible-image and the FPA file @pisarik sent.

@borondics Sorry, I used old version of Orange to develop. This error doesn't happen in Orange 3.24.0, but it indeed appeared when I pulled latest orange3 master branch as well. I think, you can checkout to the 3.24.0 to test manually.

pisarik:
Either this is a bug in orange-widgets or I don't know how to correctly initialize empty combo box.

I've explored the problem: 
**My code commit b8b3fac**
I initialize `vis_img_combo` with empty `PyListModel` (this will be filled with image names when data loaded):
```python
        self.vis_img_name_model = PyListModel()
        self.vis_img_combo = gui.comboBox(
            self.visbox, self, 'cur_visible_image_idx',
            model=self.vis_img_name_model)
```
Since Qt by default returns index `-1` when combo is empty, I initialize setting with this value:
```python
    cur_visible_image_idx = Setting(-1)`
```

**Orange codebase**
The error happens here, likely because it doesn't check if the model is empty before try to activate:

https://github.com/biolab/orange-widget-base/blob/35df1f0814b85de565437b359e33d30244698b40/orangewidget/gui.py#L1492-L1498

orangecontrib/spectroscopy/data.py
stuart-cls: We should wrap. In opusFC I just directly copy the attributes from the file, so if Bruker has used different labels in the past/future, the name will change. Unfortunately, the visible images part is not documented so I had to do some guessing there.

stuart-cls:
Thanks for this!

About the Agilent files, I have a (very) old prototype of this overlay for Agilent data. When this PR is merged, I'd be happy to add the support for Agilent files based on this PR.

pisarik:
@stuart-cls do you mean to remove from this PR `TestOpusReader` test suite, but leave changes to OpusReader?

stuart-cls:
That is what I meant, but I'd rather have the tests than not. Let's just make sure we don't conflate the hyperspectra tests with OpusReader (I think @markotoplak already covered this point well).

I can reproduce the combobox error with Orange-master but it's something to do with ContextSettings I think so I'll leave that for @markotoplak :)

markotoplak:
@pisarik, I have rebased this to master and changed the combo box to also support Orange 3.25+ (well, I dropped support for Orange 3.24). Please check if the widget still works properly. I might have missed something...

If you plan to develop this further, please continue on this branch.

The tests that only worked with big images were made optional, so not they pass.

Before merging we still need to:
- [ ] Wrap image reading code for Opus files in try-except, because those parameters might not be there in all files
- [ ] Fix BW image problems I already told you about.

Do you perhaps have time to do these shortly? If not, no problem, just tell me, I'll try to do them today/tomorrow.

You wrote beautiful code with tests. Thank you for your hard work.

pisarik:
Fixed two mentioned above issues:

- [x] Try-catch around image reading code
- [x] Cast all images to RGBA format

I have few questions:
1. Should I make interactive rebase to embed those commits in history and then force-push?
2. There is still a little bug: 
2.1 Connect test image data to HyperSpectra
2.2 Create another test image data and connect it to HyperSpectra
2.3 Then Show visible image is unchecked, but controls are enabled
3. Should I do something about PyLint problems?

I will try to fix bug with reconnection it.

Thanks a lot, I learned from your review as well!:)

markotoplak:
Thank you!  No need to do an interactive rebase in this case.

I can wrap this one now. I will remove two lint errors and will interactively rebase then in (I see that I added at least two of them yesterday).

And I already see a solution to your bug.

PR #440, opened 2020-05-14
[ENH] s-SNOM FFT Processing: Class complex fft part1

levandoskije:
Hi, @stuart-cls,
This is the [part 1](https://github.com/Quasars/orange-spectroscopy/issues/416#issuecomment-626056535), with:
 - Unit Test;
 - Check_metadata();
 - Complex FFT as an inherited class.

[Real FFT](https://user-images.githubusercontent.com/50185701/81973933-3ebb1780-95fb-11ea-9ade-627c09092030.png), [Complex FFT](https://user-images.githubusercontent.com/50185701/81973963-45e22580-95fb-11ea-834a-a5e205a1fd97.png)
Discussion here #416 and #423

PR #445, opened 2020-05-26
[NOMERGE] Test PR440

markotoplak:
I have no idea why does Travis ignore @levandoskije's PRs (it also ignored the previous ones).


stuart-cls:
Interesting, it marked this as merged and showed the test results in #440, I guess because they are the exact same commit

markotoplak:
Hahaha, we confused github. :D

levandoskije:
I felt sorry for Octocat, hahhaha.

PR #447, opened 2020-06-04
[ENH] Normalize Phase of complex data

levandoskije:
I created a new standardization pre-process by reference for the s-SNOM phase, called:
`Normalize phase by reference subtraction`
I tried to modify the code a little so that I could make the most of the existing features.
Hi @stuart-cls, this PR makes part of issue  #416.

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -380,6 +380,26 @@ class NormalizeReference(Preprocess):
                                     data.domain.metas)
         return data.transform(domain)
 
+class _NormalizePhaseReferenceCommon(CommonDomainRef):
+
+    def transformed(self, data):
+        if len(data):
+            ref_X = self.interpolate_extend_to(self.reference, getx(data))
+            return replace_infs(np.angle(np.exp(data.X * 1j) / np.exp(ref_X * 1j)))
+        else:
+            return data
+
+class NormalizePhaseReference(NormalizeReference):
+
+    def __call__(self, data):
+        common = _NormalizePhaseReferenceCommon(self.reference, data.domain)
+        atts = [a.copy(compute_value=NormalizeFeature(i, common))
+                for i, a in enumerate(data.domain.attributes)]
+        domain = Orange.data.Domain(atts, data.domain.class_vars,
+                                    data.domain.metas)
+        return data.transform(domain)
+
+
 
 def features_with_interpolation(points, kind="linear", domain=None, handle_nans=True, interpfn=None):
     common = _InterpolateCommon(points, kind, domain, handle_nans=handle_nans, interpfn=interpfn)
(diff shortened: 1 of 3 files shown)


borondics:
I think the naming might be too specific. A substraction-type normalization could be relevant for other techniques too but if we keep "PHASE" in the method name that might confuse people. Why not just calling it "Subtract Reference"?

levandoskije:
I agree, the mathematical process is like a subtraction, but not a real subtraction.
`norm = np.exp(sam*1j) / np.exp(ref*1j)`
If I do:
`norm = sam - ref`
appear some spikes in the result, the profile is the same, but I can't remove the spikes. I believe these spikes are motivated by the wrap/unwrap phase.
I can change the name to _Normalize Phase by Reference_, what do you think?

borondics:
So, what happens if some user connects a "regular" signal here? This will still do a calculation but it might make no sense, right?

We should indicate that this will be a complex calculation.

How about one of the following names:
- Normalize Complex Phase by Reference
- Normalize Phase by Reference (Complex)
- Normalize by Reference (Complex Phase)

I like the third the most.

Also, please add documentation. I am adding stuff to the Preprocess Spectra documentation in #448. We should be able to also write equations in the help, which I tried there too.

A bit off-topic: I think we should start merging pull requests only if they have tests **and documentation** when adding or significantly modifying user-accessible features. @levandoskije, @stuart-cls, @markotoplak how do you feel about this?

stuart-cls:
Thanks for this PR!

I would rather you implement it as a separate class (just subclass NormalizeReference I think) instead of passing another kwarg / attribute around. I realize the Normalize preprocessors are a bit of a jumble but let's not worry about that right now.

I agree with Feri about the naming. Let's try to keep things as general as possible. And having a good answer to what happens with a "regular" signal is a good idea.

@borondics I'm fine with adding documentation as part of our PRs (when appropriate), perhaps we could borrow the PR issue template from Orange

markotoplak:
I agree with @stuart-cls, this should be a separate class.

levandoskije:
Hey guys,
I will work on this today.

levandoskije:
Hei @stuart-cls, 
I added it as a subclass and did some tests with regular data.

Apply to regular data, doing so:
`norm = np.angle (np.exp (sample * 1j) / np.exp (sample * 1j))`
There are some peaks, but they are in the range of approximately 10 ^ -18, I think this is a numerical problem. What do you think?

levandoskije:
@borondics, I will commit the documentation after we agree with the PR.

stuart-cls:
Thanks, tt looks good to me! :+1:  

I think very small output is to be expected if you "subtract" identical spectral (probably just float/rounding offsets). As long as it doesn't crash / give an error, it's okay if the output isn't useful since it shouldn't be used anyway.

@markotoplak the tests weren't run again, any thoughts?

PR #448, opened 2020-06-08
Improved Preprocess Spectra Documentation

borondics:
I try to types of more sophisticated typesetting:
- using HTML tags for sub- and superscripts
- using Latex annotations for equations. Markdown should be able to handle this but not sure how the documentation will look like... Maybe additional packages will be needed?


stuart-cls:
@borondics just a note, I think you accidentally included some changes to the Preprocessor wdiget in your documentation PR ;)

borondics:
@stuart-cls thanks for the warning. :) I removed the owpreprocess file.

PR #459, opened 2020-07-06
[ENH] s-SNOM FFT Processing: Class complex fft part2

levandoskije:
Hi @stuart-cls !
I did some testing and it is possible to use the standard apodization and centering available.
So, I modified the code and now we have only one function to do this pre-processing.
This PR is a continuation of  PR #440  , issue #416  and I fixed some minor errors from the old one.

 - Blocked spacing editing;
 - Blocked sweep editing
 - Apodization;
 - Centralization;
 - Zero filling;
 - Update tests;

### DIFF
--- a/orangecontrib/spectroscopy/irfft.py
+++ b/orangecontrib/spectroscopy/irfft.py
@@ -1,7 +1,6 @@
 from enum import IntEnum
 
 import numpy as np
-import scipy.signal
 
 
 class ApodFunc(IntEnum):
@@ -343,9 +342,6 @@ class MultiIRFFT(IRFFT):
 class ComplexFFT(IRFFT):
 
     def __call__(self, ifg, zpd=None, phase=None):
-        number_of_points = len(ifg)
-        Nzff = self.zff*number_of_points
-
         ifg -= np.mean(ifg)
 
         if zpd is not None:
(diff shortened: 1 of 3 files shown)


stuart-cls:
Great, thanks @levandoskije ! I will try to look at this soon.

orangecontrib/spectroscopy/irfft.py
```diff
@@ -373,22 +347,18 @@ def __call__(self, ifg, zpd=None, phase=None):
         Nzff = self.zff*number_of_points
```
stuart-cls, line 347: This is no longer strictly true, as the zero_fill() function will pad to the nearest power of 2 for DFT efficiency.
I suggest putting Nzff = ifg.shape[0] after the zero_fill() on L357.
Of course check this for your particular calculations.

stuart-cls (commented):
Hi @levandoskije ,
Please accept my apologies for the extremely tardy review.

This looks great! I only have one comment about the final zero-filling size used to calculate the wavenumbers. Also you can drop the scipy import now.

I'm curious why the test values changed? Is it just small differences in the calculated window?

In general, I think at this point we could recombine the two classes in a nice way but let's not let that hold up merging this :)

levandoskije:
> Hi @levandoskije ,
> Please accept my apologies for the extremely tardy review.
> 
> This looks great! I only have one comment about the final zero-filling size used to calculate the wavenumbers. Also you can drop the scipy import now.
> 
> I'm curious why the test values changed? Is it just small differences in the calculated window?
> 
> In general, I think at this point we could recombine the two classes in a nice way but let's not let that hold up merging this :)

Hi @stuart-cls ,
The tests change because in the old code when I centralize the spectrum, zeros are added on the side and cut off the other, this can introduce peaks in the final result. Called the "Peak Fence Effect", but with your code, it doesn't happen, your code is more efficient than mine hahahaha

stuart-cls:
Ha gotcha, makes sense. Thanks for the update, I'll push a fake PR to trigger the CI tests and if all looks good I'll merge.

Thanks for your hard work!

PR #463, opened 2020-07-19
[ENH] Renishaw readers

borondics:
Added the readers and test with test data. The line scan will be changed to a smaller file later.


borondics:
We still need to add the requirement to the renishawWiRE python package. It exists only no pip, no conda as far as I know... @markotoplak, what is the best way of doing it? We can also take the source code from GitHub if necessary or ask the maintainer to create a conda package?

markotoplak:
@borondics, I reworked PR a bit, so if you play with it, please update to the latest version from github.

Reading images does not work on travis because that renishaw library is incompatible with the latest release of Pillow (7.2.0, released Jun 30). It does work with anything up to pillow 7.1.2 (the previous release). So someone needs to fix that library to also work with the newest pillow release.

About big files: we also try to avoid them in project history, so before merging this I am going to remove the image from the commit where you added it.

markotoplak:
To fix pillow, I opened alchem0x2A/py-wdf-reader#25. I also opened alchem0x2A/py-wdf-reader#26 so we could fix the code so that tests would stop complaining about not closing files.

markotoplak:
@borondics, please check my modifications and feel free to merge.

I removed the data files that were needlessly too big. Could you find smaller test files for depth and map? Preferably using smaller depth / image size and less wavenumbers. No need to do it now, we can add them in another PR.

borondics:
@markotoplak thanks for all this! Yes, the test files were big. I don't have an instrument but we will ask collaborators for small test files. @stuart-cls do you have access to a Renishaw microscope?

PR #464, opened 2020-07-20
Fixes DeprecationWarning for importing from collections

borondics:
Based on https://stackoverflow.com/questions/53978542/how-to-use-collections-abc-from-both-python-3-8-and-python-2-7

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpreprocess.py
+++ b/orangecontrib/spectroscopy/widgets/owpreprocess.py
@@ -1,5 +1,6 @@
 import random
-from collections import Iterable
+from collections.abc import Iterable
+
 from decimal import Decimal
 import time
 


markotoplak:
Because we only support Python>=3.6, we probably do not need to try/catch, just importing from collections.abc should work.

markotoplak:
Cool, tests passed even with a simple solution. Could you please squash these two commits into one?

borondics:
I tried... :D Not sure if this is what you meant.

markotoplak:
Not exactly. What you see now is what I meant. So you almost did the correct thing, but I suspect that github complained at pushing and then you first merged and then pushed... Next time try force pushing when you rewrite history (`git push -f whatever_arguments`).

PR #467, opened 2020-07-25
added spaces

borondics:


markotoplak:
I kind of dislike this linting for the sake of linting, because this makes tracing the history (where did this bug come from) much harder. I tend to avoid it myself.

orangecontrib/spectroscopy/tests/example_connect_settings.py
```diff
@@ -8,13 +8,13 @@
...
-import Orange
-from Orange.widgets.widget import OWWidget
-
 from orangecontrib.spectroscopy.widgets.gui import connect_settings, lineEditFloatRange, \
     ValueTransform, connect_line, MovableVline
 from orangecontrib.spectroscopy.widgets.owspectra import CurvePlot
 
+import Orange
+from Orange.widgets.widget import OWWidget
+
```
markotoplak, line 17: Why did you move imports from Orange? Did the linter complain?

From PEP-0008:

```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
```

orangecontrib/spectroscopy/utils/__init__.py
```diff
@@ -111,7 +111,7 @@ def get_ndim_hyperspec(data, attrs):
     new_shape = tuple([lsa[2] for lsa in ls]) + (data.X.shape[1],)
     hyperspec = np.ones(new_shape) * np.nan
 
-    hyperspec[indices] = data.X
+    hyperspec[tuple(indices)] = data.X
```
markotoplak, line 114: I'd say this is the most important fix in this PR. But it can be be dangerous. Calling `tuple()` on something copies it and also makes the structure slower for numpy to work with it later.

While your changes fix a potential future bug, I'd prefer fixing it in other way: to have indices saved directly in a proper numpy structure. This probably includes changing the output of axes_to_ndim_linspace, maybe just doing some kind of stacking on indices there.

orangecontrib/spectroscopy/tests/example_connect_settings.py
borondics: Yeah, I get lots of these: 

> C0411: third party import "from orangecontrib.spectroscopy.preprocess import Interpolate, interp1d_with_unknowns_numpy, interp1d_with_unknowns_scipy, interp1d_wo_unknowns_scipy, InterpolateToDomain, NotAllContinuousException" should be placed before "import Orange" (wrong-import-order)

orangecontrib/spectroscopy/utils/__init__.py
borondics: Well, let's try to avoid filling the memory and go with what you suggest.

borondics:
@markotoplak, so I don't fully catch the purpose of our pylint tests then. If we run them but also ignore it, what is the point? I think my local pylint is probably configured differently but if I had used the one from Travis, would that be OK to fix whatever it complains about?

orangecontrib/spectroscopy/tests/example_connect_settings.py
markotoplak: Ah, so in this case, pylint does not know that Orange is a third party import. Maybe we would need to fix lint config (probably it reads something from setup.py).

markotoplak:
@borondics, the point is to not make things any worse when adding/fixing code. :)

Sure you can clean things, but keep in mind that you are obscuring history. If is a compromise. If the code is too ugly, I clean. If I edit some nearby code, I try to clean it.

borondics:
OK, let's not clean for the sake of cleaning then, only when we improve code. :)

I'll close this one.

PR #468, opened 2020-07-30
Spectrum Subtraction Preprocessor

borondics:
New preprocessor to subtract spectra from one another. Next feature will be automatic optimization for multiple spectra. Use case e.g. water subtraction.


markotoplak:
@borondics, could you please rebase?

Also, this needs additional tests. The new preprocessor should be added to test_preprocess.py general preprocessing testing, which test how it performs with strange inputs, like NaN values. See how NormalizeReference, which also uses the reference, was added there.

markotoplak:
@borondics, please ignore my previous message. I did it myself. I also made reference input stricter (only allowing refereces of len 1).

PR #469, opened 2020-08-17
[ENH] Spike Removal Preprocessor

noahcabanas:
Hello, I've been working on a new preprocess method for ramen spectra which enables you to remove spikes from cosmic rays or other anomalous sources. First time using git so hopefully I've done this correctly, let me know if there are any changes that need to be made.


markotoplak:
@noahcabanas, thank you for implementing this feature.

I can not comment on it from the spectroscopic point of view (I'll leave this to @borondics and @stuart-cls).

It is great that you added tests for the preprocessor editor; people rarely do it in their first pull request. These tests verify that your editor returns the correct preprocessing object.

What is currently missing are the tests for the `Despike` object. Is it really working as expected? Probably it is, but to ensure that further changes will not break its functionally we should have tests anyway. See for example the `test_emsc.py`, where `TestEMSC` checks if some very simple inputs are processed correctly.

Then, there are robustness preprocessor tests. To use them, add your preprocessor (with parameters that make sense given that (non-raman) data set; just choose a cutoff somewhere in the range of that dataset) to the `test_preprocess.PREPROCESSORS_INDEPENDENT_SAMPLES` list. This will feed all kinds of strange data sets (missing values, different orders of wavenumbers) into your preprocessor. We need to test these because data might be previously processed in Orange in ways that introduce irregularities, which of course you'd never see in the data directly from the machine. I suspect passing these tests will require some modifications. Do not hesitate to ask for help if you get any issues; `CommonDomainOrder` or `CommonDomainOrderUnknowns` from `preprocess.utils` might be handy here, because they already make sure that `transformed` gets "regular" data.

Regarding use of git: what you did is fine. Before we merge, we will likely join some commits to keep history cleaner. Before merging we will also need to prettify your added code to better confirm to the PEP-8 standard. That is why the lint test is failing. If you do not have time for this, no worries, I can do it.

noahcabanas:
Thanks for the feedback @markotoplak! I'll set up a test for the `Despike` object and incorporate the robustness preprocessor tests. I'll also try to use the lint test again and see if I can clean up the code more.

markotoplak:
Thanks! Could you please comment somewhere what do parameters really mean? I saw you used longer names in the editor, but they still do not say exactly how are these parameters used. It would be nice to have a note of it as a comment in the code, in the Despike class, and in the docs.

borondics:
@noahcabanas thanks once again, this is an important contribution! 

I was playing with it and noticed that if I put 1 to the Z-score field there is an error that is not handled.

I think the explanation is a bit long for the parameters. @markotoplak is there something like a very-quick-help control in Orange, like those question marks that you can hover the mouse over and it gives an explanation to the thing next to it? It would be great to have a short name for the parameter and a longer explanation in the popup help and then a **proper documentation** that describes the calculation itself (this is also **missing** from this PR). 

Finally, we **need** to have **an icon that shows the functionality at a glance**. I can try to sketch something...

borondics:
I like the Preprocessor a lot!

I changed a few typos in the documentation. If you have a reference to a paper which you use for this it would be nice to include it, even with a link to it.

markotoplak:
For the documentation to actually appear I also needed to add a link to the preprocess documentation. Doing so, I also removed the duplicated documentation. 

I do not understand the parameters and I fear the users won't either. My main question is: would I obtain exactly the same results if (option 1) I give the method spectra per spectra and then joined these results together (option 2) gave the method all spectra at once?

> First, since spiked data is
> generally significantly larger than the nearby data points it can sort through large
> spectral datasets to pick out spectra which are spiked.

I do not understand this. Do you need a multiple spectra to perform this actions? Is is only done per spectra? If it is done individually per spectra no need to mention large databases... What does "significantly" mean in this context? Significant compared to what? Because you are only mention z-scores in the next step.

> Following this, it uses a z-scores method 
> to set a limit for these spikes within their individual spectra in order to mark spikes
> and replace them by the average of nearby spikes. 

Does it replace with average of nearby spikes or with the average near to the spikes? Does it replace by looking at the same positions in other spectra or does it only replace with the values of the same spectrum?

> 1. Threshold is used to set a limit inside of a single spectra
> and interpolate regions above.

Is this a strict limit? Does it mean the everything over the limit will be replaced? Or are there other criteria (I do not know, but before you were writing about z-scores and I do not see where the come into play).

> 2. Cutoff sets a difference limit inside of spectra and enables
> the tool to process multi spectra datasets and find spectra which
> have spikes.

Which difference are these? Are these differences within a spectrum or between spectra? Are these differences between peaks? Are these wavenumber or intensity differences? Does this parameter work individually per spectra or needs a whole context?

What is the difference between threshold and cutoff? These words mean approximately the same thing, so it might be useful to explain the difference. Also, it probably means that these words are badly chosen, but if this is the field consensus let's go with it and just explain them more clearly.

> 3. Distance sets the distance over which to average the spiked region with nearby non spiked.

What are the units for the distance? Are these wavenumbers? Or does it mean how many next fields in the data to look at. These would only be the same if the distance between columns is 1 wavenumber, which is in general not true.

markotoplak:
@borondics, please help with the documentation if you understand the method.

orangecontrib/spectroscopy/preprocess/__init__.py
```diff
@@ -721,3 +721,69 @@ def __call__(self, data):
...
+                    else:
+                        w = np.arange(i - self.dis, i + self.dis)
+                    w2 = w[spikes[w] == 0]
+                    row_out[i] = np.mean(row[w2])
+            return out.append(row_out)
+        out = []
+        Series = np.array(data)
+        if Series.size > 0:
+            for row in Series:
+                Distance = np.diff(row, n=0)
```
markotoplak, line 755: @noahcabanas, are you aware that this lines does not do anything? According to the docs for np.diff, calling it with n=0 just returns the same array. Are you sure you wanted to call it n=0? Or did you mean calling it with n=1, which would return the differences between consecutive values?

orangecontrib/spectroscopy/preprocess/__init__.py
```diff
@@ -721,3 +721,69 @@ def __call__(self, data):
...
+                #Spiked spectra are processed and non spiked are passed through
+                if np.any(Distance > self.cutoff):
+                    median1 = np.median(row)
+                    mad_int = np.median([np.abs(row - median1)])
+                    modified_z_scores = 0.6745 * (row - median1) / mad_int
+                    difference = (abs(np.array(modified_z_scores)) > self.threshold)
+                    row_out = row.copy()
```
markotoplak, line 762: row_out should be a parameter of interpolatespikes. To modify an external (semi-global) variable within a function is discouraged. Also, interpolatespikes should not modify the array out per the same principle.

orangecontrib/spectroscopy/widgets/preprocessors/spikeremoval.py
```diff
@@ -0,0 +1,45 @@
...
+    Spike Removal.
+    """
+
+    def __init__(self, parent=None, **kwargs):
+        super().__init__(parent, **kwargs)
+        self.dis = 5
+        self.threshold = 7
+        self.cutoff = 100
+        self.controlArea.setLayout(QVBoxLayout())
+        gui.spin(self.controlArea, self, "cutoff", label="Peak difference cutoff", minv=0,
```
markotoplak, line 18: This cutoff is too limiting. For normalized data, the lowest value of 1 is not enough. It should allow floating point numbers.

markotoplak:
Ok, so now I read the code and have a much better idea that I had when I was reading the documentation. I now understand that this works per-spectrum and therefore there is no need to talk about spectral databases in the documentation.

Then, parameters (@noahcabanas, please confirm if I actually understood it):
- Threshold: a z-value threshold computed for that particular spectrum. Any points exceeding this threshold will be interpolated.
- Cutoff: first filter; if no wn is exceeding this cutoff, than this particular spectra is left as-is. Beware, there might be a bug in using this cutoff; I made a separate comment about it.
- Distance: number of data points to the left and to the right that do not contain a spike are averaged. I suspect there is a bug here, because np.arange(-2, 2) is [-2, -1,  0,  1], which means that this code goes more to the left than to the right. Or is this intentional?

Could you please check these two possible bugs? Did you compare the results to the reference R implementation?

Checks for NaN (and replacement) would be better handled directly after computing each mean. Having one big check at the end can obscure bugs. Also, should these be replaced by 0 or by the median spectra value?

noahcabanas:
@markotoplak Yes, you are spot on about the parameters function. For the distance parameter I forgot to set it to `w = np.arange(i - self.dis, i + 1 self.dis)` so that way it is an even range. I'll also swap the Nan replacement to do it on a per spectra basis, Its meant to replace them by the median spectra value.  

I'll rewrite the documentation today and provide some clarity on the variables and how they operate within the code. 

I'm working on comparing it to the R implementation and have the data they used, but need to convert it into a useable format. (Rdata to something useable) Currently, it seems to be functioning the same as theirs, but I'll hopefully get this to work as a double check.

orangecontrib/spectroscopy/preprocess/__init__.py
noahcabanas: Yes it should be n=1, this was a hold over from when I was initially setting this up and wasn't as familiar with np.diff.

orangecontrib/spectroscopy/widgets/preprocessors/spikeremoval.py
noahcabanas: I'll change it to allow for hundredths values to be input, I didn't think about the need for as specific cutoffs but it would be useful to have.

markotoplak: Hundreths again present some arbitrary limit. Please do not limit it at all, except for the limit that it is non-negative. I think 0 would also present a valid cutoff for the algorithm to work and might be useful with some (high) z-score thresholds.

I'd also avoid using spinboxes. For non-integer inputs free form text windows are usually a better fit. Check, for example, `lineEditDecimalOrNone` from `orangecontrib.spectroscopy.widgets.gui` (it can have limits and default values).

The same might actually go for the z-score limit, but there it is probably not as important. There hundreths (tens?) might be enough.

noahcabanas: Oh, ignore my last change for the cutoff, I'll use the lineEditDecimalOrNone since it seems more robust. Z-score I agree would be fine with hundreths or tens but I'll also do the same for it as the cutoff to not limit it.

noahcabanas: I've changed them to the `lineEditDecimalOrNone ` windows, I've not used them before, but I believe I put them in correctly.

markotoplak:
@noahcabanas, I added a commit that makes the window symmetric and that fixes further bugs at the edges. Before, window computation for the right edge (the other if) was wrong, and also did not work for the left edge.

Of course, because the window is different, the tests now fail. Could you modify the tests in line with the next paragraph?

But all these bugs were not caught before because tests were not prepared properly. It is always beneficial to do a simple example by hand, on paper, and then make a test of this. It would probably be beneficial if your manual example also has peaks close to the sides and forces the algorithm to replace nans. Perhaps you could have `[1000, 1, 1, 1, 2, 1, 1, 1000, 1000, 1000, 2, 1, 1, 1, 1000]` (or something similar) with a distance of 1. Easy to compute by hand and that (or similar) example would catch ALL the bugs we found till now. This one would also test NaN replacement because the middle of the middle peak is problematic with distance of 1.

Also, I do not understand what are you trying to replace NaN's with (I do not understand the code). Could you please explain what were you aiming to do? Also, it does not work, while playing with the preprocessor I got IndexError within there.  That computation should be probably moved right after computing INDIVIDUAL means, right after `row_out[i] = np.mean(row[w2])` line. This will simplify it a lot.

Also, replacing nan-values is a row `r` could by 42 could be done by, for example, `r[np.isnan(r)] = 42`. But if you go for individual means you will not need that, it is only a numpy trick which might benefit you in future.

I am sorry for being such a pain, but it is only for making future code maintenance easier.

noahcabanas:
@markotoplak I'll replace the nan method, I was aiming for something that took the average of nearby non nan values but I'm not sure if thats what I produced, I think I have a better idea of how to do that now based off of your comment though. I also wasn't as sure about setting them to 0 because I thought that could skew any further processing you might do. 

I'll update the tests with some better values and methods and compare them to some hand values. No worries about all the specifics and bug fixes, I'm happy to make whatever changes it needs to be at your standards.

markotoplak:
In those rare NaN cases you could perhaps gradually increase distances for that particular index. Perhaps multiply distance * 2 until you get a valid value? I do not know. These cases are probably rare though so it does not matter much if a method is slow.

noahcabanas:
I tried to setup a new basic method for NaN's that seems to be somewhat functional. I'm not sure how to expand on it and have it iterate if there are still NaN's(Code wise unsure, I understand the idea) but I can think of something and try to sort it out.

markotoplak:
@noahcabanas, I am preparing this branch for merge.

I have:
- cleaned up the commits, now there are fewer
- fixed nan-replacement

I doubt nan replacement was working. I do not understand how did you get the test outputs? As I understood, it was supposed to replace the peak with the nearest valid value, and it was impossible to get `2` in the output, because the input did not contain and `2`s? Therefore the test results were wrong according to my understanding.

I tried to fixed this in the last commit. Replacement is now done with nearest valid values and in constant time. I also optimised the code for speed a bit.

Could you please check manually if the tests outputs are now what a user would expect? If I wrote wrong tests, just comment what the results should be and why, and I will modify the tests (and code) accordingly.

noahcabanas:
@markotoplak, I looked through the tests and they're spot on for what to expect. As for the nan replacement method I used, I was trying to do what you have done but in a much more crude and incorrect fashion which led to the values of 2. I wasn't sure about how to set up a different method for nan values.

markotoplak:
@noahcabanas, thank you for you contribution.

In your next PRs, please never fit tests to the code. The tests are gold standard and should always describe correct behaviour. It is better to have the tests fail. Having invalid tests directly leads to invalid code which is then, hard to detect.

markotoplak:
@noahcabanas, I also fixed some bugs with the Editor. It was crashing if edit fields were empty. I'll merge this now, so if you find any mistake in my code, please open a small PR with a fix. :) Thanks!

PR #472, opened 2020-09-01
[ENH] Widget SNR process

levandoskije:
Hi, I created a new widget for the signal-to-noise ratio in hyperspectral data.
I tried to create it as a pre-process, but I was unable to access information about the data in the pre-process type, so I used the Average widget as a base and built another one.
In this widget, it is possible to calculate the average, standard deviation, and Signal-to-Noise by position. The axis is selected as a category (x and y), if there are no such axes it is possible to select "None".

The icon is a draft and I don't know where to write the documentation.

pre
Workflow.

wid
Widget SNR.

Screen
Example of the application, to see the quality of the signal.


orangecontrib/spectroscopy/widgets/owsnr.py
```diff
@@ -0,0 +1,177 @@
...
+from Orange.data.filter import SameValue, FilterDiscrete, Values
+from Orange.widgets.widget import OWWidget, Msg, Input, Output
+from Orange.widgets import gui, settings
+from Orange.widgets.utils.itemmodels import DomainModel
+import Orange.data.filter as data_filter
+
+
+class OWAverage(OWWidget):
+    # Widget's name as displayed in the canvas
+    name = "snr"
```
borondics, line 12: I would use capitals here like: "SNR"

orangecontrib/spectroscopy/widgets/owsnr.py
```diff
@@ -0,0 +1,192 @@
...
+    OUT_OPTIONS = {'Signal-to-noise ratio': 0, #snr
+                   'Average': 1, # average
+                   'Standard Deviation': 2} # std
+
+    settingsHandler = settings.DomainContextHandler()
+    group_x = settings.ContextSetting(None)
+    group_y = settings.ContextSetting(None)
+    group = settings.ContextSetting(None)
+    out_choiced = settings.ContextSetting(0)
+
```
markotoplak, line 36: Context settings are those that depend on the structure of Table's domain. They mainly exist to prevent incompatible settings. They are, obviously, Group X and Group Y, but not that many others. If you do not want the other settings to be saved as defaults, you can add the `schema_only` flag.

markotoplak:
In the OWAverage this way of grouping (selecting with a filter) could be used because Nominal variables (like classes) have a lower amount of values.

Here, you aim to group a the continuous variable (or a pair of these), that has many values. This will be problematic. Also, sometimes you could get some minor noise in the X and Y values. So a proper solution should handle this noise and be fast, which means it always has to stay with efficient whole-array numpy operations. See `compute_image` in OWHyper. To find values belonging to grid where there is noise (and floating point values are always inexact thus inherently noisy) see `values_to_linspace`.

Also there are some things to talk about organization here. @stuart-cls and @borondics, would it perhaps be better if this was merged with the Average widget? Because it is a point-wise averaging. And then we also have binning which averages as well, and this is directly 1x1 binning. Perhaps a binning widget should be extended with this functionality?

borondics:
I can't select the X and Y axes for some reason so I can't test it yet...

levandoskije:
> I can't select the X and Y axes for some reason so I can't test it yet...


É necessária mudança no Multifile:
Numérico X e Y -> Categórico

raul-freitas:
@borondics, regarding your comment bellow:
> However, the SNR calculation will also depend on the baseline in this case. There was a colleague in CLS who used to calculate SNR spectrally by batching spectral regions and doing the SNR calculation for each. This way one will get a more representative/informative result as the SNR can also change over the spectrum.

If I understood well your point, we are actually doing that, however, for every measured frequency instead of a spectral batch. Bellow I show a picture that sinthetize our procedure:
 
image
In the case, we calculate the average signal (S^bar) and the noise (sigma) at every frequency f_o. Therefore, we are taking into account the noise exactly in the frequency region we estimate the signal. I believe this is a rigorous way of taking into account different noise floors across the full spectral range. Please let us know if that is reasonable.

raul-freitas:
@borondics and @stuart-cls,
do you have a suggestion for us to move on with this particular implementation? We really don't know if the best would be keeping it as a Preprocessor or if we should aim at a new SNR widget. Please, let us discuss soon in order to define the concept.

markotoplak:
@raul-freitas, are the spectra you show from one pixel or are these from a set of pixels? If a set, which ones? Neighbours?

raul-freitas:
> @raul-freitas, are the spectra you show from one pixel or are these from a set of pixels? If a set, which ones? Neighbours?

@markotoplak, the spectra are from a single pixel.

borondics:
How do you get more than one within one pixel?

I'll make you a more detailed description of what I was thinking about.

raul-freitas:
@borondics, for each pixel we accumulate several spectra runs. What you see there are the raw spectra before averaging. We usually take 20 to 30 spectra at each pixel for the sSNOM analysis. Our implementation in Orange allows accessing the raw data, something that the manufacturer's software does not allow.

borondics:
Hi Guys,

SNR calculations will be a great addition to Quasar! @levandoskije, @raul-freitas, thanks for initiating this and putting all the time into this!

I am starting to understand (I think) what is happening. Thanks for the clarifications too. :)

What do you think about the following plan that would make it the most useful for everyone?
- We should consider usability with general datasets, so: single spectrum, series of spectra, hyperspectral maps, multiple-spectra-per-pixel hyperspectral maps (like the SNOM data) and make the new widget work for all of these.
- As I said previously, it would be great to be able to see (hyper)spectral SNRs. @levandoskije, your figure is similar to what I was thinking but this would work only if you had a multispectral pixel map, not in the general case. 
- To deal with this I suggest adding a parameter in the calculation and the GUI as well, that would allow SNR calculation in a sliding window where the window size is defined by the user. The calculation for a single spectrum would be `SNR(x) = mean(x-window -> x+window) / sd(x-window -> x+window)`. The window could be as small as zero, in which case the calculation would need multiple spectra in one pixel or as large as the whole spectrum, where the user would get just one SNR value for each spectrum.
- Please keep in mind @markotoplak's suggestions about speed and the way the data and computation is handled. 
- Is outputting the averages and the SD necessary?
- Having to change the X and Y coordinate feature makes the widget harder to use and might break some functionality where other widgets would need Continuous values.

How does this sound?

raul-freitas:
@borondics, thank you so much for the discussion on this. I believe we are actually aiming at the same goal: a SNR widget that can be applied to any kind of spectral data arrangement. Please see the scheme bellow:
image

I believe that, if we agree with this concept, we can handle the SNR datasets in order to extract specific informations from selected spatial-spectral windows, like you have mentioned. In fact, this analysis can be done in general for point spectra, Spectral Series and HS Maps of any quantity (SNR, Average Signal, Amplitude, Phase, Absorbance, Reflectance...).

Are we in the same page? If yes, we will definitely try to implement your suggestions above.
I believe this widget will be highly used!!! Let's keep the discussion to refine this development.

raul-freitas:
@borondics, everything I organized above works for a "spectral" SNR analysis. If we consider dispersive experiments, like for instance QCL-based IR microscopy, we cannot use the SNR widget described. Then your concern becomes more critical and maybe we can consider the following idea:
image
For that I believe we need to develop a widget for a classical intensity map (like a contour plot) and then allow the selection to be processed in order to provide Average Intensity, SD or SNR values for the selected area. Again, in this case, the data is not a spectral data but a single frequency map with a single value per pixel.

borondics:
Yes, I completely agree for the last two posts, @raul-freitas! It would be great to have this SNR widget. For single wavelength images the SNR could be calculated spatially but I am not sure what benefit that would give, so I would leave that out for now.

We were discussing with @markotoplak for the actual implementation so that the user doesn't need to do changes for the variable types. He will guide you for the best approach.

Thanks guys for the work!!!

raul-freitas:
Hello @borondics and @markotoplak! Thank you for the review on this. We will then focus at the spectral SNR widget and we let the "single line" SNR as a future development. @markotoplak, can you give us some guidance on how we proceed with the coding considering that we will keep this concept for the widget? @levandoskije will stay tuned and will proceed following your recommendations.

markotoplak:
Sorry @raul-freitas, I forgot about this PR for a while. Fortunately @levandoskije contacted me by email and we are moving forward.

levandoskije:
Hi guys,  @markotoplak, @borondics 
I need some help, with domain and metas.
I am trying to change a meta from a data table, a need to substitute just one column from data.
example:
```python
attr_x = "map_x"  # whatever the user chooses in the widget

### TODO insert lsx in metas attr_x not in 0
table_2_coord.metas[:, 0] = np.linspace(*lsx)[unq_coo[:, 0]]
```
So there is this "attrs" in domain
but I want to know what is the index in metas
something like "get_index_from_metas"
Is this possible?

levandoskije:
Hey guys,
@markotoplak , I improved the calculation speed, using the Stack Overflow trick.
I ran all the tests on my computer and there is no problem (except lint), but I don't understand why all tests failed on GitHub.

markotoplak:
@levandoskije, thank you for this contribution. I added some small changes:
- different setting of metas (we talked about this once)
- widget settings that work with saving/loading of workflows

I will merge this. Please check my changes and test them. If you find something to improve, open a new PR.

PR #473, opened 2020-09-11
[ENH] Asymmetric Least Squares Smoothing

noahcabanas:
I've added the ALS Smoothing  modules I mentioned in the issues. I believe I covered all the tests needed but let me know if I'm missing anything, or if there is anything else I need to change. If needed I can add in the documentation that was  in their [collection](https://irfpy.irf.se/projects/ica/_modules/irfpy/ica/baseline.html) but I wasn't sure if that was needed.

### DIFF
--- a/doc/widgets/als.md
+++ b/doc/widgets/als.md
@@ -1,0 +1,73 @@
+#Asymmetric Least Squares Smoothing
+
+**Inputs**
+
+- Data: input dataset
+
+**Outputs**
+
+- Processed: fitted background
+
+The **ALS** tool provides three different methods for Least Squares smoothing. The three processes
+share a smoothing value & factor as well as the iteration variable. In addition to this,
+each process has its own respective variable used in calculation.  
+
+Variable Descriptions:
+
+- Smoothing Constant & Factor: Determines the degree of smoothing of the background. 
+Larger constants will lead to larger smoothing. Values can be input as 1E+10 or 1E-10 for easy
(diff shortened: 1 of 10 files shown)


borondics:
The interface is funny for me at least on Mac. The bottom three lines are very long.
image

How do you initialize the parameters? The exponent starts at 0 for me but that seems to be quite a bad starting point. 
It would be good to add a button to reset the parameters to default values.

noahcabanas:
@borondics I was trying to distinguish between the three parameters which were set for all types(lambda, exponential, and iterations) by having them at the bottom. I realize now that it's a bit jarring having it that way. 

For the exponential initialization, I set it to 0 since one of them was set to 100 in the original repository. I wasn't sure if there was a way to change the base value with the `als_type` variable without adding in an excessive amount of boxes that are for each type. 

Alternatively, if is a way to hide boxes rather than just gray them out when not in use to have a unique initialization for each process's values which would solve the issue for both the initialization and factor issues. If not, I can rearrange the outline to be uniform similar to below and set the base exponential value to 6.  

image 

I'll add in a button to reset the parameters and some more thorough documentation with a link to the original repository.

markotoplak:
@noahcabanas, thank you for this PR.

I slightly edited it, mainly for including the original source code as-was and added its license. While reviewing I also tried to simplify some variable names and strings. If you notice that I did anything strange, please comment here.

PR #475, opened 2020-09-16
PTIR file reader

AustinTisor:
I've added a widget for reading PTIR file datasets (an HDF5 file type used by Photothermal Spectroscopy Corp). The widget copies the File widget format for displaying data. Two test data files are included.


markotoplak:
Thank you for contributing a file format. I think you are the first company that did it. Or did I miss anyone, @borondics?

Now that we started getting more outside contributions, I added CLA signing. I used the license we use for the Orange project. For biolab/orange3 we introduced it a few years ago so that we could (in principle) also start offering Orange under different licenses, for example swap GPL for BSD. I remember it was quite painful to contact all previous contributors to Orange and ask them to sing it...

As I saw, your PTIR file widget adds one combobox selector. Therefore, if I make the following change to the reader code, the existing File widget can be used:

```
@@ -1318,9 +1318,19 @@ class PTIRFileReader(FileFormat, SpectralFileFormat):
             raise IOError("Error reading channels from " + self.filename)
         return channel_map

+    @property
+    def sheets(self):
+        return [label.decode("utf-8") for label in self.get_channels().values()]
+
     def read_spectra(self):
-        if self.data_signal == '':
-            return
+        channels = self.get_channels()
+
+        for c, label in channels.items():
+            if label.decode("utf-8") == self.sheet:
+                self.data_signal = c
+                break
+        else:
+            self.data_signal = list(channels.keys())[0]
```

I added a commit with these changes so that reading with the File widget can be easily tested by pulling this branch. As far as I saw, this small change allows the File widget to read PTIR files perfectly (it will temporarily make PTIR widget inoperable though).

To me, not having a separate widget for a specific file format is a plus. The interface is more uniform and maintenance is much easier. As it is, if some components used by PTIR widget would change, the widget would break. Furthermore, newly added (modified) code will unlikely be tested as well as the always-used ordinarily File widget. A separate reading widget would be useful if the file format was especially complex and there were multiple format-specific settings, which, at least for now, is not the case here.

So, I vote for using (a slightly modified) reader and dropping the widget. What do the others think?

borondics:
@markotoplak, I think you are right, Photothermal is the first company to add code. 👍 

I also think that reading files with the File widget would be the best. Of course, if there are many parameters that the users would need to set we can always come back to an individual widget, but for now I would say @markotoplak's solution is great.

@AustinTisor, I was also thinking about the channel selector: say, a file has both IR and Raman channels. In the current case they would need to load it twice to be able to work with those together. Maybe it would be better to simply load everything and just add a feature that would identify them as OPTIR or Raman spectra? This would introduce a problem with the plotting widgets but we can add a filter between the loader and the display that would select the channel the user wants. Or we can add such functionality to the plotters, to be able to select the channel there.

In this case we could drop the channel selector from the File widget. What do you guys think?

AustinTisor:
@markotoplak Using the file widget instead of having a file-specific widget does seem far cleaner. That was the original plan but I ran into the issue of selecting a data channel to use.

@borondics I definitely agree that it would be best if all the data could be loaded once. Then the selector would only change what data is being sent to the output rather than reading the file again. Adding a filter makes sense to me to make it work through the File widget. Should I change it to work with a filter?

markotoplak:
@AustinTisor, could you therefore modify this PR (or open a new one) where sheets are used using the previous patch? I would do it, but I do not have permissions to edit this branch. Also, `data_signal` can now be removed from class attributes, it can now become private. Maybe tests will need to be modified slightly. Please remove everything concerning the widget (widget tests, icons, docs).

The idea of @borondics, I'd call it multi part data tables, is something that we should think more about in the future. It is not something that we will quickly do. Let's not delay the reader, which is already very useful in this stage.

borondics:
Yes, let's add a meta variable to each spectrum of the imported PTIR files that identifies them between all the possible categories. I would say we should have possible values of ['OPTIR', 'Raman', 'OPTIR-single-wavelength-image', 'OPTIR-visible-image'] according to what data is read. I know the single wavelength image and the visible image is not implemented yet, but I hope it will be :D

borondics:
@AustinTisor, we were discussing this with @markotoplak and came up with a roadmap that would both make it possible to get PTIR files into Orange fast and leave the door open for the future enhancement.

So, we think it would be best to use the built-in selector option in the File widget for deciding if the user wants to load IR or Raman data. This will then make is possible to use all the widgets in Orange immediately.

In the long term we will come up with a strategy to have all the data read at once and be tagged somehow but for this we will need to add the functionality to many other widgets to be able to easily deal with those tags.

What do you think about this plan?

AustinTisor:
@borondics Yes that sounds good to me. Although there are potentially many more categories for the data channel. There is also no way currently to import single wavelength or visible images- the reader is purely for spectral data. Part of the reason for making a separate widget for file reading was in case we would want to extend the reader to include that functionality, but I don't think there is any desire to analyze images in Orange at this time.

Where is the built-in selector of the File widget? I must be missing that. I could also add you both as collaborators if you would like to make edits to this branch.

markotoplak:
The selector is just under the file name. This is the version with my modification.

Screenshot 2020-09-19 at 09 26 21

AustinTisor:
Awesome, thanks @markotoplak. I'll work on making it use the sheet selector instead.

markotoplak:
@AustinTisor, I would like to make a new release soon and I would really like to include this reader. Therefore I opened a PR which is this PR - (the widget) + (compatibility code for sheet support).

The replacement PR, #480, keeps your authorship of the reader. Please confirm if you agree with #480.

markotoplak:
Replaced by #480.

PR #482, opened 2020-10-16
[FIX] hyper: Use Feature combobox show all variables and searchable

### LINKED ISSUE #330: OWHyper: Show all variables in "Use feature" selection
It would be nice to use the new Filter/List view from SelectColumns in the Use Feature area. Less nice but still better would be to include Attributes in the DomainModel if there are less than say 20.

Currently our workflows are full of `processing widget` -> `SelectColumns-to-put-feature-in-Metas` -> `OWHyper` -> `SelectColumns-to-put-feature-back-in-Attributes`

@markotoplak I started looking at this but it wasn't clear what was the best approach for future compatibility

stuart-cls:
Fixes #330

Earlier, I wasn't sure how to handle backwards compatibility but I think our minimum supported Orange version contains this `comboBox` now.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owhyper.py
+++ b/orangecontrib/spectroscopy/widgets/owhyper.py
@@ -943,10 +943,11 @@ class OWHyper(OWWidget):
 
         self.box_values_feature = gui.indentedBox(rbox)
 
-        self.feature_value_model = DomainModel(DomainModel.METAS | DomainModel.CLASSES,
+        self.feature_value_model = DomainModel(DomainModel.SEPARATED,
                                                valid_types=DomainModel.PRIMITIVE)
         self.feature_value = gui.comboBox(
             self.box_values_feature, self, "attr_value",
+            contentsLength=12, searchable=True,
             callback=self.update_feature_value, model=self.feature_value_model)
 
         splitter = QSplitter(self)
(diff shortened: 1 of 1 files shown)


markotoplak:
@stuart-cls, thank you, this works really well. 

But we are braking sharing workflows between different versions of spectroscopy (in some cases). If I save on the older version and open on new, there are no problems. But if I save on new (using one of the newly shown features) and open on and old version we could do either:
1. Fail loudly, as happens with this PR. An exception dialog get shown with `
ValueError: Combo box does not contain item ContinuousVariable(name='sepal length', number_of_decimals=1)`. PROS: users see that something is wrong and hopefully think of updating. CONS: The widget is dead afterwards; because these are context settings even deleting/reinserting widget does not help.
2. Fail silently (show default colors). We would have to use a new setting name for the new combobox, and migrate the contents from the old setting names. Widget will keep working, selections will keep working, but default coloring will be shown and it is impossible to notify the user that not all settings were loaded properly.

@stuart-cls, what do you think?

stuart-cls:
Good catch!

It's really two bad options, but I asked around and consensus was to (1) fail loudly. 

I think this is justified because it will only crash in the case they are collaborating with someone with newer software, so it won't break existing/on-going work they might have.

In addition, it seems like a lot of confusion trying to explain to a colleague why when they open your workflow they see something different than you describe and there is no way for them to select what you describe.

It's too bad there isn't a better way to tell users this workflow is from a newer version and the settings didn't apply properly but I guess that's pretty low-level Orange stuff :)

markotoplak:
@stuart-cls, I agree, let's go with it. I bumped the settings version so that one day, if biolab/orange-widget-base#67 gets implemented, we could see a friendly warning.

PR #483, opened 2020-10-20
[ENH] Agilent: Load visible images for mosaic data

stuart-cls:
Building on the great work in #443 , thanks again @pisarik 

Ref #118 

One thing I've noticed is that having the image stored in `attributes` dict as bytes makes the **Data Info** widget and `data.attributes` command slow / unusable. I think because they try to print/format megabytes of data. 

Not for this PR, but what about passing around a `BytesIO` object (or basically anything that could be an input to the `Image.open()` function in `update_visible_image()`)? 

For the latter the Agilent data could then wait to actually load the image data until requested, since the images are stored as separate files.

### DIFF
--- a/orangecontrib/spectroscopy/agilent.py
+++ b/orangecontrib/spectroscopy/agilent.py
@@ -1,4 +1,6 @@
-__version__ = "0.3.2"
+__version__ = "0.4.0"
+
+import configparser
 from pathlib import Path
 import struct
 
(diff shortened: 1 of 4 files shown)


markotoplak:
Good find about Data info. Yeah, we need to fix that bug somehow.

Could you share any test files with me so that I can check it out? :) Thanks!

pisarik:
Yeap, I think it should be quite safe to store in attributes io.BytesIO instead of raw bytes and it really can help if Data Info tries to print objects in attributes dict:) And I think it is also very good choice for the stuff that can be loaded from disk later:)

markotoplak:
Works nicely with your test file! Thanks.

PR #490, opened 2020-11-02
[FIX] Don't put bytes in Visible Images dict

stuart-cls:
(Builds on #483, look at last commit for changes just in this PR)

Fixes the problem discussed in #483 where **Data Info** and `print(data.attributes)` struggle with printing large amounts of image `bytes`. 

My solution is to change the object passed in the dict to be something directly consumable by `PIL.Image.open()`:
* `io.BytesIO` for OPUS
* `pathlib.Path` for agilent

A few notes:

1. I found the tests insensitive to changing the structure of the img dict to remove img_bytes. I guess because the tests use a lot of synthetic code paths to avoid large actual data files. 
2. I don't seem to have the file for `test_one_visible_image_read` in `test_readers.TestOpusReader` so unsure if I correctly adapted those.
3. Also in this file is the comment (L113)
	`# decompress bytes only in widgets to reduce memory footprint`
Which I'm not sure if it's a test comment or architecture comment, and in any case, is it still accurate with this change?

### DIFF
--- a/orangecontrib/spectroscopy/agilent.py
+++ b/orangecontrib/spectroscopy/agilent.py
@@ -1,4 +1,4 @@
-__version__ = "0.4.1"
+__version__ = "0.4.2"
 
 import configparser
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


markotoplak:
The failing test is unrelated to your changes. Probably the hdf5 file library got updated.

markotoplak:
Thanks! In the future we should probably transform this image dictionary to a class, which should contain most of the image handling logic.

I tested you PR with that `one_visible_image.0` from @pisarik. It works perfectly.

PR #491, opened 2020-11-03
[FIX] Renishaw map reading

markotoplak:
The library we use to open the files opens maps and line scans differently. The current code can now do both.


borondics:
Thanks Marko!

PR #495, opened 2020-11-15
owspectra: Remove (not hide) controlArea

irgolic:
Currently, the controlArea is only hidden, so the splitter is still visible. The proper way to dispose of the controlArea is to use teh `want_control_area` layout option.

Although, looking at the UX of the widget, wouldn't some controlArea options be relevant?
Similar to OWScatterPlot, different tools like Select, Zoom, Reset Zoom... As well as an autocommit control?

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owspectra.py
+++ b/orangecontrib/spectroscopy/widgets/owspectra.py
@@ -1562,6 +1562,8 @@ class OWSpectra(OWWidget):
                 "orangecontrib.infrared.widgets.owspectra.OWSpectra"]
     keywords = ["curves", "lines", "spectrum"]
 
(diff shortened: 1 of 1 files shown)


markotoplak:
Thanks, I somehow never noticed `want_control_area`.

I do agree about the autocommit control. Let's discuss this widget once - you'll probably have many ideas how to improve it.

stuart-cls:
My two cents, I agree that a control area would be nice to expose some of the features hidden under the little menu.

borondics:
@irgolic, thanks for catching this! I would be also very happy to discuss with you about the GUI aspects of the Spectroscopy widgets. Do you think we could schedule a video call together with @markotoplak and @stuart-cls?

PR #504, opened 2020-12-08
Refactor File Readers

borondics:
I am initiating this change to better organize non-generic file readers, especially for large facilities and instrument manufacturers. 

Each facility should have their own file readers in one file so we can hard-code instrument/beamline/endstation specific readers easily.

For now I created empty files for several facilities and didn't move any of the existing files such as `agilent.py` but if you agree to the principle of this change @stuart-cls and @markotoplak I'll go ahead and start doing that as well.

### DIFF
--- a/orangecontrib/spectroscopy/__init__.py
+++ b/orangecontrib/spectroscopy/__init__.py
@@ -1,7 +1,7 @@
 import Orange.data
 import os.path
-from . import data #register file formats
-
+from . import data
+from .io import SOLEIL
 
 def get_sample_datasets_dir():
     thispath = os.path.dirname(__file__)
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -286,68 +286,6 @@ class EnviMapReader(FileFormat, SpectralFileFormat):
         return _spectra_from_image(X, features, x_locs, y_locs)
 
 
-class HDF5Reader_HERMES(FileFormat, SpectralFileFormat):
-    """ A very case specific reader for HDF5 files from the HEREMES beamline in SOLEIL"""
-    EXTENSIONS = ('.hdf5',)
-    DESCRIPTION = 'HDF5 file @HERMRES/SOLEIL'
-
-    def read_spectra(self):
-        import h5py
-        hdf5_file = h5py.File(self.filename)
-        if 'entry1/collection/beamline' in hdf5_file and \
-                hdf5_file['entry1/collection/beamline'][()].astype('str') == 'Hermes':
-            x_locs = np.array(hdf5_file['entry1/Counter0/sample_x'])
-            y_locs = np.array(hdf5_file['entry1/Counter0/sample_y'])
-            energy = np.array(hdf5_file['entry1/Counter0/energy'])
-            intensities = np.array(hdf5_file['entry1/Counter0/data']).T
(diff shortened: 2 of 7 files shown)


markotoplak:
These do not work. :) Try moving one of the readers there. You will see that it won't be shown anywhere, it wont work. These reader files need to be imported when importing orange.spectroscopy. See that __init__.py file where we import data (where the current readers reside).

Also, I'd avoid underscore in module name, so `filereaders` or `readers`. Actually, I'd name it `io`.

markotoplak:
Also, I'd suggest about moving the Soleil readers there. I think there are at least 2. And they should still keep working...

borondics:
Well, they contain nothing at the moment, so of course they don't work yet. This PR is just to discuss the principle of the restructuring. If both of you are OK with the idea I will start making it properly. I'll do the renaming and moving the SOLEIL readers.

stuart-cls:
Hi @borondics 

Thanks for bringing this up, I've been thinking about it a lot lately too.

I was thinking more along the lines of reorganizing data.py because it's too large and hard to wrangle. Something like:

````
data/__init__.py - utils, getx, etc (so we don't break utility imports). Also import all the readers here as needed
data/io/agilent.py
data/io/opus.py
data/io/hdf5/__init__.py (general, and to handle file extension conflicts)
data/io/hdf5/soleil.py
data/io/hdf5/ptir.py
data/io/hdf5/nexus.py ( ;) )
data/io/txt/__init__.py (as above for hdf5)
data/io/txt/soleil.py
...etc
````

The key features for me are: 
* each reader (or group of tightly related readers) is its own file for easier change tracking
* readers that share file types / extensions are grouped together as much as possible to make implementing metareaders easier

The hardest thing is not separating the code, but having all the readers work together. So we can't really have completely independent files for each facility. But I like the idea for conceptual separation.

What do you think?

borondics:
@stuart-cls,

Great that we agree on the main idea. @markotoplak seems to be on board as well as long as things are not rushed and done properly. Good to have quality control!

However, I think creating such a complicated folder structure would make this a bit hard for new developers. Why would we put everything into a "data/" folder and then add an "io/" subfolder? After all all io operations would be data related, no? Then we can save one level of subfolders. Maybe this is not so important though.

The other thing, which I don't know if it is just a design choice or one way is better than another is whether to group readers by extension or by facility? I would vote for facility because then it is very clear who needs to maintain/develop what but if it makes things hard we should go the file extension grouping way. After all we can separate facilities inside the extension files as well.

stuart-cls:
Yes, let's not rush it!

I'm not too worried about the io subfolder, let's drop it. How about something like:
````
data/__init__.py - utils, getx, etc (so we don't break utility imports). Also import all the readers here as needed
## Generic readers (general, and to handle file extension conflicts)
data/hdf5.py 
data/txt.py
## Instruments
data/agilent.py (not important here but for clarity, this is a new file with readers, not the python-agilent-file-formats module)
data/opus.py
data/ptir.py
data/neaspec.py
...etc
## Facilities
data/soleil.py (try to limit to the most custom stuff, most logic is still in .data and .hdf5 / .txt)
...etc
````

Let's not be afraid to iterate here too.

borondics:
I actually meant to drop the data folder. :)

How about a conf call with the three of us this week?

markotoplak:
Superseeded by #575

PR #507, opened 2020-12-20
[MNT] Jaka's unified testing for add-ons

markotoplak:
From biolab/orange3-associate#37


JakaKokosar:
Issue https://github.com/biolab/orange3/issues/5145

stuart-cls:
@markotoplak Great! Does this mean we should update our testing practice (and the documentation?) Currently I mostly run the tests directly with unittest in pycharm, but I think the docs say to use `python setup.py test`

markotoplak:
I still use "python setup.py test" because it is much faster... Tox is more thorough because it installs from scratch. For now, I'll keep testing locally with the old command.

PR #514, opened 2021-01-27
[ENH] Automated peak finding in Spectra

noahcabanas:
I added in a new functionality to the spectra menu to label peaks automatically based off of the scipy find peaks function. In tangent with this is the functionality to add individual lines.


orangecontrib/spectroscopy/utils/Peak_line.py
```diff
@@ -0,0 +1,569 @@
...
+__all__ = ['PeakLine', 'PeakLabel']
+
+"""
+    A modification of the Infinite Line Graphicsobject.
+    Uses the base function as a means to create labels specific for peak
+    marking in spectra. 
+"""
+
+
+class PeakLine(GraphicsObject):
```
markotoplak, line 19: Thanks for the PR. This file mostly seems to be a copy of the Infinite Line from pyqtgraph. This will be very hard to main. Instead, the infinite line should be subclassed, because the small modification the subclass will do are likely to be sufficient.

If subclassing is not possible, all changes to the pyqtgraph code need to clearly marked with a very visible comment.

orangecontrib/spectroscopy/utils/Peak_line.py
```diff
@@ -0,0 +1,569 @@
...
+    """
+    **Bases:** :class:`GraphicsObject <pyqtgraph.GraphicsObject>`
```
markotoplak, line 21: For subclassing or encapsulation of InfiniteLine see Orange.widgets.evaluate.owrocanalysis.InfiniteLine (subclassing with small changes) or orangecontrib.spectroscopy.widgets.gui.MovableVLine (for encapsulation with friendlier signals). Is there any particular reason why you did not use MovableVLine (or its slight modification) for your PR?

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1062,101 @@ def grid_apply(self):
...
+                    # used for first value computation as it isn't in the array
+                else:
+                    count.append(None)
+            peak_locations = []
+            for j, val in enumerate(count):
+                if count[j] is not None and count[j] == 1:
+                    peak_locations.append(peak[j])
+                    # find any peaks which aren't duplicates(None) or nearby mislabels
+            for k, val in enumerate(peak_locations):
+                Label_line = pl.PeakLine()
```
markotoplak, line 1129: Creating new peaks should be a function, because now, the same code pattern repeats 3 times. Perhaps peak_apply could the function if it was given proper arguments.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1062,101 @@ def grid_apply(self):
...
+        else:
+            # covers singular spectra data instances
+            single_spectra = []
+            for i, val in enumerate(data):
+                single_spectra.append(data[i])
+            peaks, _ = find_peaks(single_spectra, height=([minHeight,
+                                                           maxHeight]), prominence=prominence)
+            peaks = x_axis[peaks]
+            for i, val in enumerate(peaks):
+                Label_line = pl.PeakLine()
```
markotoplak, line 1149: Same, this should call a function (a large chunk of code is repeating)

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1062,101 @@ def grid_apply(self):
...
+            Label_line.setSpan(mn=self.minimum_point, mx=self.maximum_point)
+            Label_line.label.setColor(color=QColor(Qt.black))
+            Label_line.label.setPosition(1)
+            Label_line.label.setMovable(True)
+            self.plot.addItem(Label_line)
+            Label_line.updateLabel()
+
+    def peak_apply_auto(self, prominence, minHeight, maxHeight, line_overlap):
+        x_axis = getx(self.data)
+        data = np.array(self.data)
```
markotoplak, line 1096: f you need spectral values, there are in data.X. Data can also contain other meta or class data.

Otherwise, self.data_x already contains the sorted x_axis, and data.X[:, self.data_xsind] would give you data with already sorted wavenumbers. I think you find_peaks function expects ordered inputs, while a data table, after transformations in Orange widgets  became unordered wavenumber-wise. Here it is important to use the ordered variants.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1062,101 @@ def grid_apply(self):
...
+        peak = []
+        if np.shape(data) != (len(data),):
+            for i, val in enumerate(data):
```
markotoplak, line 1099: Did you perhaps try your code on data with 100000 spectra? Please do.

I am just guessing, but grouping of same peaks will need to be written much more efficiently. Perhaps you could sort the list and then go through it only once. You algorithm complexity is O(n**2) where n is number of all peaks in the whole data set, which will not fly. It needs to be at most O(nlogn), achievable with sorting + inspection of the sorted list. There are probably better ways to do it as I thought of now, though.

Also, if there is a huge data set, looking at its subset of spectra for peaks will probably be sufficient.

markotoplak:
Thank you, @noahcabanas, this is a much needed feature.

I looked through the code and left some comments that could help you make the code more robust, faster, and maintainable.

Writing code that is short and does not repeat is really important from the maintenance perspective. For any code we add to the repo, we increase the maintenance burden (which is, I think, for any project the majority of work involved after some time). So we really try to keep all additions as clean as possible.

markotoplak:
Also, maybe I have missed it, but I do not recall any mechanism for saving the peak markings. Ideally, if you save a workflow with marked peaks and then open it on a different computer, the same peaks should be shown. But this is likely the least important of my comments, please tackle this one last.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1095,15 +1095,16 @@ def peak_apply(self, position):
             self.plot.addItem(Label_line)
 
     def peak_apply_auto(self, prominence, minHeight, maxHeight, line_overlap):
-        x_axis = getx(self.data)
-        data = np.array(self.data)
+        x_axis = self.data_x
+        data = self.data
```
markotoplak, line 1099: I think the data here is still problematic. See my previous comment. We have to insure, that the code also works in the presence of meta variable and class attributes, this we need to use just X. So it could be `data.X`. But there is an additional problem: wavenumbers might not be ordered. Therefore I suggested `data = data.X[:, self.data_xsind]` in the previous comment. That should be robust.

doc/widgets/labeling.md
```diff
@@ -0,0 +1,21 @@
+# Peak Labeling
+
+The peak labeling menu enables allows for individual or automatic peak labeling.
```
markotoplak, line 3: Reduce wordiness and try to make writing clearer. Perhaps use Grammarly for help?

doc/widgets/labeling.md
```diff
@@ -0,0 +1,21 @@
...
+This processes is achieved through the use of the Scipy find peaks toolkit.
+Using this, it is possible to find find peaks and use adjustable parameters 
+to adjust what is labeled as a peak even when plotting multiple spectra with
+overlapping peak positions. This process is limited by the clarity of data.
+As such, while it can preform on noisy datasets, it will yield clear results
+after data has been filtered. Added peaks can be remvoed by double right clicking.
```
markotoplak, line 9: Double-clicking is not something that can be easily discovered. Possible solutions, ideally both:
1. Right click on the line and the delete option in the submenu. 
2. An option to select the line by clicking, this selection should have a visual indicator and them pressing delete or backspace for removal.

orangecontrib/spectroscopy/tests/test_owspectra.py
```diff
@@ -146,6 +146,18 @@ def test_select_line(self):
             sa = out.transform(Domain([out.domain[ANNOTATED_DATA_FEATURE_NAME]]))
             np.testing.assert_equal(sa.X, 1)
 
+    def test_Label_peaks(self):
```
markotoplak, line 149: Add peak label tests that tests on ugly data. Whenever we write code we have to make as few assumptions about the data as possible. If a user merged two datasets, wavenumbers would not be continuous and there are going to be many unknown values. So, test also with data that:
1) Contains unknowns, lots of them, also some rows and columns that are completely unknown.
2) Has a mixed order of wavenumbers.

The code will surely have problems with such data, because I saw amax used which does not account for NaNs.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+        if label is None:
+            self.label = VerticalPeakLineLabel(self,
+                                               text=str(self.getXPos()), position=(1))
+
+    def setSpan(self, mn, mx):
+        if self.span != (mn, mx):
+            self.span = (mn, mx)
+            self.update()
+
+    def _computeBoundingRect(self):
```
markotoplak, line 507: Also, this function is mostly copied, what is the difference? Is the difference needed at all? If needed, but predictable, maybe the `super()` function can be called and then the output could be adapted.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
     def focusInEvent(self, *e):
         self.focusIn.emit()
         return QWidget.focusInEvent(self, *e)
+
+
+class VerticalPeakLine(pg.InfiniteLine):
```
markotoplak, line 482: This subclass still redefines too much (copies from pg.InfiniteLine). The first superfluous line is `self.moving = False` and it is just the first.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+        self._boundingRect = None
+        self._endPoints = [0, 1]
+        self._bounds = None
+        self._lastViewSize = None
+
+        if label is None:
+            self.label = VerticalPeakLineLabel(self,
+                                               text=str(self.getXPos()), position=(1))
+
+    def setSpan(self, mn, mx):
```
markotoplak, line 502: This is exact the same function as in the superclass and should be removed.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+            self._bounds = br
+            self._lastViewSize = vs
+            self.prepareGeometryChange()
+
+        self._endPoints = (left, right)
+        self._lastViewRect = vr
+
+        return self._bounds
+
+    def boundingRect(self):
```
markotoplak, line 542: The same as in superclass and therefore needs to be removed.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+    def paint(self, p, *args):
+        p.setRenderHint(p.Antialiasing)
+        left, right = self._endPoints
+        pen = self.currentPen
+        pen.setJoinStyle(QtCore.Qt.MiterJoin)
+        p.setPen(pen)
+        p.drawLine(Point(left, 0), Point(right, 0))
+
+
+    def mouseDragEvent(self, ev):
```
markotoplak, line 556: Should not be modified. The only added thing is `updateLabel` could be connected to the sigDragged signal in `__init__`.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+
+    def mouseClickEvent(self, ev):
+        if self.moving and ev.button() == QtCore.Qt.RightButton:
+            ev.accept()
+            self.setPos(self.startPosition)
+            self.moving = False
+            self.sigDragged.emit(self)
+            self.sigPositionChangeFinished.emit(self)
+
+    def mouseDoubleClickEvent(self, ev):
```
markotoplak, line 583: Double clicking to remove is something we should not use. We need to stick with established UI paradigms.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+        self.line = line
+
+    def valueChanged(self):
+        if not self.isVisible():
+            return
+        value = self.line.value()
+        self.setText(self.format.format(value=value))
+        self.updatePosition()
+
+    def mouseDragEvent(self, ev):
```
markotoplak, line 613: With a quick check I do not see any difference. What is different? If nothing, this function can be removed.

orangecontrib/spectroscopy/widgets/gui.py
```diff
@@ -474,3 +477,152 @@ def set_default(self, v):
...
+            leng = 4
+        self.label.setText(str(round(self.getXPos(), leng)))
+        self.update()
+
+class VerticalPeakLineLabel(pg.InfLineLabel):
+    def __init__(self, line, text="", movable=False, position=0.5, anchors=None, **kwds):
+        super().__init__(line)
+        self.line = line
+
+    def valueChanged(self):
```
markotoplak, line 606: No changes compared to the super class, should be removed.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
         self.plot.showGrid(self.show_grid, self.show_grid, alpha=0.3)
         self.show_grid_a.setChecked(self.show_grid)
 
+    def find_peak_variables(self):
+        if self.maximum_point is None:
+            self.maximum_point = np.amax(self.data.X[:, self.data_xsind])
```
markotoplak, line 1072: amax is problematic for data with unknown values (nans)

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -668,6 +665,13 @@ class CurvePlot(QWidget, OWComponent, SelectionGroupMixin):
     range_x2 = Setting(None)
     range_y1 = Setting(None)
     range_y2 = Setting(None)
+    prominence = Setting(None, schema_only=True)
+    Line_Overlap = Setting(None, schema_only=True)
+    peak_min = Setting(None, schema_only=True)
+    peak_max = Setting(None, schema_only=True)
+    peak_locations = Setting([], schema_only=True)
+    minimum_point = Setting(None, schema_only=True)
```
markotoplak, line 673: minimum_point, maximum_points should not be a setting, because they are always defined by data and never changed by the UI.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -694,12 +698,12 @@ def __init__(self, parent: OWWidget, select=SELECTNONE):
         self.clear_data()
         self.subset = None  # current subset input, an array of indices
         self.subset_indices = None  # boolean index array with indices in self.data
-
+        self.start_point = None  # automatic position for adding single lines in peak_apply
```
markotoplak, line 701: `start_point` should not be a module attribute, as it can be quickly computed when needed (is totally data dependent).

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -694,12 +698,12 @@ def __init__(self, parent: OWWidget, select=SELECTNONE):
...
+        self.peak_state = 0  # Value for determining what our peak clear should do
```
markotoplak, line 702: Why does `peak_state` do? What do its values mean? It is not clear from the code why it is needed, I'd guess it is really not. Peak reappearance should be handled differently.

orangecontrib/spectroscopy/widgets/owspectra.py
markotoplak: Also, ordering is not needed here.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -809,6 +811,46 @@ def __init__(self, parent: OWWidget, select=SELECTNONE):
...
+        self.single_peak = QAction(
+            None, self, shortcut=Qt.Key_P, checkable=False,
+            triggered=self.peak_apply
+        )
+        self.single_peak.setShortcutContext(Qt.WidgetWithChildrenShortcut)
+        self.single_peak = QPushButton("Add Single Line", self)
+        self.reset_labels = QPushButton('Reset graph', self)
+        self.single_peak.clicked.connect(self.find_peak_variables)
+        self.single_peak.clicked.connect(self.peak_apply)
+        self.reset_labels.clicked.connect(self.change_state)
```
markotoplak, line 829: `change_state` should be removed. Peak updating needs to be handled differently.

orangecontrib/spectroscopy/widgets/owspectra.py
markotoplak: Also, does it really have to be computed from the whole data set or just for the data shown? I'd say only on the data shown because this is much faster. And then these can be recomputed when the curves that are shown are resampled.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
         self.plot.showGrid(self.show_grid, self.show_grid, alpha=0.3)
         self.show_grid_a.setChecked(self.show_grid)
 
+    def find_peak_variables(self):
+        if self.maximum_point is None:
+            self.maximum_point = np.amax(self.data.X[:, self.data_xsind])
+        if self.minimum_point is None:
+            self.minimum_point = np.amin(self.data.X[:, self.data_xsind])
+        if self.start_point is None:
+            self.start_point = round(np.median(getx(self.data)), 2)
```
markotoplak, line 1076: Do not round here. Only round strings at the output. Who knows what is the user's range (perhaps it can be 0.0000001 to 0.0000300).

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+
+    def peak_apply(self, position):
+        if self.viewtype == INDIVIDUAL:
+            Label_line = VerticalPeakLine()
```
markotoplak, line 1080: `Label_line` is not a PEP-8 name.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+
+    def peak_apply_auto(self, prominence, minHeight, maxHeight, line_overlap):
+        if self.viewtype == INDIVIDUAL:
+            if line_overlap is None:
+                line_overlap = 0
+            x_axis = self.data_x
+            data = self.data.X[:, self.data_xsind]
+            peak = []
+            if np.shape(data) != (len(data),):
+                for i, val in enumerate(data):
```
markotoplak, line 1103: We should never go though the whole dataset, which can be VERY large, just through the curves shown.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+                        # if its greater than than nearby values
+                        peak_locations.append(val)
+                    elif not peak_locations:
+                        peak_locations.append(val)
+                        # used for first value computation as it isn't in the array
+                self.peak_locations = peak_locations
+                for k, val in enumerate(peak_locations):
+                    self.peak_apply(position=val)
+            else:
+                # covers singular spectra data instances
```
markotoplak, line 1125: Do not duplicate code for single and multiple spectra. Code that works for multiple spectra also needs to work for a single spectrum. Maybe this remained from a previous version.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+                for i, val in enumerate(data):
+                    single_spectra = val
+                    peaks, _ = find_peaks(single_spectra, height=([minHeight,
+                                                                   maxHeight]), prominence=prominence)
+                    peaks = x_axis[peaks]  # array with all locations of peaks
+                    for z, var in enumerate(peaks):
+                        peak.append(var)
+                peak_locations = []
+                sorted_peaks = np.sort(peak)
+                for i, val in enumerate(sorted_peaks):
```
markotoplak, line 1112: peak_locations needs to be a set so that `in` predicate operates in constant amortized time.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+                    if val not in peak_locations and peak_locations != [] \
+                            and abs(sorted_peaks[i-1]-val) >= line_overlap:
+                        # First check for used peaks then check
+                        # if its greater than than nearby values
+                        peak_locations.append(val)
+                    elif not peak_locations:
+                        peak_locations.append(val)
```
markotoplak, line 1119: This is not critical and can always be improved later, just as a note: when peaks are now close, the first one is selected. It would be better if we selected the "middle" one.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1128,7 +1259,7 @@ def current_selection():
 
         # redraw whole selection if it increased or decreased over the threshold
         if len(old_sel_ci) <= MAX_THICK_SELECTED < len(new_sel_ci) or \
-           len(old_sel_ci) > MAX_THICK_SELECTED >= len(new_sel_ci):
```
markotoplak, line 1131: Do not modify the code you did not need to edit. And if you do, ALWAYS do it in a separate commit.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1074,6 +1197,14 @@ def clear_graph(self):
         for m in self.markings:
             self.plot.addItem(m, ignoreBounds=True)
 
+        if self.peak_state != 1:
```
markotoplak, line 1200: `peak_state` needs to be removed and drawing solved differently.

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+            self.minimum_point = np.amin(self.data.X[:, self.data_xsind])
+        if self.start_point is None:
+            self.start_point = round(np.median(getx(self.data)), 2)
+
+    def peak_apply(self, position):
+        if self.viewtype == INDIVIDUAL:
+            Label_line = VerticalPeakLine()
+            Label_line.setMovable(True)
+            Label_line.setPen(pg.mkPen(color=QColor(Qt.black), width=2, style=Qt.DotLine))
+            Label_line.setSpan(mn=self.minimum_point, mx=self.maximum_point)
```
markotoplak, line 1083: Why is this setSpan needed? Why not draw Infinite lines also out of the data set?

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1009,6 +1067,71 @@ def grid_apply(self):
...
+            Label_line.label.setColor(color=QColor(Qt.black))
+            Label_line.label.setPosition(1)
+            Label_line.label.setMovable(True)
+            if position:
+                Label_line.setPos(position)
+                Label_line.label.setText(str(round(position, 3)))
+            else:
```
markotoplak, line 1090: What if a user zoomed-in before adding peak lines? Then the line should appear in the middle of the zoomed-in view. Now it appears in the middle of the data.

markotoplak:
@noahcabanas, thanks for your PR. Sorry for not getting to it sooner.

The code was substantially improved compared to the previous versions, thanks. There is still some work to d (I wrote individual comments). To summarize:
- The code is currently not robust to unknown values or strange cases (like no data in the widget, then it crashes) and is also not tested as such. Perhaps the easiest way to test is to call the automatic peak labeler in a function similar to the existing `test_average` that tests on obnoxious data.
- Any code performing computation should only work on the spectra shown (that is a random subset of 1000 spectra) because who knows how big is the user's data set.
- We should have as few settings and object properties as possible. If something can be inferred from the data, it should not be a setting. Or object property, except in rare cases where we aim to save computation time (but that needs a very good reason for it). We try to avoid object properties too, because they need careful updating.
- The subclasses of InfiniteLine and the Label should be simplified further.
- The `peak_state` property and the drawing tricks should be removed. These need to be solved differently.

Please tell me if you are still prepared to work on this code. If not, no worries, then someone else will take this PR and try to refactor it a bit. Thanks!

If something is unclear, please ask about it.

noahcabanas:
@markotoplak Thanks for getting back to me, I appreciate all the feedback! I'm going to try and clean up some of the issues: working with only plotted data, being more robust, simplifying infinite line, and rewriting the documentation to be direct. One part I'm not sure how to handle is redrawing when reopening a file as I used peak_state as a means to accomplish this.

markotoplak:
Thanks @noahcabanas! About `peak_state`: maybe I imagined what it does incorrectly, so do not worry about it too much. After your changes, I'll try to look again if I can remove it. Maybe I will not be able to. :)

noahcabanas:
@markotoplak I believe I've addressed most of the comments. However, for the redrawing upon open, I'm trying to use the list of peaks I made in the most recent commit called `self.peak_positions` this stored all the peaks objects so you now it will actually plot the peaks as you had them. I'm not sure where to place it to best redraw all the peaks when you reopen a canvas as there are multiple instances where the plotview is cleared. Perhaps I'm misunderstanding the application of settings?

orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -1151,15 +1137,19 @@ def peak_apply_auto(self, prominence, minHeight, maxHeight, line_overlap):
                 elif not peak_locations:
                     peak_locations.append(val)
                     # used for first value computation as it isn't in the array
-            self.peak_locations = peak_locations
             for i, val in enumerate(peak_locations):
                 self.peak_apply(position=val)
 
     def delete_selected_labels(self):
-        for i, val in enumerate(self.line_list):
+        for i, val in enumerate(self.peak_positions):
```
markotoplak, line 1144: This function does not work for clearing multiple labels: the first deleted one will be correct, all others will not because of the shift. Possible solution: go through peak_position in the reverse order.

markotoplak:
Thanks for the fixes. We should not store custom objects in widget settings. There would be problems if we modified that object later in any significant way: the settings then would not unpack properly.

The settings should only be lists, tuples, dicts, ints, floats, strings, bytes and None (basically whatever can be cleanly packed into json), see biolab/orange-widget-base#20

noahcabanas:
@markotoplak I stored the information now as a list of the position and updated span of each peak. I tried a few ways to get the data to save when a label is moved but I wasn't able to get it to work and currently have a manual save button for peak labels.

markotoplak:
Superseeded by #586

PR #524, opened 2021-03-02
PLS: add documentation

ajdapretnar:
Rudimentary documentation for PLS.
Widgets.json is now up to date.

Fixes #524.


borondics:
Since this widget is in Orange Spectroscopy, I think it would be more practical to use a Spectroscopy dataset (Collagen probably) for the Documentation.

markotoplak:
@ajdapretnar, thanks. Also for keeping widgets.json current!

@borondics, I think that non-spectroscopic data sets are fine here, when this widget is improved, we will move it into Orange3 anyway. Perhaps you can add another example from spectroscopy?

ajdapretnar:
I wanted to do it, but collagen doesn't work (as discussed). Any regression data sets for spectroscopy?

stuart-cls:
@markotoplak @borondics wasn't there an example regression dataset included in the original PLS PR?

borondics:
I think there is...

I'll discuss this issue with @markotoplak live.

PR #529, opened 2021-03-04
owpreprocess: fix scroll area size propagation into controlArea

markotoplak:
This fix is needed because the scroll area from orange-widget-base 4.12.0 (and possibly 4.11.0) reacts to size changes differently.

The problem looked like this (hiding and showing the control area would fix it though).

image


markotoplak:
@borondics, could you test and merge this one?

borondics:
The fix works great. However, two tests are failing...

Also, it would be interesting to figure out why the Normalize preprocessor expands so far to the right. There are no visible elements there.
image

PR #531, opened 2021-03-11
Replacing slow numpy functions with faster bottleneck imlementations

borondics:
A very low hanging fruit after @markotoplak told me about it but I am willing to do the easy stuff ;)

If this makes sense, we could replace those functions that have been implemented in faster libraries.

In this case I got a 10x speedup for a 1E6 size vector with 2 NaNs. Admittedly, the speedup was from 500 µs to 5 µs, but still, the 10x sounds good. I can't verify if this saves any memory though.

@markotoplak: should I move on with hunting down all the instances for this function? I can just add everything to this same pull request.


borondics:
Also, should we make the tests faster too or we leave numpy there?

PR #538, opened 2021-03-18
[ENH] Peak Fit widget

### LINKED ISSUE #527: Peak Fit Widget
Expect a PR soon, just wanted to have a note on here that I'm working on it with @scott-cls and Kaiyang.

stuart-cls:
Closes #527 

Hi all, (esp @markotoplak @borondics @clsandt )
Please test this version of the widget and provide feedback! I'm interested in both technical programming feedback and general use/design for actual peak fitting.

Thanks a lot to @markotoplak for the lovely platform of OWIntegrate etc to build on and thanks to @scott-cls Kaiyang and @amanda-cls for working on this with me.

### Working:
- [x] Fits
- [x] Previews of fits and initial parameters
- [x] Threaded computation / progress / cancellation
- [x] (almost) all models from lmfit
- [x] Set constraints (min, max, fixed, delta)

 
### To-do:

- [x] Tests are faster now, but still failing
- [x] Re-base to master
- [x] Documentation :)
- [x] Finalize outputs (area type, other fit indicators)
- [x] Residuals
- [x] ~Fix layout on MacOS~
- [x] Handle Average in Preview

### Bugs

- [x] Adding an Editor pops up a bunch of tiny widget windows
 
### Not this round:
See related Project for more.
 * Parallelization: although I kept this in mind with the design, so in principle I don't think it's much more work, I'd like to make that a separate PR
 * Visual feedback of fitted peaks for a specific spectrum


stuart-cls:
About the possible _Residuals_ output: While useful, it does mean increasing our output memory usage by 50% since we would be making a third `Table` with the size of the original spectra. In principle, the residuals can be calculated by subtracting the _Fits_ output from the spectra when desired. 

Thoughts?

borondics:
I think we should add it by default. The memory usage increase would become
an issue if someone tries to fit a >1GB dataset, which is probably not
feasible for speed reasons anyway. This actually also calls for an "Are you
sure you want to do this" type of warning if the data is large.

stuart-cls:
> I think we should add it by default. The memory usage increase would become an issue if someone tries to fit a >1GB dataset, which is probably not feasible for speed reasons anyway. This actually also calls for an "Are you sure you want to do this" type of warning if the data is large.

Okay, sounds good. Since the calculation is trivial to cancel, I'll let the estimated time to completion stand in for an explicit warning :)

borondics:
I am getting this GUI error when trying to add a Gaussian Model.
Everything is the latest github version on Mac OSX 11.2.3.

```
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/preprocessors/utils.py", line 74, in focusInEvent
    self.focusIn()
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpeakfit.py", line 180, in focusInChild
    self.focusIn()
AttributeError: 'ParamHintBox' object has no attribute 'focusIn'

```

stuart-cls:
Thanks @borondics , looks like a race condition that I win on Windows and you lose on Mac :) I'll add it to the list to fix.

stuart-cls:
Actually, @markotoplak I could use some advice on this one. I think the `editor.focusIn` being set to `self.activateOptions` (on line 322) after instantiation might be a hold-over from Qt4 days, I just copied it from the `IntegrateOneEditor`. 

Is it better to set the callback via a keyword during initialization or should it use something with the `focusInEvent` instead?

markotoplak:
@stuart-cls, I have to admit I have no idea why I wrote the code like this. Now I'd rather create a new signal and not expect some function there to avoid bugs like this.

So it seems that ParamHintBox needs to have focusIn defined earlier on. Maybe you can make a focusIn method in ParamHintBox which redirects to some other function if set? That should avoid the crash. What I do not understand is why would the Qt event loop gets in there already. I thought it would only activate after current python operations finished processing (and is waiting on GUI). Therefore, calling a constructor and then immediately setting focusIn, as you do now, should be fine according to my mental model. Could you replicate the problem on your machine? On Linux it seems to work fine.

@borondics, could you reliably replicate this? Does it happen every time? Do you get the same error regardless of the chosen peak to fit? What are your PyQt versions?

stuart-cls:
Let's not check the git history in case it was me :)

On my Windows machine, I don't see it. I can trigger the bug if I run through the code line-by-line in debug. Which is why I thought it was a race thing. It also comes up when I test the ParamHintBox by itself, which I was planning to come back to eventually. I see it in the latest test logs on the github CI as well.

I will try to fix it by indirection as you suggest, although it seems like it's already just there to collect focus events from the child widgets, I don't think it's called directly.

stuart-cls:
@borondics I think I fixed it, can you try when you get a chance?

borondics:
It works now! Thanks a lot!!!

However, the fixed/limit/delta dropdowns are not very visible.
[image: image.png]

When would it be appropriate to give some feature suggestions to make it
more intuitive/practical to use?

stuart-cls:
@borondics Now is a good time :) With the constraints work done, I'm pretty much feature complete on my original design.

I can't see the screenshot but I'm sure some tweaking of the constraints editor box is in order!

borondics:
@stuart-cls, this is the screen shot on Mac.
image

Visually we could make it more condensed as people will fit multiple peaks, so it would be better not to scroll forever. Not sure if there is a parameter for this in the layout.

For features:
- [ ] It would be great to add to each model a + button, which would add another peak of the same type. This is not essential but would save time for building the full list of peaks.
- [ ] Double clicking the spectra could add a new model of the same type that is currently selected with the center position and intensity filled out (where appropriate)
- [ ] I think exchanging the drop-down menu of fixed/limits/delta with icons would make it visually and functionally better too, where the icons would be sticky buttons (turn on/off)
- [ ] It seems that the fitting is done only on one CPU, but this could be significantly sped up if split to all available (-1 for the user to be able to do other stuff) CPUs.

Also, a very important point is how to inspect the fits... For now the output is only the numbers, which is hard to comprehend using just a data table especially when there are a lot of peaks.

borondics:
I am getting errors when and after adding some models.

```
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpreprocess.py", line 83, in _rowsInserted
    super()._rowsInserted(parent, start, end)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 358, in _rowsInserted
    self._insertWidgetFor(row, index)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 398, in _insertWidgetFor
    widget = self.createWidgetFor(index)
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpreprocess.py", line 71, in createWidgetFor
    w = super().createWidgetFor(index)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 415, in createWidgetFor
    widget = definition.viewclass()
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpeakfit.py", line 603, in __init__
    cb.insertItems(0, self.model.valid_forms)
AttributeError: type object 'ThermalDistributionModel' has no attribute 'valid_forms'
```

Another one:

```
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpreprocess.py", line 83, in _rowsInserted
    super()._rowsInserted(parent, start, end)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 358, in _rowsInserted
    self._insertWidgetFor(row, index)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 398, in _insertWidgetFor
    widget = self.createWidgetFor(index)
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpreprocess.py", line 71, in createWidgetFor
    w = super().createWidgetFor(index)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 415, in createWidgetFor
    widget = definition.viewclass()
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpeakfit.py", line 686, in __init__
    super().__init__()
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpeakfit.py", line 315, in __init__
    m = self.model()
TypeError: __init__() missing 1 required positional argument: 'degree'

```

After this adding a Gaussian model produces the following error:

```
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpreprocess.py", line 83, in _rowsInserted
    super()._rowsInserted(parent, start, end)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 358, in _rowsInserted
    self._insertWidgetFor(row, index)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 400, in _insertWidgetFor
    self.view.setIcon(row, index.data(Qt.DecorationRole))
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 642, in setIcon
    widget = self.widgets()[index]
IndexError: list index out of range
```

Removing the Gaussian model gives this:

```
  File "/Users/borondics/Documents/code/Python/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owpreprocess.py", line 88, in _rowsRemoved
    super()._rowsRemoved(parent, start, end)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 362, in _rowsRemoved
    self._removeWidgetFor(row, None)
  File "/Users/borondics/Documents/code/Python/orange3/Orange/widgets/data/utils/preprocess.py", line 407, in _removeWidgetFor
    widget.edited.disconnect(self.__edited)
TypeError: disconnect() failed between 'edited' and '__edited'
```

stuart-cls:
@borondics wow! I promise on Windows things are much more compact. Is your mac a retina screen? That would explain why some things are large and some are small as we have some hard-coded widget sizes. I've got a mac kicking around so I'll try to fix that. I also wondered about collapsing the extra parameters (amplitude, sigma, etc) and only showing center position by default.

Features:
* Model "+" button: Are you imagining that this would copy all the relevant parameters as well? I'm not sure if it's really that much faster
* Double click + guess: In principle I think this is possible, but maybe this can be a later improvement
* fixed/limits/delta: On mine the drop-down is compact and readable. I like the idea of buttons but I wonder if 3 (sometimes 4) buttons in a row would end up being larger. Do you mean something like the example in https://doc.qt.io/qt-5/qhboxlayout.html#details
* Parallelization: I agree this is important. As I wrote above, I kept it in mind but I'd really rather have it be a separate self-contained PR so we have a solid base to build on
* Inspecting Fits: Currently the total fit and residuals are output, but indeed there's no way to visualize a specific fit component for a specific spectrum. I'm not really sure how to tackle this, it almost requires a dedicated visualization tool

Bugs:

> AttributeError: type object 'ThermalDistributionModel' has no attribute 'valid_forms'
> TypeError: __init__() missing 1 required positional argument: 'degree'
Update your lmfit to >= 1.0.2 (it's in the requirements but these get lost on dev updates)

I guess the follow errors happen because of the crash, let me know if they still happen once you update lmfit.

Thanks for the great feedback!

stuart-cls:
@borondics I tried my best but I just don't understand how to make the layouts work on Mac. I notice the giant spacing is also a problem with the related widgets `OWIntegrate` and `OWPreprocessSpectra` so maybe it's a global Mac thing.

Here's the current look:
windows10
linux
macos

Sorry I couldn't do better but I'm going to have to throw in the towel on this one.

markotoplak:
@stuart-cls, thanks, this looks very good!

I also wouldn't worry too much about the spacing on Mac. It is usable as is.

borondics:
@stuart-cls, thanks a lot for trying to solve this! I hope eventually we can find out what the problem is and solve it.

@markotoplak yes, it is functional, but that doesn't mean it shouldn't be beautiful. Especially on a Mac, right? 🤪

In principle it would be nice to have a great UI.

borondics:
I got this new version and also updated `lmfit`. Everything seems to work fine!

I have a long list of comments here...sorry in advance.

When I try to fit transmittance-like data, things don't look so great though. How could we consider having negative peaks?

@stuart-cls the link you provided above is very similar to what I was thinking about for the fixed/limit/delta dropdown replacement.

For those who are not familiar with `lmfit` models it would be very hard to comprehend what c0, c1, c2, c3, etc. mean for a specific model. Therefore I think we should show the formula of the model and simplify the variables. Some say math formulas are possible to write with Qt (https://stackoverflow.com/questions/14097463/displaying-nicely-an-algebraic-expression-in-pyqt). This could also go to the help, we have [tested](https://github.com/Quasars/orange-spectroscopy/blob/master/doc/widgets/preprocess-spectra.md) that markdown works with math formulas - at least on Mac. 

I have noticed some funny behavior, but this might be a Mac thing again.
When I add a model and press "Commit" (sidetrack: should we change this to "Start Fit"?) the visual feedback is hard to understand. Sometimes the colored curves move as a result of the fit, but sometimes they don't.

I have a suggestion to consider, although not sure how easy it would be to implement it.
The `Spectra` window shows either individual curves or the average. In any case, it seems that at least one curve needs to be selected for which the plots are done. We should always add the average spectrum to the fit. Then, we could display all the peaks for the selected (or the average) **and**, in this case we could also update the numbers on the left with the fitted values with the selected curve's values. I think having the initial guess is very confusing.

The "Zoom to fit" function in the menu should zoom out to include the fits as well, not only the spectra.

stuart-cls:
Glad it's working!

Here we go... ;)

> When I try to fit transmittance-like data, things don't look so great though. How could we consider having negative peaks?

The positive or negative constraints are controlled by the limits of the "amplitude" parameter. 

I tried to implement a global default +ve/-ve button but the Editor abstraction prevents that kind of communication. It's also hard to imagine what would happen if you change that global setting and the user has already set their own amplitude constraints.

> @stuart-cls the link you provided above is very similar to what I was thinking about for the fixed/limit/delta dropdown replacement.

I just don't know what icons unambiguously represent "fixed" "use limits" "delta" and "fixed expression". I'm also not sure if it will end up being smaller? Suggestions welcome :)

> For those who are not familiar with lmfit models it would be very hard to comprehend what c0, c1, c2, c3, etc. mean for a specific model. This could also go to the help, we have tested that markdown works with math formulas - at least on Mac.

Indeed, I was planning on adding that to the documentation. I was planning on just linking to the specific model page in the lmfit documentation as it has the equation and some brief explanation and then usually a nice link to a wikipedia page. Or do you think we must reproduce that?

I also think you unfairly cherry-picked the only confusing model :) I actually can't even get the Polynomial model to work on my test data and was wondering if we should include it.

> When I add a model and press "Commit" the visual feedback is hard to understand. Sometimes the colored curves move as a result of the fit, but sometimes they don't.

There are two separate calculations (just as with Preprocess Spectra and Integrate): the preview calculation and the output calculation. The preview should never change after you press "Commit" as the preview will have already run, but you might end up where if you change the model and then click commit right away it seems like they do.

So to be clear, the visualization is _not_ of the output fit, but of the preview fit.

> sidetrack: should we change this to "Start Fit"?

Sure? Labelling these buttons is always tricky: sometimes the fit will start anyway. 

> The Spectra window shows either individual curves or the average. 

I forgot about the average :)

> In any case, it seems that at least one curve needs to be selected for which the plots are done. We should always add the average spectrum to the fit. Then, we could display all the peaks for the selected (or the average) 

This is a good idea.

> and, in this case we could also update the numbers on the left with the fitted values with the selected curve's values. I think having the initial guess is very confusing.

Hmm, logically it makes sense to me to separate initial fit parameters (which are saved in the workflow) from the fit result of a specific spectrum. Otherwise the workflow would not be generally re-usable.

One idea for this I had was to (optionally) write the fit result parameters in matching colours under each component curve. Would that work?

> The "Zoom to fit" function in the menu should zoom out to include the fits as well, not only the spectra.

I agree! I couldn't figure this out.

I will try to finish the documentation and capture these items either as to-do for this PR or as future items for implementation/discussion in the Project I made.

stuart-cls:
@borondics I started looking at the Average spectrum fit, and it occurs to me it's not very useful: at most you have the average of 3-10 spectra randomly selected from the total dataset. Should we instead just disable the average view?

borondics:
Yeah, I think disabling it would make sense. That also solves the display
problem.

borondics:
@stuart-cls, thanks for creating the Project on the repository. I think this will be a great way of dealing with suggestions and following what has been done, etc.

I think we might be getting close to a merge here, right? I was going to send out an email to Quasar users about the 1.1 release, but then I'll wait until this is in the master and tested a bit by various people to send that out.

clsandt:
Hello
Regarding the fitting of transmittance data, Omnic for example just forbids you to do it. It only allows fitting of absorbance data. Since we have a transmittance <=> absorbance transform widget, why not do the same?

Christophe

De : Stuart Read ***@***.***
Envoyé : vendredi 9 avril 2021 19:28
À : Quasars/orange-spectroscopy
Cc : SANDT Christophe; Mention
Objet : Re: [Quasars/orange-spectroscopy] [ENH] Peak Fit widget (#538)


Glad it's working!

Here we go... ;)

When I try to fit transmittance-like data, things don't look so great though. How could we consider having negative peaks?

The positive or negative constraints are controlled by the limits of the "amplitude" parameter.

I tried to implement a global default +ve/-ve button but the Editor abstraction prevents that kind of communication. It's also hard to imagine what would happen if you change that global setting and the user has already set their own amplitude constraints.

@stuart-cls<https://github.com/stuart-cls> the link you provided above is very similar to what I was thinking about for the fixed/limit/delta dropdown replacement.

I just don't know what icons unambiguously represent "fixed" "use limits" "delta" and "fixed expression". I'm also not sure if it will end up being smaller? Suggestions welcome :)

For those who are not familiar with lmfit models it would be very hard to comprehend what c0, c1, c2, c3, etc. mean for a specific model. This could also go to the help, we have tested that markdown works with math formulas - at least on Mac.

Indeed, I was planning on adding that to the documentation. I was planning on just linking to the specific model page in the lmfit documentation as it has the equation and some brief explanation and then usually a nice link to a wikipedia page. Or do you think we must reproduce that?

I also think you unfairly cherry-picked the only confusing model :) I actually can't even get the Polynomial model to work on my test data and was wondering if we should include it.

When I add a model and press "Commit" the visual feedback is hard to understand. Sometimes the colored curves move as a result of the fit, but sometimes they don't.

There are two separate calculations (just as with Preprocess Spectra and Integrate): the preview calculation and the output calculation. The preview should never change after you press "Commit" as the preview will have already run, but you might end up where if you change the model and then click commit right away it seems like they do.

So to be clear, the visualization is not of the output fit, but of the preview fit.

sidetrack: should we change this to "Start Fit"?

Sure? Labelling these buttons is always tricky: sometimes the fit will start anyway.

The Spectra window shows either individual curves or the average.

I forgot about the average :)

In any case, it seems that at least one curve needs to be selected for which the plots are done. We should always add the average spectrum to the fit. Then, we could display all the peaks for the selected (or the average)

This is a good idea.

and, in this case we could also update the numbers on the left with the fitted values with the selected curve's values. I think having the initial guess is very confusing.

Hmm, logically it makes sense to me to separate initial fit parameters (which are saved in the workflow) from the fit result of a specific spectrum. Otherwise the workflow would not be generally re-usable.

One idea for this I had was to (optionally) write the fit result parameters in matching colours under each component curve. Would that work?

The "Zoom to fit" function in the menu should zoom out to include the fits as well, not only the spectra.

I agree! I couldn't figure this out.

I will try to finish the documentation and capture these items either as to-do for this PR or as future items for implementation/discussion in the Project I made.

stuart-cls:
@borondics I'm done with this phase of the widget now (added documentation). I'd like @markotoplak to look at it if possible as it has some concurrency and settings details I'm not 100% on. But if he's not available, I'm happy with how it is now.

> Regarding the fitting of transmittance data, Omnic for example just forbids you to do it. It only allows fitting of absorbance data. Since we have a transmittance <=> absorbance transform widget, why not do the same?

@clsandt I think Feri was referring to negative-going peaks in the general sense. I agree that blocking transmittance fits would be good but we don't currently have the metadata to do that.

stuart-cls:
Actually, one small thing with the docs: I don't know how those nice black circles with numbers in them are done :)

borondics:
There is a tool from biolab for making those. You could ask @Pretnar, Ajda
***@***.***> to send it to you. I once go it from her but
don't have it anymore....

bluewin4:
Hello!

I've been using this recently and it's been super helpful. I really appreciate a lot of the features and the diversity of functions I can use for my peak fitting as well as its ability to interface with the orange ecosystem effectively. As I was working I noticed a bunch of things that I thought might be useful to post here. 

Peakfitting comments:

**Curve fitting interface:**

- It would be amazing if we could copy-paste different peaks that have similar parameters and just alter the center or whatever afterward
- A bit on how to stop or start the calculation of curve fitting, so I can alter several parameters without having it recalculate in the middle of the alterations.
- Being able to stop calculations on new peaks while attempting to alter parameters would be clutch, it's frustrating to want to enact large scale changes and have it chug for 15 seconds between each alteration to calculate a new set of parameters
- Scrolling causing the parameters to suddenly change is very frustrating because the mouse may suddenly go over a box and cause it to scroll in that instead
- I wish I could disable some curves without having to delete them
- The way to input numbers directly into the boxes is kind of frustrating, realizing I can increment with the arrow keys is helpful but not sufficient. If I want to put in 0.001 it is impossible, if I want to put in 0.01 it's a hassle to type because of the way it keeps inserting digits automatically.
- Sometimes when updating the fitting parameters it won't reload a new curve fit
- Lags out when attempting to adjust peak center because it thinks the incomplete entry is out of range

**Display:**

- The hotkey binding for auto-rescale for the peak fitting doesn't work. It also only rescales to the actual data, being able to scale out to see the fitting curves would be nice.
- Resampling using the plot interface doesn't seem to work, although I can change the number of curves displayed
- There are a lot of times the calculation percentage doesn't appear for curve fitting and it can get difficult to tell what's happening
- When moving between computer monitors the graph axes are unable to keep up and distort (although this seems to be a general UI issue with all plots)
- Using the thermal distributions made orange crash. Painfully haha
- Wish I could highlight the curves in the figure for the fitting, so I can click a curve and the related box lights up
- Occasionally the peak fitting will seemingly randomly fail for a curve very similar to others

stuart-cls:
@bluewin4 Hello! Thank you very much for your detail testing and feedback, it is really invaluable at this point. 

Some things I'm just going to capture as future features so that we can get this first version of the widget out there.

> **Curve fitting interface:**
> 
>     * It would be amazing if we could copy-paste different peaks that have similar parameters and just alter the center or whatever afterward

I tried this but ran into some technical stumbling blocks. I'll add it to the feature list and try again at some point.

>     * A bit on how to stop or start the calculation of curve fitting, so I can alter several parameters without having it recalculate in the middle of the alterations.

If you leave the "Auto-Commit" checkbox unselected, it shouldn't start the full calculation until you press "Commit". If it does, that's a bug.
If you mean the preview calculation, it's a balance between delaying the start of calculations and starting too many times.
 
>     * Being able to stop calculations on new peaks while attempting to alter parameters would be clutch, it's frustrating to want to enact large scale changes and have it chug for 15 seconds between each alteration to calculate a new set of parameters

Hmm, it should just cancel any existing calculation when anything is cancelled. I'll look at this more closely.

>     * Scrolling causing the parameters to suddenly change is very frustrating because the mouse may suddenly go over a box and cause it to scroll in that instead

Good catch, I'll look at disabling scroll for those inputs.

>     * I wish I could disable some curves without having to delete them

I can look to see if this is possible.

>     * The way to input numbers directly into the boxes is kind of frustrating, realizing I can increment with the arrow keys is helpful but not sufficient. If I want to put in 0.001 it is impossible, if I want to put in 0.01 it's a hassle to type because of the way it keeps inserting digits automatically.

I limited the input to 2 digits to keep the interface compact, I didn't realize it would limit the input precision so permanently.

>     * Sometimes when updating the fitting parameters it won't reload a new curve fit

Maybe related to the update sluggishness above.

>     * Lags out when attempting to adjust peak center because it thinks the incomplete entry is out of range

Okay.

> 
> **Display:**
> 
>     * The hotkey binding for auto-rescale for the peak fitting doesn't work. It also only rescales to the actual data, being able to scale out to see the fitting curves would be nice.

Thanks, I noticed this as well.

>     * Resampling using the plot interface doesn't seem to work, although I can change the number of curves displayed

Good catch, I didn't test that button. I'm not sure if it's possible.

>     * There are a lot of times the calculation percentage doesn't appear for curve fitting and it can get difficult to tell what's happening

Maybe related to the weirdness you mentioned above. Also, I think the calculation percentage only triggers for the full calculation and not for the preview.

>     * When moving between computer monitors the graph axes are unable to keep up and distort (although this seems to be a general UI issue with all plots)

Yes, there are some big multimonitor/changing resolution issues with Qt and Windows.

>     * Using the thermal distributions made orange crash. Painfully haha

Ah, sorry. This model and a few others need `lmfit>=1.0.2`

>     * Wish I could highlight the curves in the figure for the fitting, so I can click a curve and the related box lights up

We have the inverse, if you click the little play triangle on the model, it will show which model is which colour.

>     * Occasionally the peak fitting will seemingly randomly fail for a curve very similar to others

This may just mean you need tighter constraints, perhaps on amplitude or sigma. 

Thanks again for this great feedback, it's really valuable.

markotoplak:
Good morning. I woke up after my month-long nap. :)

Thanks for this PR. I glanced at the widget before and it seems fine. I'll check the code a bit (especially concurrency) and then likely merge.

It was interesting to see the discussion with so many valuable comments. Seems like a highly anticipated feature. Thanks everyone!

clsandt:
Hello Marko ;
Nice to hear from you.
How is your life?

C

De : Marko Toplak ***@***.***
Envoyé : mercredi 12 mai 2021 17:15
À : Quasars/orange-spectroscopy
Cc : SANDT Christophe; Mention
Objet : Re: [Quasars/orange-spectroscopy] [ENH] Peak Fit widget (#538)


Good morning. I woke up after my month-long nap. :)

Thanks for this PR. I glanced at the widget before and it seems fine. I'll check the code a bit (especially concurrency) and then likely merge.

It was interesting to see the discussion with so many valuable comments. Seems like a highly anticipated feature. Thanks everyone!

bluewin4:
So I'm having a problem and I don't know what to do, I'm doing my peak fitting and pulling the areas for each of my underlying peaks and then dividing it by the integrated area of the fitted peak. The results I'm getting don't make sense though, as it only accounts for ~63% of the total peak area. 

Anyone have any idea what's going on? Is this just an error or am I missing something?
image

PR #540, opened 2021-03-20
STXM/NEXAFS .hdr/.xim import

ctroein:
Hiya, here's a loader for STXM files from MAX IV and CLS, described in an email a while back. I suppose more metadata could be extracted from the files. The parser makes it easy enough to get the info out but I don't know what assumptions can be made about what's in the files, names of properties etc.


markotoplak:
@ctroein, thank you! I really appreciate small test files. And thanks for the meta reader!

PR #543, opened 2021-03-21
Added file reader for .nxs HDF5 STXM files from Diamond Light Source

ctroein:
Refer to email from MP to @borondics on 8 Dec for more test cases.


orangecontrib/spectroscopy/data.py
```diff
@@ -357,6 +357,29 @@ def read_spectra(self):
         return _spectra_from_image(intensities, energies, x_locs, y_locs)
 
 
+class HDF5Reader_NXS(FileFormat, SpectralFileFormat):
```
borondics, line 360: Hi @ctroein,

Thanks a lot for this addition, everything looks great! We have quickly discussed it and there is only one change we would like to ask you to implement before merging into the master. As synchrotron facilities like the NXS format more and more and they each have multiple beamlines/endstations it is not practical to create a class named `HDF5Reader_NXS` and then use it for reading a very special case of the whole set. So, could you rename your class to something like `Diamond_NXS_STXM`?

Having the beamline name/number in the class name would be even better I think as we could easily identify what we need to be able to support. 

The specific beamline should also be reflected in the `DESCRIPTION` and the `IOError`.

stuart-cls: Thanks for this great addition.

@borondics from my tiny understanding of Nexus, I think a given implementation is supposed to be standardized across facilities (like https://manual.nexusformat.org/classes/applications/NXstxm.html) so unless this reader only works for Diamond it seems to me that we could call it NXstxm or something related?

Hope this is helpful and not just drive-by pedantry :)

ctroein: Thanks! My impression after skimming through some of the documentation for the format (and I wasn't aware of its existence until after the comment by @borondics) is that these files don't have all the stuff that's required. Also, the names "entry1" and "Counter1" seem arbitrary, and "photon_energy" too. So I expect that some work would be needed to get it to work with files from other sources. I made the changes mentioned by @borondics but I see now that I failed to push them. Just a sec...

borondics: @stuart-cls, what you are saying makes perfect sense and should be that way. However, from @ctroein's comment it seems the actual naming is not consistent between facilities. Maybe you could double check with some CLS STXM data from a CLS-native Nexus file?

It would be great to unify the readers and decrease the semi-duplicate code. We could even do that if the structure is the same and we just need to build dictionaries for the entry names...

stuart-cls: Okay, well let's not hold this up on naming, we can always generalize later as @borondics suggests. I'll try to run down some CLS Nexus data to test at some point but don't wait for us!

borondics:
Thanks for this PR @ctroein!

Issue #544, opened 2021-03-23
Visual features for the Spectra widget


borondics:
- [ ] If we display only a few lines in the Spectra widget the lines are very thin. I would like to have an option (and keyboard shortcut) that makes the lines thicker progressively in a cycle (1->2->3->4->1->2...). We could disable this feature when there are more than 20 curves so that the widget doesn't get slow.

- [ ] Sometimes it is important to display the 0 level, we could add a menu item to do so.

PR #546, opened 2021-03-24
Zero line feature for the Spectra widget

borondics:
Added an option to the Spectra widget to toggle a horizontal zero-line that can be sometimes useful for inspecting PCA loadings. The line can be toggled from the dropdown menu or by pressing 0.

Partially implements https://github.com/Quasars/orange-spectroscopy/issues/544

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owspectra.py
+++ b/orangecontrib/spectroscopy/widgets/owspectra.py
@@ -669,6 +669,7 @@ class CurvePlot(QWidget, OWComponent, SelectionGroupMixin):
     invertX = Setting(False)
     viewtype = Setting(INDIVIDUAL)
     show_grid = Setting(False)
+    show_zeroLine = Setting(False)
 
     selection_changed = pyqtSignal()
     new_sampling = pyqtSignal(object)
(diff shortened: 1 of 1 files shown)


orangecontrib/spectroscopy/widgets/owspectra.py
```diff
@@ -669,6 +669,7 @@ class CurvePlot(QWidget, OWComponent, SelectionGroupMixin):
     invertX = Setting(False)
     viewtype = Setting(INDIVIDUAL)
     show_grid = Setting(False)
+    show_zeroLine = Setting(False)
```
markotoplak, line 672: Dont use camelCase.

stuart-cls:
Works for me, looks nice with 2nd derivatives too! Doesn't clash with show grid setting either. Although the list of visualization tweaks that poor little popup has to hold is getting long...

borondics:
Actually, you have a point for the Grid. I should turn off the zero line if the grid is on. 

For the menu - yeah... I really hope someone will implement a GMail-like overlay menu. Or we could add submenus...

stuart-cls:
I think it's better to leave zero line / grid settings unconnected: if we're going to expose both might as well let the user choose.

markotoplak:
@stuart-cls, @borondics, do you think that this is really needed?

From a functionality perspective, the grid covers the use cases I envision. I would avoid adding such features unless they are very good reasons for it. This feature is nicely written and not problematic for maintenance, but I afraid of feature creep in the UI.

borondics:
This was a request during a workshop in relation to looking at second derivatives. I agree that the grid takes care of it though. Maybe what we could do is to add this zero-line in the preprocessors automatically and not expose it as a menu item. However then we still have the problem with the Spectra or Hyperspectra widgets and the behaviours of the different spectral plots would diverge... What do you think?

markotoplak:
I think the grid is enough and that we should not cater to every request. We probably forgot about the grid at the time - if we showed it then, they would likely be satisfied.

borondics:
Does the grid always go through 0? If it does I think we can drop this zero-line feature.

PR #548, opened 2021-03-25
Improve status bar for Spectra

borondics:
Added the little icons and information to the status bar for showing the amount of incoming and outgoing data.


markotoplak:
Have you been following the new way to do it? The new way was done recently and it is prepared in the new (unreleased) widget-base. Please see biolab/orange-widget-base#136 and follow that one (and the corresponding Orange PR). So we can skip the interim solution and directly implement the new way.

Feel free to prepare a PR for it. We will merge when a new version of orange-widget-base is released.

stuart-cls:
@borondics I think this is a great idea in general!
@markotoplak So great there is an automated solution now! Does the new new way require much/any intervention in our widgets or can we just skip to using the magic :)

borondics:
In the current latest masters it is working for some widgets but
Hyperspectra is only partially working. I think we will need to check what
is missing, but it will require a lot fewer lines of code now to implement
this for all Spectroscopy widgets.

PR #555, opened 2021-05-12
peakfit: avoid Table.from_numpy

markotoplak:
`Table.from_numpy` is evil because:

1) We easily forget to copy stuff. Before, .ids were copied, which is    great, but .attributes could also be important in some cases.
2) Addressing Table explicitly can make problems with subclasses of Table. Currently in Orange that happens with the text add-on, which uses Corpus. We also intend to implement out-of-memory tables, so let's start being more careful. :)

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpeakfit.py
+++ b/orangecontrib/spectroscopy/widgets/owpeakfit.py
@@ -71,8 +71,9 @@ def fit_results_table(output, model_result, orig_data):
     domain = Domain(features,
                     orig_data.domain.class_vars,
                     orig_data.domain.metas)
-    return Table.from_numpy(domain, X=output, Y=orig_data.Y,
-                            metas=orig_data.metas, ids=orig_data.ids)
+    out = orig_data.transform(domain)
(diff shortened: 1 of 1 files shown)


markotoplak:
@stuart-cls, could you perhaps check this? To me, it seemed fine in the canvas.

stuart-cls:
I'm very excited about all the points you mention here! These two things often distract me when working with the Table data structure. I will not be sad to see the end of `Table.from_numpy`

I looked and everything seems reasonable. I also had fairly good tests of the output so I'm feeling confident :) There does seem to be a problem still with one of the fitting tests taking too long but not reliably so. (And not related to this PR)

About the shallow copies, it's not a problem for multiple outputs to share sub-arrays?

markotoplak:
> About the shallow copies, it's not a problem for multiple outputs to share sub-arrays?

It is preferable as long as widgets don't edit their inputs (which they should not). It saves memory. Selections, for example, try to make a view into the original table; that is possible if it is continuous or, more generally, anything that can be specified as a python `range`.

PR #559, opened 2021-05-25
[FIX] OWPeakfit handle unordered x for area calculations

### LINKED ISSUE #557: Peakfit: Handle and test for unordered x input
As found in #556 , the widget currently doesn't do anything about unordered x inputs.

stuart-cls:
The scope of this problem was actually very small: only peak area calculations were wrong.
Fixes #557 
Depends on #558

Otherwise:
* `lmfit` is insensitive to x ordering
* I ordered the x array for preview plotting evaluation, otherwise you get some exciting preview plots
* I didn't touch the output generation as the behaviour of most widgets is to leave the ordering alone and just handle it gracefully (as we now do)

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owpeakfit.py
+++ b/orangecontrib/spectroscopy/tests/test_owpeakfit.py
@@ -10,7 +10,7 @@ from Orange.widgets.tests.base import WidgetTest
 from orangewidget.tests.base import GuiTest
 
 from orangecontrib.spectroscopy.data import getx
-from orangecontrib.spectroscopy.preprocess import Cut, LinearBaseline
+from orangecontrib.spectroscopy.preprocess import Cut, LinearBaseline, Integrate
 from orangecontrib.spectroscopy.tests.spectral_preprocess import wait_for_preview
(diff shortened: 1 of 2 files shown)


stuart-cls:
@markotoplak in c63b77ef338f90f72934be11a16e16cf7a3767df I introduced a bad bug where `getx(data).sort()` can return None which then chokes the lmfit `eval`. 

This is fine, except it didn't cause any of my tests to fail. There are lots of exceptions when you run the widget directly, I'm surprised that none of the tests in at least `TestOWPeakFit` catch these exceptions. Any hint how to better design the tests so this doesn't happen again?

markotoplak:
@stuart-cls, I now understand what you are writing about. At first, I thought that tests do not run these functions at all, but then I saw they do (I even added code that enabled preview display). If I replicate your bug, I see exceptions in the terminal, but the tests do not fail.

I do not know why. The function is executed by a signal, but so are any button presses...

markotoplak:
So, what I found out is that if I explicitly call `self.widget.redraw_integral()` the test fails. If the function is called as a consequence of a signal, the exception is only printed, but the test does not fail. I still do not understand why.

PR #560, opened 2021-06-02
[FIX] Multifile widget: proper handling of nominal features with non-overlapping values

markotoplak:
The code had bugs after Orange transitioned from `Variable.make` to a different kind of handling values for `DiscreteVariables`; see the tests in the first commit.

Thanks to @VesnaT for finding the bug and working out the solution for non-spectral files.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owmultifile.py
+++ b/orangecontrib/spectroscopy/widgets/owmultifile.py
@@ -1,43 +1,28 @@
 import os
 from functools import reduce
-from itertools import chain, repeat
-from collections import Counter
+from itertools import chain, count, repeat
+from collections import Counter, namedtuple
 from typing import List
 
 import numpy as np
 
 from AnyQt.QtCore import Qt
-from AnyQt.QtWidgets import QSizePolicy as Policy, QGridLayout, QLabel, QFileDialog,\
-    QStyle, QListWidget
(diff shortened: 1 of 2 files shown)


orangecontrib/spectroscopy/widgets/owmultifile.py
```diff
@@ -377,16 +420,16 @@ def show_error(li, msg):
                     xs, vals, additional = reader.read_spectra()
                     if additional is None:
                         additional = Table.from_domain(empty_domain, n_rows=len(vals))
-                    data_list.append((xs, vals, additional))
+                    additional.special_spectral_data = xs, vals
```
stuart-cls, line 423: Small nit about the `special_spectral_data` naming (throughout): this data is not usually spectral data but rather non-spectral metadata

Also, I couldn't tell if the additional table handling is tested in this PR? `AsciiColReader` always returns `additional=None`

Otherwise :+1:

markotoplak: Thanks!

`special_spectra_data` are the actual wavenumbers and spectra that I just attach to the meta table. I agree the name is funny. :)

Good point on testing the additional data! They probably work as they are handled as normal Orange data in concatenation with the same algorithm as the concatenate uses (multifile does special stuff only for special spectral data), but should be tested still. I'll add a test.

Issue #168, opened 2018-04-27
Hyperspectra: include ROI group on Selection output when multiple groups


stuart-cls:
In the specific case of multiple ROI groups, it would be nice to include the assignment of those groups in the `Selection` output.

This is already done for the `Data` output, but if the user has a large image and only wants to look at the spectra in a few small regions, they have to manually remove the "Unselected" instances using **Select Rows**.

For single ROI group case, nothing needs to change.

markotoplak:
Fixed in #561

PR #561, opened 2021-07-05
[ENH] Visualizations: selection outputs groups

markotoplak:
Orange widgets, like Scatterplot, add a "Group" feature also to their "Selected data" output. Spectroscopy widgets only outputted selected groups in their Data output.

Spectroscopy visualizations were modified to behave as Orange visualizations. Because a new feature is added, this should be backward compatible.

Implements #168

~~**TODO** Think if this really is backward compatible. Some basic contexts, like the ones in Scatterplot, should work. How about Select Columns or perhaps some weird PerfectDomain contexts? If it is not perfectly backward compatible, add compatibility mode to the widgets.~~

I added the compatibility mode because there are cases where the new column makes trouble. The most obvious ones are when there are other widgets in the pipeline that also add "Groups", such as Scatter Plot. Then, downstream widgets would mix these two. Also, if someone used a selection from Spectra and then created classes with, for example, Scatter Plot, these would work differently (the Group feature becomes the target if there is no target in the data).

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owhyper.py
+++ b/orangecontrib/spectroscopy/tests/test_owhyper.py
@@ -625,10 +625,10 @@ class TestVisibleImage(WidgetTest):
         OWHyper.migrate_settings(settings, 6)
         self.assertEqual(settings, {})
         self.widget = self.create_widget(OWHyper, stored_settings=settings)
-        self.assertFalse(self.widget.Information.compat_no_group.is_shown())
+        self.assertFalse(self.widget.compat_no_group)
 
         settings = {}
         OWHyper.migrate_settings(settings, 5)
         self.assertEqual(settings, {"compat_no_group": True})
         self.widget = self.create_widget(OWHyper, stored_settings=settings)
-        self.assertTrue(self.widget.Information.compat_no_group.is_shown())
+        self.assertTrue(self.widget.compat_no_group)
--- a/orangecontrib/spectroscopy/tests/test_owhyper.py
+++ b/orangecontrib/spectroscopy/tests/test_owhyper.py
@@ -17,7 +17,7 @@ from orangecontrib.spectroscopy.data import _spectra_from_image, build_spec_tabl
 from orangecontrib.spectroscopy.preprocess.integrate import IntegrateFeaturePeakSimple
 from orangecontrib.spectroscopy.widgets import owhyper
 from orangecontrib.spectroscopy.widgets.owhyper import \
-    OWHyper, ANNOTATED_DATA_SIGNAL_NAME
+    OWHyper
 from orangecontrib.spectroscopy.preprocess import Interpolate
 from orangecontrib.spectroscopy.widgets.line_geometry import in_polygon, is_left
 from orangecontrib.spectroscopy.tests.util import hold_modifiers, set_png_graph_save
@@ -199,6 +199,8 @@ class TestOWHyper(WidgetTest):
         self.widget.imageplot.select_square(QPointF(9.4, 9.4), QPointF(11.6, 10.6))
         out = self.get_output("Selection")
         np.testing.assert_equal(out.metas, [[10, 10], [11, 10]])
+        np.testing.assert_equal([o[out.domain["Group"]].value for o in out],
+                                ["G1", "G1"])
 
     def test_select_polygon_as_rectangle(self):
         # rectangle and a polygon need to give the same results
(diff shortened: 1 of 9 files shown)


markotoplak:
Even though some Orange developers I consulted thought that the compatibility mode is not needed (that errors would be rare), I decided to implement it. 

Now old workflows will show info boxes on all visualizations.  Now it says `"Compatibility mode: Selection does not output Groups."`; any better suggestions? Alternatively, we could just decide not to show it - it is probably not too important. The sum(expected confusion) is probably lower if we decide to just not show it - users won't likely search for the missing output. What do you think? @borondics, @stuart-cls

stuart-cls:
This is great! 🎉

About the compatibility mode: I don't think we should do it at all.

What is the compatibility mode protecting against? In my experience, users are either (a) using the Selected output to get a single group, and so adding a redundant "Group" meta column will not break anything or (b) using the annotated Data output and then some kind of mangling with Edit Domain or Select Rows to remove the Unselected instances. Since the annotated Data output hasn't changed, these kind of workflows won't break either.

There's also a context settings bug: The compatibility mode is permanent.
 -  Once I open a workflow that enables it, I can't turn it off even in new workflows. (probably just for that domain context, but still).
 - You can see this by using the Liver spectral image dataset and i) create workflow in previous version ii) open in this version iii) create brand new workflow which still uses Liver image dataset.
  - I was expecting that creating new selections in the widget or at least new widgets would remove the compat mode.

Other comments:

When there is a single selection, it seems redundant to have the Group column on the selected output. If this is standardized with the rest of the Orange selection widgets, then better to be consistent I guess. 

One funny thing I hadn't noticed before is that OWSpectra supports multiple selection groups by holding shift/ctrl. They show up in the output (G1, G2) but they don't seem to be visualized in any way. I guess colours are already very over-loaded in meaning. Line selection and click selection seem to handle the groups differently: click selection doesn't seem to support group append.

Thanks!

markotoplak:
@stuart-cls, thanks for the thorugh check!

> What is the compatibility mode protecting against? In my experience, users are either (a) using the Selected output to get a single group, and so adding a redundant "Group" meta column will not break anything or (b) using the annotated Data output and then some kind of mangling with Edit Domain or Select Rows to remove the Unselected instances. Since the annotated Data output hasn't changed, these kind of workflows won't break either.

Yes, I agree, it is rare, but when it bites, it could bite hard (and hard to notice/debug). The cases I see as problematic is when they use a scatter plot selection after the selected spectra: before they one Group, now they have Group and Group (1), and the name Group now describes something else. Another case: if they did k-means after the selected spectrum, the Cluster would become a class. Now, it can't, because the class was taken.
 
> There's also a context settings bug: The compatibility mode is permanent.
>     * Once I open a workflow that enables it, I can't turn it off even in new workflows. (probably just for that domain context, but still).
>     * You can see this by using the Liver spectral image dataset and i) create workflow in previous version ii) open in this version iii) create brand new workflow which still uses Liver image dataset.
>     * I was expecting that creating new selections in the widget or at least new widgets would remove the compat mode.

Thanks, can confirm! This points out a potential bug in settings handling. New widgets should not ever use the compatibility mode. Did you ever notice a similar bug somewhere else? The new setting is schema_only, which Orange should theoretically respect.

> 
> Other comments:
> 
> When there is a single selection, it seems redundant to have the Group column on the selected output. If this is standardized with the rest of the Orange selection widgets, then better to be consistent I guess.

Yes, I agree, it is redundant, but Orange does it... So..

> One funny thing I hadn't noticed before is that OWSpectra supports multiple selection groups by holding shift/ctrl. They show up in the output (G1, G2) but they don't seem to be visualized in any way. I guess colours are already very over-loaded in meaning. Line selection and click selection seem to handle the groups differently: click selection doesn't seem to support group append.

Yup. thanks! I will fix the single selection. Any ideas of how to show differently selected groups? I was thinking of overlaying dashed lines of group color onto the selected curves, but am not fixing this in this PR.

> Thanks!

markotoplak:
I added a workaround for biolab/orange-widget-base#164. The workaround avoids migrating for widget defaults by checking if there is a schema-only setting present. In defaults, it is not. Thanks for finding this obnoxious bug!

I was able to add elements into a current group in Spectra, though.

markotoplak:
As for the test failures: there was a new version of pyqtgraph released. :) Need to check what to do about it, but it is not related to this PR.

stuart-cls:
d0ead9c seems really nice! I like the Mixin. I guess this supersedes #439 as well?

markotoplak:
@stuart-cls, what exactly do you mean? Ugh, yes, thank you for reminding me of that PR I should be working on merging.

With this PR, I was meaning to remove the info "warning" and merge it.

stuart-cls:
Ah whoops, mixed up my bugs there :) That one is related but not impacted by this actually.

I guess I was thinking of #168 and mixed myself up!

Don't let me hold you up :)

PR #565, opened 2021-08-02
[FIX] A migration for a wrongly released owmultifile_vesna

markotoplak:
0.6.0 packages (git repo was fine) unintentionally included a file owmultifile_vesna, which was a multifile widget with differently written internals. Because I deleted that file later, here is a migration (if users used that one).


markotoplak:
Failed tests are unrelated.

PR #573, opened 2021-09-01
[ENH] Reshape Map: can invert individual axes

### LINKED ISSUE #571: Issue with hyperspectra and reshape map widgets for PTIR images
Hello
I noticed that the Hyperspectra widget flips the spectral maps imported from PTIR studio on the vertical axis after they are reshaped with the Reshape widget.
I did not managed to get it right by multiplying the original map_x variable by -1.
If you don't use the reshape map widget, Quasar reads directly map_x and map_y and draws the map correctly but it loooks like in the following image:
image

Cheers
Christophe

markotoplak:
Closes #571.

image

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owreshape.py
+++ b/orangecontrib/spectroscopy/tests/test_owreshape.py
@@ -55,10 +55,24 @@ class TestOWReshape(WidgetTest):
         self.assertEqual(self.widget.ypoints, 100)
         self.widget.commit()
         m = self.get_output("Map data")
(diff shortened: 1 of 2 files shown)


markotoplak:
Any opinions on this? Do we want it?

markotoplak:
@borondics suggested at a meeting to add a drop-down with rotation (0-90-180-270).

markotoplak:
Rename to "Create Map"?

markotoplak:
Maybe make a separate box with these selectors.

PR #574, opened 2021-09-10
[RFH][FIX] Preprocessors should not crash when samples have one attribute

stuart-cls:
So far this is mostly a bug report. I'm not sure the best way to handle this

It affects at least Interpolate and Integrate.


markotoplak:
A good catch, thanks!

PR #575, opened 2021-09-10
[MAINT] Refactor file readers

stuart-cls:
@borondics @markotoplak Replaces #504, I hope I captured all the discussion there.

There is one outstanding problem: The ROCK/`SelectColumnsReader`. It currently had/has no meta reader and conflicts with the `NeaReader`. 

But I don't really see the point of this Reader: Why would I want to load data one column at a time? Why not just use the `AsciiColReader` and **Select Column**s later if there is data I don't want?


markotoplak:
Stuart, thanks, this is good. Could you perhaps follow the next procedure? It is a non-trivial portion of code, and just a few weeks ago I was looking at its history. I would not want to lose this option.

https://stackoverflow.com/questions/38729453/how-to-preserve-git-history-when-refactoring-into-multiple-files

I think it should be easy to update the PR. Just backup the new .io files somewhere safe, return to master, copy the original file into the new .io files, commit, and then copy your backup .io files over them again, commit.

markotoplak:
The columns reader makes sense for an early user of Spectroscopy. Yes, the trick you write should make an equivalent result, but it was added for convenience. Let's not break their workflows. :)

stuart-cls:
Good catch on the git history!

Try it now, it seems to work for me.

About the `SelectColumnsReader`, you are much more sympathetic than I :) Making a meta-reader is difficult as the two readers don't sniff-test their files well and implementing the `sheets()` function in a meta-reader when only one sub-reader supports sheets is hard.

So I just changed the priority to match the effective priority of the old class order from `data.py`

borondics:
Hi @stuart-cls, thanks a lot! This PR is immense but I am glad we restructure it for the future.

markotoplak:
Thanks, @stuart-cls.

What I am worried about is that these things were moved but not deprecated. We could, therefore:
- Make data.py inactive, perhaps move its data.py into utils/data.py
- Import all the moved functions(from io and utils.data) into data.py.
- Add a deprecation warning in data.py that will run on module import.

That should make all the old code still work. Do you perhaps have a better suggestion?

What I suggest for testing moves like in the future is that we commit changes to the tests in a separate PR (no need to modify it now). That makes it possible to test new code with old tests, which should still all work, only raise deprecation warnings.

stuart-cls:
I see four classes of things that were in data.py or otherwise moved:

1. Readers and FileFormats
2. general data functions (`getx`)
3. private data functions (`_spectra-from-image`)
4. External code copies used by readers

I think that 1 and 4 don't need deprecation, they are not used elsewhere except in tests.

I left the general data functions in data.py on purpose, I have `getx` in particular in a hundred scripts. Plus it's a logical place for them and (relatively) short to import.

The "private" data functions which I moved: I was thinking they were not used anywhere else as their output is designed to be consumed in SpectralFileFormat. However I see now that they are used by some widgets. 

So I propose:
* Leave 1, 2, 4 as they are.
* Move the three private functions (`_metatable_maplocs`, `_spectra_from_image`, `_spectra_from_image_2d`) back to data.py

markotoplak:
As for (1), I was using specific readers in scripts once. I'm fine with removing them from `data.py` module though.

The reason why I suggested to retire the module is as follows. If I wanted to open spectral data from Orange with `Orange.data.Table(fn)`, I needed to register them first by importing `orangecontrib.spectroscopy.data`. For that to work, the new version of `data.py` will always need to import all the readers - you made sure that it does and from this perspective the change is correct. Which means it will be forever impossible to import data.py without importing all the readers code. It costs some performance and can give some side-effects.

Now I looked further and see that `orangecontrib.spectroscopy` actually imports data, so we can never avoid importing them, so scrap that reason. I will do as you suggest: move the privates back to data.py and merge.

stuart-cls:
Thanks Marko!

PR #583, opened 2021-10-08
Adapt for Orange's table-lock

markotoplak:
Adapt for #5381

### DIFF
--- a/orangecontrib/spectroscopy/io/util.py
+++ b/orangecontrib/spectroscopy/io/util.py
@@ -99,15 +99,13 @@ class TileFileFormat:
 
     def read(self):
         ret_table = None
+        append_tables = []
         for tile_table in self.read_tile():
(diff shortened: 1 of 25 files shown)


orangecontrib/spectroscopy/io/util.py
```diff
@@ -99,15 +99,13 @@ def read_tile(self):
 
     def read(self):
```
markotoplak, line 100: @stuart-cls, could you please verify if my changes did not break anything? I did not explicitly test them. They do pass the tests though. Thanks!

stuart-cls: I tested with the included workflow on both the example dataset and a larger mosaic. It works fine.

markotoplak: Thanks!

PR #586, opened 2021-11-05
[ENH] Peak labelling in Spectra

markotoplak:
Automated peak finding started by @noahcabanas (thanks!).

This is a squashed and a rebase of #514, and some additional work is needed prior to merging. Notes to self:
- [x] Finish settings that now have a manual button
- [x] Check subclassed line
- [x] Test adding/removing labels
- [x] Test save/load


markotoplak:
I removed automatic peak labeling. It does not work properly (misses real peaks).

To verify, try it on collagen data with the distance between peaks set to 10. The peaks it chooses are a bit off to where they should have been. The code is still in the commit history if anyone wants to improve it.

markotoplak:
I have added some tests, so this seems done. If anyone wishes to try it out...

@borondics recommended adding the line label at the mouse cursor position. I agree, but did not implement this (yet). I might do it now, but perhaps in a later PR. I do not think it makes sense to wait for that. Feel free to add the feature though.

PR #587, opened 2021-11-05
[FIX] Non-limited and typing-friendly spin-boxes

### LINKED ISSUE #564: Peak fit: number entry fields assume two decimals and the decimals are hard to type
The peak fit widget assumes two decimals. Which could be a problem is the units of the data would be different. I do not think we should limit users to specific kinds of data. Even if we only use wavenumbers, the y axis heavily depends on data processing.

Plus, at least on Linux, I can not just type the number in. It corrects as I type it. To type the decimals, I first need to delete whatever number is there. Then I can type the first decimal. The second decimal autofills to 0, which means I have to hit `delete` again to remove the current (autofilled) second decimal, and only then can I type it. The inputs should never obstruct directly typing the complete number it.

I mention these two problems together because the likely share the same fix.

markotoplak:
I always disliked decimal limitations for input wavenumbers. I reviewed biolab/orange3#5647 today which used just what we need here.

Also, now spin boxes now do not call valueChanged after each edit which means that strange issues while typing are gone! So, fixes #564.

TODO:
- [x] Now that there are no decimal limitations, change vertical lines not to set needlessly precise numbers.
- [ ] After moving VerticalLine in PeakFit the fit is not updated; fix this.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - recommonmark
   run:
     - python
-    - orange3 >=3.25
+    - orange3 >=3.28
     - scipy >=0.14.0
     - spectral >=0.18
     - serverfiles >=0.2
--- a/orangecontrib/spectroscopy/widgets/owintegrate.py
+++ b/orangecontrib/spectroscopy/widgets/owintegrate.py
@@ -48,7 +48,7 @@ class IntegrateOneEditor(BaseEditorOrange):
             v = 0.
             self.__values[name] = v
 
-            e = SetXDoubleSpinBox(decimals=4, minimum=minf, maximum=maxf,
+            e = SetXDoubleSpinBox(minimum=minf, maximum=maxf,
                                   singleStep=0.5, value=v)
             e.focusIn = self.activateOptions
             e.editingFinished.connect(self.edited)
--- a/orangecontrib/spectroscopy/widgets/owintegrate.py
+++ b/orangecontrib/spectroscopy/widgets/owintegrate.py
@@ -60,7 +60,9 @@ class IntegrateOneEditor(BaseEditorOrange):
             layout.addRow(name, e)
 
             l = MovableVline(position=v, label=name)
-            l.sigMoved.connect(cf)
+            def set_rounded(_, line=l, name=name):
+                cf(float(line.rounded_value()), name)
+            l.sigMoved.connect(set_rounded)
             self.__lines[name] = l
 
         self.focusIn = self.activateOptions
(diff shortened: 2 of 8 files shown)


markotoplak:
@stuart-cls, could you perhaps try this out? Thanks!

stuart-cls:
@markotoplak This is great! Two small things:

1. Can we disable mouse scroll events resulting in number changes? In both **Integrate** and **Peak Fit** where these are used, the boxes are in lists of models that are almost always long and require scrolling. The natural mouse wheel to scroll this can end up changing the values by mistake, which is definitely not what you want
2. On my system, the numbers are localized in a way that is not what I want:
double-comma
Specifically, the comma in the wavenumber value is weird.

markotoplak:
@stuart-cls, thanks for your comment. I think I solved both of your issues. Could you check?

Apart for the stuff already in TODO, do you see anything else that should be fixed?

stuart-cls:
Works nicely! I like how the vertical line precision changes with the view zoom.

I can't think of anything else. Re the last TODO, I found the preview fit updates after moving the vertical line already (try with 2 offset-center Gaussian)

markotoplak:
Peak fitting updates do not work well. Especially in delta mode. I think it also did not work properly before, but now that changed notification events go out more rarely this became more apparent.

I tried to quickly fix it but I could not. The main issue is that the editors store the same settings in two places: in the `__values` and within the edit-boxes components. Furthermore, state is stored in two different formats (`lmfit`'s and check boxes can not correspond perfectly, because of `delta`), and kudos for keeping them synchronized well enough that this was not visible before.

I was thinking of a rewrite, which would store the state at one spot, just in `__values`. I'd always store it in the form that makes it convenient for editing (that would simplify edit box lines) and only at the end "export" into `lmfit` hints. That would simplify it and perhaps even reduce some signal overhead I remember hearing about.

stuart-cls:
The delta mode is indeed a hack. I'm not very happy with it actually: as you say keeping it in sync between the three layers (lmfit, `__values`, edit boxes) was hard and I probably should have taken that as sign to do something else.

My design goal for the settings was indeed to have `__values` be the source of truth and translate to the edit boxes / lmfit hints only when needed. My secondary goal was to have the saved setting be directly usable as param_hint dict but I see what you mean about making life easier on the GUI side.

I've been thinking about what to do about better initial guesses when placing new peaks and about storing fit results as new initial conditions and how those might influence the delta setting type. The use case is to set a center wavelength and allow it to vary only within a sensible range. If the initial conditions are changed by the user, they likely want the range to change to match. But if they are updated otherwise, then the initial restraints are still correct. 

All of which is to say the delta mode adds a lot of complexity and I'd want to make sure we want to keep it long term before re-writing the settings to accommodate it.

markotoplak:
I'll merge this as is and continue working on Peak Fit in #588.

PR #588, opened 2021-11-08
[FIX] Peakfit: refactor for stability and responsiveness

markotoplak:
Prototyping what was discussed in #587.

This moves the complexity from `editor.param_hints` and `editor.setValues` into `ModelEditor._translate`. The `_translate` is not finished, but I think you'll get the idea. Yust wanted to show you something today...

@stuart-cls, what do you think? How does it perform for you?

To finish this we'd need at least:
- [x] finish `_translate`
- [x] do a setting migration from the old version, which should be easy.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpeakfit.py
+++ b/orangecontrib/spectroscopy/widgets/owpeakfit.py
@@ -26,7 +26,7 @@ from orangecontrib.spectroscopy.widgets.peak_editors import GaussianModelEditor,
     LognormalModelEditor, DampedOscillatorModelEditor, DampedHarmOscillatorModelEditor, \
     ExponentialGaussianModelEditor, SkewedGaussianModelEditor, SkewedVoigtModelEditor, \
     ThermalDistributionModelEditor, DoniachModelEditor, ConstantModelEditor, \
-    LinearModelEditor, QuadraticModelEditor, PolynomialModelEditor
+    LinearModelEditor, QuadraticModelEditor, PolynomialModelEditor, set_default_vary
 
 
 def init_output_array(data, model, params):
@@ -266,6 +266,7 @@ class OWPeakFit(SpectralPreprocess):
     description = "Fit peaks to spectral region"
     icon = "icons/peakfit.svg"
     priority = 1020
+    settings_version = 2
 
     PREPROCESSORS = PREPROCESSORS
     BUTTON_ADD_LABEL = "Add model..."
@@ -400,6 +401,19 @@ class OWPeakFit(SpectralPreprocess):
             self.Outputs.fits.send(None)
             self.Outputs.residuals.send(None)
             self.Outputs.annotated_data.send(None)
+
+    @classmethod
+    def migrate_preprocessor(cls, preprocessor, version):
+        name, settings = preprocessor
+        settings = settings.copy()
+        if version < 2:
+            for n, h in settings.items():
+                if isinstance(h, dict):
+                    h = h.copy()
+                    set_default_vary(h)
+                    settings[n] = h
+            version = 2
+        return [((name, settings), version)]
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/orangecontrib/spectroscopy/widgets/owpeakfit.py
+++ b/orangecontrib/spectroscopy/widgets/owpeakfit.py
@@ -149,15 +149,13 @@ def create_model(item, rownum):
 
 
 def prepare_params(item, model):
+    desc = item.data(DescriptionRole)
+    translate_hints = desc.viewclass.translate_hints
     editor_params = item.data(ParametersRole)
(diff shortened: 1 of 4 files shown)


markotoplak:
Do you have a good example where `expr` is used, so that I can test? The code handling them seems a bit convoluted for me; in the end, it seemed to me that `expr` is always removed, but I have probably read the code wrongly. Thanks!

stuart-cls:
Overall, I think the responsiveness is better (5+ `Gaussian` for example) so it's helping with the Signals etc

I had a quick read-through, I think you caught all the edge cases. I do like having all the logic in one place (`_translate`), not only `editor.param_hints` and `editor.setValues` but also `prepare_params` where a little extra logic was hiding.

Obviously `Voigt` and friends with default `expr` isn't handled yet in the GUI. I also like how you made determining the defaults more explicit.

I think even if we missed some edge case this is already more maintainable and performant, so :+1: from me! I can look through again in more detail or help with any gotchas as needed.

stuart-cls:
> Do you have a good example where expr is used, so that I can test? The code handling them seems a bit convoluted for me; in the end, it seemed to me that expr is always removed, but I have probably read the code wrongly. Thanks!

Sorry, I missed this.

Basically, I decided not to support arbitrary `expr` because it requires translating all the variables from their single-model names ("value") to their composite model names ("g0_value"). However, some of them are required by default, i.e. `Voigt`. 

To work around this, I made sure to remove the expr and allow the lmfit code to generate the correct expression for me, **unless** the user wanted to set their own values for `gamma` in which case `expr` _must_ equal `""` (and thereby be ignored).

Make sense?

markotoplak:
@stuart-cls, I think I finished `_translate`. The previous version still had some big bugs, the biggest of which was improper handling of parameters that are expressions by default. Now I think this is more or less it (unless I find new bugs working further). I'd be grateful if you check it.

I hope to find the time to do settings migration and modify tests later in the afternoon.

orangecontrib/spectroscopy/widgets/peak_editors.py
```diff
@@ -354,6 +317,72 @@ def model_lines():
...
+
+        # special delta case
+        if vary == 'delta' and delta == 0:
+            vary = 'fixed'
+
+        if vary == 'delta':
+            hints["min"] = value - delta
+            hints["max"] = value + delta
+        elif vary == 'limits':
+            hints["value"] = value
```
stuart-cls, line 337: I can't see the point of this line, but it seems harmless

stuart-cls:
Looks good to me. I read through `_translate` carefully and it matches with my understanding and memory of how that all worked previously. I don't see any logic changes (except fixes to the delta case).

Thanks for this!

orangecontrib/spectroscopy/widgets/peak_editors.py
markotoplak: Now the Editor only saves things to dict that are specified. If they are left as defaults, they do not get into dict. But UI still shows you something, specifically, the default 0. This like is there to force that this 0 from the UI actually gets into lmfit, otherwise we would be relying that lmfit's default is also 0. Well, if that is so (I did not check) this line could be removed. :) I rather went for something more explicit here. Good catch, btw!

stuart-cls: Often the `lmfit` defaults we care about are actually `1` rather than `0` (i.e. Gaussian sigma['value']). But I checked and at least in the GUI they are still being set correctly somewhere else. As long as that default is propagating to the `lmfit` Model creation (and it seems it is) we should be good.

markotoplak:
@stuart-cls, this PR is ready for final review. While editing text I found a small bug. I hope I did not break anything while fixing it. :) Thanks!

orangecontrib/spectroscopy/widgets/owpeakfit.py
```diff
@@ -403,6 +402,20 @@ def on_exception(self, ex):
...
+    @classmethod
+    def migrate_preprocessor(cls, preprocessor, version):
+        name, settings = preprocessor
+        settings = settings.copy()
+        if version < 2:
+            for n, h in settings.items():
+                if isinstance(h, dict):
+                    h = h.copy()
+                    set_default_vary(h)
+                    #FIX for vary fixed!!!!!
```
stuart-cls, line 414: Is this fix still needed?

stuart-cls (approved):
@markotoplak Thanks again for this, it is a big improvement. I learned some cool things too :)

Everything looks good to me and works nicely.

orangecontrib/spectroscopy/widgets/owpeakfit.py
markotoplak: Thanks, I wrote this down as a note but then found out that the above function already handles it.

PR #590, opened 2021-11-10
[ENH] Peak Fit: finally responsive (with multiprocessing)

markotoplak:
`lmfit` uses `scipy.optimize`'s functions to perform the actual fitting. The problem there is that the optimizing functions block other threads from executing (time between threads is not shared). Therefore the interface blocks while fitting a single row, which, in some cases, can take long.

This PR aims to use multiprocessing in the processing threads.

On my Linux machine I get completely non-problematic previews. @stuart-cls, could you test this on Windows? Is it responsive? Asking because there might be a substantial process-spin-up difference between platforms.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owpeakfit.py
+++ b/orangecontrib/spectroscopy/widgets/owpeakfit.py
@@ -1,8 +1,13 @@
 import time
 from functools import reduce
+import concurrent.futures
+import multiprocessing
 
-from lmfit import Parameters
+from lmfit import Parameters, Model
+from lmfit.model import ModelResult
 import numpy as np
+
+import pebble
 
 from Orange.data import Table, ContinuousVariable, Domain
 from Orange.widgets.data.owpreprocess import PreprocessAction, Description, icon_path
@@ -29,39 +34,41 @@ from orangecontrib.spectroscopy.widgets.peak_editors import GaussianModelEditor,
     LinearModelEditor, QuadraticModelEditor, PolynomialModelEditor, set_default_vary
 
 
-def init_output_array(data, model, params):
-    """Returns nd.array with correct shape for best fit results"""
-    number_of_spectra = len(data)
+def n_best_fit_parameters(model, params):
+    """Number of output parameters for best fit results"""
     number_of_peaks = len(model.components)
     var_params = [name for name, par in params.items() if par.vary]
     number_of_params = len(var_params)
-    return np.zeros((number_of_spectra, number_of_peaks + number_of_params + 1))
+    return number_of_peaks + number_of_params + 1
 
 
-def add_result_to_output_array(output, i, model_result, x):
-    """Add values from ModelResult to output array"""
+def best_fit_results(model_result, x, shape):
+    """Return array of best-fit results"""
     out = model_result
     sorted_x = np.sort(x)
     comps = out.eval_components(x=sorted_x)
     best_values = out.best_values
 
+    output = np.zeros(shape)
+
     # add peak values to output storage
     col = 0
-    for comp in out.components:
+    for comp in out.model.components:
         # Peak area
-        output[i, col] = np.trapz(comps[comp.prefix], sorted_x)
+        output[col] = np.trapz(comps[comp.prefix], sorted_x)
         col += 1
         for param in [n for n in out.var_names if n.startswith(comp.prefix)]:
-            output[i, col] = best_values[param]
+            output[col] = best_values[param]
             col += 1
-    output[i, -1] = out.redchi
+    output[-1] = out.redchi
+    return output
 
 
 def fit_results_table(output, model_result, orig_data):
     """Return best fit parameters as Orange.data.Table"""
     out = model_result
     features = []
-    for comp in out.components:
+    for comp in out.model.components:
         prefix = comp.prefix.rstrip("_")
         features.append(ContinuousVariable(name=f"{prefix} area"))
         for param in [n for n in out.var_names if n.startswith(comp.prefix)]:
@@ -72,7 +79,7 @@ def fit_results_table(output, model_result, orig_data):
                     orig_data.domain.class_vars,
                     orig_data.domain.metas)
     out = orig_data.transform(domain)
-    with out.unlocked(out.X):
+    with out.unlocked_reference(out.X):
         out.X = output
     return out
 
@@ -90,12 +97,14 @@ def fit_peaks(data, model, params):
     Returns:
         results_table (Orange.data.Table): Table with best fit parameters as features
     """
-    output = init_output_array(data, model, params)
+    output = []
+    shape = n_best_fit_parameters(model, params)
     x = getx(data)
     for row in data:
-        i = row.row_index
         out = model.fit(row.x, params, x=x)
-        add_result_to_output_array(output, i, out, x)
+        output.append(best_fit_results(out, x, shape))
+
+    output = np.vstack(output)
 
     return fit_results_table(output, out, data)
 
@@ -183,6 +192,7 @@ class PeakPreviewRunner(PreviewRunner):
 
     def __init__(self, master):
         super().__init__(master=master)
+        self.pool = pebble.ProcessPool(max_workers=1)
         self.preview_model_result = None
 
     def on_exception(self, ex: Exception):
@@ -227,14 +237,14 @@ class PeakPreviewRunner(PreviewRunner):
             # Pass preview data to widgets here as we don't use on_partial_result()
             for w in self.master.flow_view.widgets():
                 w.set_preview_data(data)
-            self.start(self.run_preview, data, pp_def)
+            self.start(self.run_preview, data, pp_def, self.pool)
         else:
             master.curveplot.set_data(None)
             master.curveplot_after.set_data(None)
 
     @staticmethod
     def run_preview(data: Table,
-                    m_def, state: TaskState):
+                    m_def, pool, state: TaskState):
 
         def progress_interrupt(_: float):
             if state.is_interruption_requested():
@@ -242,10 +252,10 @@ class PeakPreviewRunner(PreviewRunner):
 
         # Protects against running the task in succession many times, as would
         # happen when adding a preprocessor (there, commit() is called twice).
-        # Wait 500 ms before processing - if a new task is started in meanwhile,
+        # Wait 100 ms before processing - if a new task is started in meanwhile,
         # allow that is easily` cancelled.
         for _ in range(10):
-            time.sleep(0.050)
+            time.sleep(0.010)
             progress_interrupt(0)
 
         orig_data = data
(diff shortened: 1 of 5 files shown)


stuart-cls:
@markotoplak nice!

I very much like that the interface doesn't freeze.

I'm sorry to say it "feels" about twice as slow to get the preview result when I switch back and forth. Note that both are way faster since #588 and the responsiveness is very nice now in either case. Perhaps some less subjective timing is in order.

Finally, I have to apologize: I started working on multiprocessing in the spring, I should have pushed up the branch so at least you didn't have to rediscover the `.dumps`/`.loads` pickling problem. I put it now at https://github.com/stuart-cls/orange-spectroscopy/tree/parallel-peakfit , see the last two commits. The main difference is the goal was to chunk up the calculation step, not fix the preview. Perhaps we could include that in this PR as well?

markotoplak:
I also multiprocessing for the calculation step. On my computer it purrs like a kitten.

Screenshot_2021-11-10_20-50-48

I suspect that on Windows pool initialization is a lot slower. Here are my outputs and this should be most of my slowdown. What do you get? Do not run it in debug mode though.

```
DEBUG 1 process pool init 0.02588486671447754
DEBUG Complete pool init 0.13579130172729492
```

Sorry for duplicating your work, now I remember you told me about it once but unfortunately I forgot. With this prototype I somehow screwed up fits or residuals; I was only comparing the coefficients and now I see a failing test. Also, a Model you subclassed is now making trouble. :) Suggestions welcome!

markotoplak:
I see now that you tackled and worked around some of the problems this PR is still facing. I tried working around `.components` not working by going through `.model.components`, but something is still not working properly. I see you did more work in parallel... If you want to improve on this, feel free, I am finished for today. :)

borondics:
In [2]: from multiprocessing import Pool
   ...: import time
   ...: t = time.time()
   ...: with Pool(processes=None) as p:
   ...:     print(time.time() - t)
   ...:
   ...:
0.09644126892089844

stuart-cls:
Using the timing I see (adding 3 Voigt models one-by-one)

````
DEBUG 1 process pool init 0.14500880241394043
DEBUG 1 process pool init 0.09117007255554199
DEBUG 1 process pool init 0.08419442176818848
DEBUG 1 process pool init 0.1093592643737793
DEBUG Complete pool init 0.6375992298126221
````

For just plain process init I get exactly the same kind of numbers as "1 process pool" above.

If I test the completion time as well (adding 3 Gaussian one-by-one):
````
DEBUG 1 process pool init 0.011259794235229492
DEBUG 1 process pool init 0.09433937072753906
DEBUG 1 process pool init 0.09729361534118652
DEBUG result finish 7.688007593154907
````

whereas on `master` I get
````
DEBUG result finish 4.1494622230529785
````
and same result for 3 Voigt. So it seems the slowness is not due to the pool init.

I'm trying to remember what I found when testing different chunk sizes, as always with multiprocessing there is a minimum useful chunk size beyond which it is actually slower.

I didn't have a chance to look at the other stuff, it's Remembrance Day here tomorrow so I'll be able to look again on Friday.

**Comment by markotoplak** [comment]
@stuart-cls in @borondics, thanks!

**Comment by markotoplak** [comment]
@stuart-cls, @borondics, could you try this script please? It needs to be saved in a file, so don't paste into an interpreter.

```
import time
from multiprocessing import Pool
import lmfit

def f(i):
    return i


if __name__ == "__main__":

    t = time.time()
    with Pool(processes=1) as p:
        res = p.map_async(f, range(10))
        print("result", res.get())
    print(time.time() - t)

    t = time.time()
    with Pool(processes=None) as p:
        res = p.map_async(f, range(10))
        print("result", res.get())
    print(time.time() - t)
```

I get 
```
result [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
0.018646955490112305
result [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
0.07419586181640625
```

**Comment by stuart-cls** [comment]
@markotoplak in my dev conda environment, I get 
````
result [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
1.447796106338501
result [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
2.4620189666748047
````
in the installer Quasar conda environment, I get the same values.

**Comment by markotoplak** [comment]
@stuart-cls, could you perhaps try with the latest commit?

I'm trying to use a library that is supposed to be able to cancel individual tasks, but I have no idea whether it works on windows. At least, whether it works fast enough. On my system, I still see that canceling kills the worker process, but a new one is restarted very quickly on this computer. 

Your previews will likely still be slow AFTER one of them is cancelled. Please try it. Next, try to increase `max_workers` here:
`self.pool = pebble.ProcessPool(max_workers=1)`

A bigger number would hopefully put additional workers on standby and consecutive/canceled previews and require less spin-up time. Thanks!

**Comment by markotoplak** [comment]
Hm, now I see I can "simulate" windows behavior myself: `multiprocessing.set_start_method('spawn')`. I will try to make a compromise between what is best for my system and what works on Windows. It is too bad that Windows preview won't ever be as responsive as on my system, but at least it won't block and I suspect the elapsed time on Windows should be the same as before.

I'll try to wrap this up with the help of your code.

**Comment by stuart-cls** [comment]
@markotoplak re:`pebble`:

I see sometimes it is very fast now! **0.4 s**. Sometimes slow as you suspected, ~4 s, but that's as good as the single-threaded.

Increasing `max_workers=2` has no impact, it's still either very fast (~0.4 s) or 4 s.

> at least it won't block and I suspect the elapsed time on Windows should be the same as before.

💯it's much better to not have the GUI block.

**Comment by markotoplak** [comment]
@stuart-cls, thanks. I think that parallelization-wise this should be it:
- for previews, pebble with graceful exits on platforms that do not support "fork",
- and normal multiprocessing.Pool for computations on commit.

Now "just" to take care of result correctness and lmfit's obnoxious saving... I'll take your pattern, where you compute most things on the process and return numpy arrays.

**Comment by markotoplak** [comment]
@stuart-cls, I think this should now be ready for review.

I found another problem for the `ConstantModel` (it could not be "pickled"), and I hacked it in. Also, instead of reimplementing its `eval`, as you did, I used numpy's broadcasting after evaluation (second commit).

File: orangecontrib/spectroscopy/widgets/owpeakfit.py [hunk 1]
```diff
@@ -349,28 +376,45 @@ def progress_interrupt(i: float):
...
+                    except AttributeError:
+                        return 0
+
+                while not res.ready():
+                    progress_interrupt(done() / n * 99)
+                    res.wait(0.05)
+
+            fitsr = res.get()
+
+            for fit, bpar, fitted, resid in fitsr:
```
stuart-cls, line 430: What guarantees do we have here that the order of rows in the result matches the order of rows in the source table? Perhaps I am overly paranoid, but this is why I was using `row.id` in the previous code.

orangecontrib/spectroscopy/widgets/owpeakfit.py
```diff
@@ -255,10 +265,27 @@ def progress_interrupt(_: float):
         model_result = {}
         x = getx(data)
         if data is not None and model is not None:
+
             for row in data:
                 progress_interrupt(0)
-                model_result[row.id] = model.fit(row.x, parameters, x=x)
-
+                res = pool.schedule(pool_fit2, (row.x, model.dumps(), parameters, x))
```
stuart-cls, line 289: More for my own learning: my understanding here is that each preview spectrum is running in it's own task. 

1. Does the for loop block on each `while not res.done()` or is that handled by the `concurrent.futures.wait()`. I suppose it's a moot point as there is only 1 preview worker anyway.
2. Is this really efficient? I think that for 3-10 fits the start-up time might dominate, but maybe this is what we were discussing earlier in the PR and it's not a big deal

orangecontrib/spectroscopy/widgets/owpeakfit.py
```diff
@@ -349,28 +376,45 @@ def progress_interrupt(i: float):
...
-                add_result_to_output_array(output, i, out, x)
-                fits.append(out.eval(x=x))
-                residuals.append(out.residual)
-                progress_interrupt(i / n * 100)
-            data = fit_results_table(output, out, orig_data)
+
+            with multiprocessing.Pool(processes=None,
+                                      initializer=pool_initializer,
+                                      initargs=(model.dumps(), parameters, x)) as p:
+                res = p.map_async(pool_fit, data.X, chunksize=1)
```
stuart-cls, line 416: I had previously seen that chunksize needed to be optimized, so I tested it (N=736, collagen.csv, quad-core CPU):
````
with chunksize 1: 17.978233337402344
with chunksize 10: 18.698899269104004
with chunksize 100: 17.8134982585907
with chunksize 1000: 48.91136884689331
````

To my surprise this implementation has no measureable start-up overhead on my machine! Nice.

stuart-cls (commented):
@markotoplak This is great! 

I have to admit it feels strange to have so much extra complication just for the `Constant` model but if it works 🤷‍♂️

I tested it with the default WidgetPreview data and 4 Voigt peaks (which are slow) and it stays responsive the whole time. So much so that it's impossible to tell the preview is running! Can we tap into the `progress_interrupt`/`set_progress_value()` stuff in the preview thread?

orangecontrib/spectroscopy/widgets/owpeakfit.py
markotoplak: The map function from the Pool guarantees the order.

orangecontrib/spectroscopy/widgets/owpeakfit.py
markotoplak: The start-up overhead is certainly there, but is equal in all your tests. :)

I think Chunking could only improve performance for data sets with a very low number of wavenumbers, which are fitted more easily. If we went with it, I am for making it dependant on the length of x.

orangecontrib/spectroscopy/widgets/owpeakfit.py
markotoplak: 1 `.done()` should not block, so the only blocking call is `wait()`.
2. There is no start-up time, except for the first fit, OR when previews preview could not be canceled nicely. If the optimization function used by `lmfit` was not so stubbornly written, we could ALWAYS stop it nicely. Oh yes, but then we would not even need separate processes, so... What is here is certainly a compromise. I

markotoplak:
I would not want to use progress bars for previews, because that could seem confusing: the widget is not doing anything that would be useful to other widgets. 

Feel free to add some visual indication of computing previews, if you think of something (we can always add it later).

That startup time, while not that obnoxious in use, is very obnoxious for testing... I tried one idea, and if I do not think of something better I will increase the timeouts. :)

stuart-cls:
> I would not want to use progress bars for previews, because that could seem confusing: the widget is not doing anything that would be useful to other widgets.

Makes sense to me. Let's add it later.

Thanks for the explanations above!

orangecontrib/spectroscopy/widgets/owpeakfit.py
stuart-cls: I see. Let's leave it.

markotoplak:
(With the kind help of our tests) I managed to decrease processes setup time by moving computation functions into a separate `.py` file.

The difference only shows when running the widget from canvas or in tests, but not if running `owpeakfit` directly. Which makes sense, because each process needs to import the module your application was started with + modules needed for compute.

I learned a lot doing this PR.

markotoplak:
I merged this into `master`.

@bluewin4, are you still using the widget? You had comments on responsiveness with the original release. Could you perhaps try if it works better on `master`.

@borondics, could you give `master` a spin? For now, we are missing Mac testers. :)

I there are no glaring issues, I'll release a new version.

stuart-cls:
🎉 What a big job, thanks for your hard work @markotoplak ! I learned a lot following along as well :D

I tried it from the canvas, everything feels pretty zippy now.

borondics:
Guys, thanks a lot for this! Awesome progress.

PR #594, opened 2021-11-23
Quasar mod

callumgassner:


orangecontrib/spectroscopy/preprocess/__init__.py
```diff
@@ -310,7 +310,7 @@ def transformed(self, data):
                     data.X[nans] = float("nan")
             elif self.method == Normalize.Area:
                 norm_data = Integrate(methods=self.int_method,
-                                      limits=[[self.lower, self.upper]])(data)
+                                    limits=[[self.lower, self.upper]])(data)
```
markotoplak, line 313: Please do not modify the code that is not part of your new feature.

orangecontrib/spectroscopy/preprocess/__init__.py
```diff
@@ -456,7 +456,6 @@ def __call__(self, data):
 class Interpolate(Preprocess):
     """
     Linear interpolation of the domain.
-
```
markotoplak, line 459: Also, please avoid modifications like this when they do not touch your new feature. It makes history harder to read.

callumgassner: Sorry Marko, this was an accidental pull. Please delete if possible

markotoplak: I noticed, thanks, I just commented because these things are easier to solve when the code is still smaller. :)

PR #595, opened 2021-11-24
[ENH] Atmospheric gas correction

ctroein:
Atmospheric gas correction, subtracting as much of a reference spectrum as needed to minimize the sum of squares of differences of consecutive points. Implemented as a preprocessor with options for the wavenumber ranges (which should be non-overlapping), replacing a region with a smooth curve, denoising the corrected region, and whether multiple reference spectra should be averaged first or later (with spectrum-specific weighting).

### DIFF
--- a/doc/widgets/atmcorr.md
+++ b/doc/widgets/atmcorr.md
@@ -10,7 +10,7 @@ and 2190-2480 cm<sup>-1</sup> for CO<sub>2</sub>.
 In each of these ranges (individually), the preprocessor either subtracts
 (or adds) as much of the reference spectrum as necessary to maximize the
 smoothness of the output (**Correct**), replaces the data with a smooth line
-(**Remove**) or does nothing (**Ignore**). Ranges to be corrected must not overlap.
+(**Bridge**) or does nothing (**No-op**). Ranges to be corrected must not overlap.
 
 For each range to be **Correct**ed and for each spectrum, the amount of reference
 subtracted from (or added to) the spectrum is chosen such that the sum of
@@ -25,9 +25,10 @@ references; this should be investigated and developed further.
 Optionally, the corrected ranges are smoothed with a Savitzky-Golay
 filter of user-defined **Savitzky-Golay window size** and polynomial order 3.
 
-For each range to be **Remove**d, data are replaced with a non-overshooting
-spline that merges gradually with the data at its edges (using a Tukey
-window with alpha=0.3).
+For each range to be **Bridge**d, data are replaced with a spline that merges
+gradually with the data at its edges. The spline is derived from the level and
+slope in a window of **Bridge base window size** points, while the transition
+between data and spline follows a Tukey window with alpha=0.2.
 
 **Reference spectrum**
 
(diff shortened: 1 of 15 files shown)


markotoplak:
@ctroein, thanks a lot, a much-needed addition.

I do not feel qualified to test it though. :) @stuart-cls, @borondics, perhaps you? I can help only with technical/Orange stuff if anything is needed...

Carl, could you perhaps also cite the method you used somewhere in the code for reference to the future programmers? Also, could you add a line with a short description (and citation, if applicable) into preprocess-spectra.md?

Thanks for adding tests!

borondics:
Indeed, this is a great feature! Thanks @ctroein for adding it.
As I am testing I will put my findings here. (I'm on MacOS 11.6, Orange is the latest git master)

One question about the interface: the "Use mean of references" seems unnecessary because we don't handle multiple inputs on the "Reference" input channel of the Preprocess Spectra widget, usually just average everything that arrives there. What would be the functionality if this check box is unchecked take the first spectrum?

When use the "Liver cirrhosis" dataset from Datasets and just add the preprocessor I get this:
image

Then, if I click any of the "spline" checkboxes which are not checked initially something weird happens:
image

The original spectra are never recovered even if I unclick the "spline" checkbox. Different ranges give different results.
image

stuart-cls:
Thanks for this! I don't have time to look at it right now. @amanda-cls @kaiyang-tu and @scott-cls do you want to test and compare performance of this approach?

ctroein:
Many thanks for the comments! I'm reworking the UI a bit to make it a lot clearer what things mean, and documentation has now been written. I'm just going to sort out some newly introduced bugs before pushing the changes.

About the reference: It's supposed to be a spectrum of H2O, CO2 etc, e.g. from the difference of spectra recorded with/without a person exhaling into expensive lab equipment. Maybe it would be good to provide a generic reference, but I don't know how well it would work between machines, spectral resolutions etc.

About "spline": The idea is to replace (with a spline) a whole region that's too messy to be corrected. I'm renaming it "remove", but maybe "replace" is better.

About multiple reference spectra: ideally, multiple spectra should show the natural variability so the correction can mix the most suitable spectra or at least pick the best one. At present, it mixes them in proportion to the square of how much of each one would be used if it was the only reference; this is too similar to just taking the mean of them. I was thinking about taking the linear combination that maximizes smoothness of the output, but then we could get stuff like 1000*ref1 - 1000*ref2 which is clearly wrong too. I'm considering using a nonnegative solver with all-positive or all-negative. There's an interesting 2006 paper by Bruun et al where they used the first few PCA components of atmospheric spectra, but the lesson there was also that it's tricky to do something about the rather small variability in atmospheric spectra. The first component captures something like 99%, so the rest is very nearly noise anyway.

markotoplak:
@GioB81, are you interested in trying this out?

GioB81:
> @GioB81, are you interested in trying this out?

Sure I think it will be super useful, I am a bit swamped right now but I'll check it out also with Diana ASAP :)

markotoplak:
I rebased and fixed two bugs:
- A control was not used/saved properly,
- Initial run did not use default parameters, thus did not do anything. This is likely why @borondics reported that he could not get the first data back.

markotoplak:
@ctroein, thanks for the contribution! I added some minor fixes.

To make use (and testing) of the feature easier, I am merging it. If anyone has suggestions/comments, please open new issues or pull requests.

PR #600, opened 2021-12-14
[FIX] Fix loading pre-0.6.1 workflows with explicit reader settings

markotoplak:
We were sloppy in #575. In the owfile settings, explicitly selected readers are stored with the full module and class names. Moving but not adding imports for compatibility showed the missing reader error. I remembered this bug while studying how owfile works.

I also needed to move `getx` somewhere out of `orangecontrib.spectroscopy.data` because of circular imports. `orangecontrib.spectroscopy.data.util` did not work because that still imported the parent module. I am open to better names than `orangecontrib.spectroscopy.util`.

### DIFF
--- a/orangecontrib/spectroscopy/data.py
+++ b/orangecontrib/spectroscopy/data.py
@@ -1,5 +1,4 @@
-import bottleneck
-import numpy as np
+# pylint: disable=unused-import
 
 # import for compatibility before the restructure
 import Orange.data.io  # legacy import so that file readers are registered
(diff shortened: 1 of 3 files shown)


stuart-cls:
Whoops! I didn't notice this directly but sometimes strangeness like that gets missed when you're just quickly loading an old workflow to show someone.

I won't have time to test but it looks sensible to me. As long as scripts can still use `from orangecontrib.spectroscopy.data import getx` we should be fine there.

PR #607, opened 2022-03-23
Support changed pyqtgraph 0.12.4 updateScaleBox semantics

markotoplak:
Workaround for pyqtgraph/pyqtgraph#2237


markotoplak:
Replaced with #607

PR #620, opened 2022-08-17
[FIX] Agilent visible image loading on case-sensitive systems

stuart-cls:
`agilent` module had a (case) typo in the hard-coded filenames for visible image loading. Doesn't show up on Windows as it is case-insensitive.

I added some files for testing in https://github.com/stuart-cls/python-agilent-file-formats/commit/07f278a33da008fc4e21872b9d4e98a35e02acc7 but I wasn't sure if we wanted even those small-ish `.bmp` files here. If yes, I can add a similar test to this PR

### DIFF
--- a/orangecontrib/spectroscopy/utils/agilent.py
+++ b/orangecontrib/spectroscopy/utils/agilent.py
@@ -1,4 +1,4 @@
-__version__ = "0.4.2"
+__version__ = "0.4.3"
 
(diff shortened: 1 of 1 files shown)


markotoplak:
Thank you, looks good.

If the other repository gets regular releases it could become a dependency in the future - if you have time, of course,

PR #622, opened 2022-08-29
Migrate widgets to the new deferred commit

markotoplak:


markotoplak:
`-latest` tests fail because merged https://github.com/biolab/orange-widget-base/pull/204 pointed out some errors that will need to be fixes separately.

PR #624, opened 2022-09-01
Fix a crash in athmospheric correction (in an extremely degenerate case)

markotoplak:
biolab/orange-widget-base/pull/204 pointed out an error that we did not catch before. @stuart-cls, remember when we were wondering why some exceptions in peak fitting did not fail the tests? Well, now they would.

I am surprised there was just a single error like this.


stuart-cls:
That's a tricky one! Must have to been hard to track down...

PR #625, opened 2022-09-02
[ENH] Implement equivalence of (some) spectral tranformations

markotoplak:
Two different Preprocess Spectra widgets with the same parameters for Savitzky-Golay produced transformations, that always gave equivalent results, but because nothing guaranteed that they were different they were regarded as incompatible. Implementing `__eq__` and `__hash__` on transformations educates Orange about equivalence. With Savitzky-Golay it now looks like this (both workflows would trigger a warning before the change):

dif

This feature builds on biolab/orange3#5985 and company.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -24,15 +24,15 @@ requirements:
   run:
     - python >=3.8
     - numpy >=1.20.0
-    - orange3 >=3.32.0
-    - orange-canvas-core >=0.1.24
-    - orange-widget-base >=4.16.1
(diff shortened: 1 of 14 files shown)


markotoplak:
I rebased this to master and edited a bit. Now, tests output about 3000 lines; on master, the output (full of warnings) was about 7000 lines long.

Now, the way these `__eq__` and `__hash__` are implemented is more or less final, so I'll just merge this. We can implement remaining preprocessor in new PRs. Better to have at least some done than none. :)

PR #626, opened 2022-09-13
[FIX] Agilent loading edge-cases on case-sensitive systems

stuart-cls:
A series of case-sensitivity bugs like #620. This time it affects data loading. Some were due to non-standard naming from reprocessed datasets, and some were triggered by a race condition in the directory listing code.

Again, I added tests to my [parallel repo](https://github.com/stuart-cls/python-agilent-file-formats) so I only propose the changes to the module here. I may release the module in the future but for now I will not.

### DIFF
--- a/orangecontrib/spectroscopy/utils/agilent.py
+++ b/orangecontrib/spectroscopy/utils/agilent.py
@@ -1,4 +1,4 @@
-__version__ = "0.4.3"
+__version__ = "0.4.4"
 
 import configparser
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


markotoplak:
@stuart-cls, do you have any ideas where the following error came from? I know it is not from this PR, but did perhaps any dependency change?

```
======================================================================
ERROR: test_envi_comparison (test_readers.TestAgilentReader)
----------------------------------------------------------------------
TypeError: data type 'complex256' not understood
```

PR #627, opened 2022-09-22
Fix OmnicMap's reader getPositionFromIndexAndInfo

markotoplak:
@clsandt found a map where:
- deltaX = 49.97058868408203
- x0 = -815.0
- x1 = 884.0

The number of jumps (intervals) in X is ((x1 - x0) / deltaX) which is 33.9999996946446, which means that the number of measurements nX is 33.9999996946446 + 1.

The old code wrongly rounded this down. The new code rounds before converting to int so it is more robust to numeric inaccuracies.

The reading code comes from PyMCA so I also opened a pull request there: vasole/pymca#906

### DIFF
--- a/orangecontrib/spectroscopy/utils/pymca5/OmnicMap.py
+++ b/orangecontrib/spectroscopy/utils/pymca5/OmnicMap.py
@@ -309,7 +309,7 @@ class OmnicMap(DataObject.DataObject):
         x1, y1 = ddict['Last map location']
         deltaX = ddict['Mapping stage X step size']
         deltaY = ddict['Mapping stage Y step size']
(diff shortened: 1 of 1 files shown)


markotoplak:
@stuart-cls, could you double check this changed line? I think this should be fine conceptually.

stuart-cls (approved):
@markotoplak This looks good to me. I'm surprised this hasn't come up before. Honestly range/delta seems like a backwards way to store the number of positions, but so it goes.

markotoplak:
@stuart-cls, the problem is that the format is closed. I saw specs once somewhere and there all is beautifully organized, but if you have to reverse engineer... Thanks for the review!

PR #630, opened 2022-09-23
[ENH] Integrate with a separate baseline

markotoplak:
a

As suggested by @clsandt, @borondics and I worked a bit on this one today.

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/integrate.py
+++ b/orangecontrib/spectroscopy/preprocess/integrate.py
@@ -300,7 +300,12 @@ class Integrate(Preprocess):
                 methods = [methods] * len(self.limits)
             names = self.names
             if not names:
-                names = [" - ".join("{0}".format(e) for e in l) for l in self.limits]
+                names = []
+                for l, m in zip(self.limits, methods):
+                    if m in [IntegrateFeatureSeparateBaseline]:
+                        names.append("{0} - {1} [baseline {2} - {3}]".format(*l))
+                    else:
+                        names.append(" - ".join("{0}".format(e) for e in l))
             # no names in data should be repeated
             used_names = [var.name for var in data.domain.variables + data.domain.metas]
             for i, n in enumerate(names):
--- a/orangecontrib/spectroscopy/preprocess/integrate.py
+++ b/orangecontrib/spectroscopy/preprocess/integrate.py
@@ -10,8 +10,9 @@ from Orange.preprocess.preprocess import Preprocess
 from AnyQt.QtCore import Qt
 
 from orangecontrib.spectroscopy.data import getx
-from orangecontrib.spectroscopy.preprocess.utils import nan_extend_edges_and_interpolate, CommonDomain, \
-    edge_baseline
+from orangecontrib.spectroscopy.preprocess.utils import nan_extend_edges_and_interpolate, \
+    CommonDomain, \
+    edge_baseline, linear_baseline
 
 INTEGRATE_DRAW_CURVE_WIDTH = 2
 INTEGRATE_DRAW_EDGE_WIDTH = 1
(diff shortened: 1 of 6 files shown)


markotoplak:
@borondics, does the image show what we discussed? I did not check if the computed values are correct yet: this should be written as a test.

borondics:
From the picture it looks great!

stuart-cls:
Sweet! This is an old one but will be appreciated I think.

It crashes when you use it from the Preprocess Spectra widget, I assume because it's not getting the expected number of limits.

There's a couple other places where it is assumed the limits values has length of 2:

- Name of the output variable: works, but not obvious what the 4 numbers are
- Scripting API: Now you just need to know to specify 4 limits when using this particular method. 
  `Integrate(methods=IntegrateFeatureSeparateBaseline, limits=[[1349.5, 1432.6, 910.2, 1791.5]], metas=True)` for example

Not sure if it's better to give an exception/error when provided the wrong number of limits or just silently use the limits for both (which would make it identical to "Integral from baseline")

Sorry if too much feedback on a draft PR, I got excited :)

markotoplak:
@stuart-cls,

> Name of the output variable: works, but not obvious what the 4 numbers are

Hmm, I fixed this, but in practice, we need to rework what the final names are. I do not like them (they are too precise and do not convey integral types). Perhaps each integral should have an editable field where a user can optionally set it. What do you think?

> Scripting API: Now you just need to know to specify 4 limits when using this particular method. 

Yes. :) I think the current interface does not work properly. We'd need an easier way to define separate integrals.

> Not sure if it's better to give an exception/error when provided the wrong number of limits or just silently use the limits for both (which would make it identical to "Integral from baseline")

I would not worry about this. Now it is an exception. :)

markotoplak:
I think this one is not ready for review. @stuart-cls, @borondics, can anyone play with it?

stuart-cls:
> I do not like them (they are too precise and do not convey integral types).

Agreed, they don't quite contain enough information to recreate what was done.

> Perhaps each integral should have an editable field where a user can optionally set it. What do you think?

This would be excellent! I'd forgotten but I've had this as a feature request from users on multiple occasions.

I will play with it some more again, probably next week.

stuart-cls:
I tried this a bit more, I don't have any more comments about it specifically, seems to work well!

Two ancillary comments though:
1. I think the user-editable labels for the peak analysis / Integrate would be a great (separate) feature. Probably also useful in Peak Fit which uses the same GUI, but I'd have to check that.
2. I just noticed the Preprocess Spectra Integrate deprecation (kind of hidden in the commit message!). I agree the Integrate widget is much better. My only concern is the (real) use case with TileFile (which is on my mind obviously) where you might want to do a global baseline correction etc, then integrate the peak. I'm not aware of any way to chain two `PreprocessorList`s in the gui (can be done in a Python script, but that would be a step backwards). 
My first thought was to change the TileFile preprocessor Input to Multiple type. Correct me if I'm wrong, but I think the ordering in Multi-inputs is arbitrary and couldn't be used reliably in this way?

markotoplak:
Wow, I think you just figured an elegant solution to preprocessor chaining we were looking for some time ago. :D

The multi-inputs are now stable and predictible (they are in connection order), just currently not shown in the canvas. But I think we could fix this. So that Tile File could indeed use multi-inputs already. Within, it could already shown info about connected preprocessor lists. And bam, done! If we are going to use multi-inputs regularly we should also somehow update the canvas to enumerate such connections, but that would not influence functionality.

I did never plan to destroy any compatibility, at least not for a long time, and surely not until we have a solution to your example. But it is at least a warning that users should look into other ways of doing integrals.

stuart-cls:
Great! I will make two issues so we don't lose these ideas :D

markotoplak:
Merging this. If anyone finds bugs, open issues. :) @borondics, play with it, when you'll have time, please.

clsandt:
> a
> 
> As suggested by @clsandt, @borondics and I worked a bit on this one today.

Thanks Marko for taking care of this. This already looks great.

Would it be possible to make the lines a different color depending whether they are for the area or for the baseline instead of text label?
I would prefer to have different colors (or having dashed/solid lines) instead of the Low limit/high limit written in the spectral windows. The labels might end-up hiding some spectral features you would want to see.
Is that easily doable?

PR #632, opened 2022-10-03
[FIX] Agilent Tile Reader: Generate correct y-axis positions

stuart-cls:
I found that the `agilentMosaicTileReader` had an error in the way it calculated the y-axis positions. Although the agilent data xy positions are arbitrary from (0,0), it means the TileReader and the normal reader generate _different_ positions.

I'm asking for input because this has implications for backwards compatibility. In the case where someone uses the **OWTileFile** widget and then selects regions in **OWHyper**, the selection will be incorrect when that workflow is loaded with this code.

Thoughts?

Edit: This now fixes both the coordinates and the table row ordering so that the two readers result in identical output.

### DIFF
--- a/orangecontrib/spectroscopy/io/agilent.py
+++ b/orangecontrib/spectroscopy/io/agilent.py
@@ -183,7 +183,8 @@ class agilentMosaicTileReader(FileFormat, TileFileFormat):
         am = agilentMosaicTiles(self.filename)
         info = am.info
         tiles = am.tiles
-        ytiles = am.tiles.shape[0]
+        xtiles = am.tiles.shape[0]
+        ytiles = am.tiles.shape[1]
 
         features = info['wavenumbers']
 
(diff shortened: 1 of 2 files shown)


markotoplak:
So, is it still a problem after your edit? :)

Hmm, it is always hard to decide whether to keep backward compatibility in cases like this. I try to do it if it is reasonably possible (with that "if old settings are different behave differently trick"). But sometimes, if we judge the use-case to be particularly rare, I am fine with forgoing compatibility. Your call.

markotoplak:
Yes, as you say, you'd need different reading code...

stuart-cls:
Decided to skip backwards compatibility which would add significant complexity. Impact is only on single-dataset archival workflows, which can be opened using old versions if necessary.

markotoplak:
Thanks!

PR #633, opened 2022-10-05
github actions: run tests monthly and on-demand

stuart-cls:
The "oldest" tests in #632 aren't passing, but there's no way currently to check if the master branch CI tests pass. In travis we could trigger a run (and run on a schedule) so that's what this change does.

My cron-fu is not strong but I think it's set to run on the 1st of each month.


markotoplak:
I updated oldest tests in #630. I thought that would be merged quckly. :)

Issue #634, opened 2022-10-06
OWTileFile: Multi-input for Preprocessor chaining


stuart-cls:
From #630 

Preprocessor Input should be changed to Multi-input to allow preprocessor chaining, for example baseline correction, then integral.

Widget already gives feedback on the connected Preprocessor(List), but should check it still looks readable.

PR #637, opened 2022-10-20
[FIX] Integrate: Fix call to new deferred commit

stuart-cls:
While working on #634 I found a call to commit that got missed in #622

I wasn't sure which call ("now" or "deferred") was appropriate, I chose "now"


markotoplak:
I changed to "deferred". So the current "standard" is to use deferred anywhere except for the new data received or explicit commit button clicks.

PR #638, opened 2022-10-20
[ENH] OWTileFile: Multi-input for Preprocessor chaining

stuart-cls:
Implements #634, depends on #637 for OWIntegrate testing.

A few notes/items for discussion:
 - **OWIntegrate** "Output as metas" checkbox is not reflected in the preprocessor box
   - I found looking at all the PreprocessorList and PreprocessorListMoveMetas was distracting, but the preprocessors from OWIntegrate all have meta=True set, and then move metas around later. It works fine, but isn't shown.
   - Coming back to #439, we had talked about just getting rid of this checkbox and have the Preprocessor output simply always do `move_metas=True` + `metas=True` (which is the equivalent of multi-method `metas=False` from Preprocess Spectra)
 - I found that the WidgetPreview doesn't support multi-input: a bug? I worked around it with some awkward syntax.
 - I didn't update the code-copy from OWFile, if this is worth doing I will do in a separate PR

Feedback welcome!

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_tile_reader.py
+++ b/orangecontrib/spectroscopy/tests/test_tile_reader.py
@@ -87,10 +87,10 @@ class TestTileReaderWidget(WidgetTest):
         pp_out = self.get_output("Preprocessor", widget=self.preproc_widget)
         self.send_signal("Preprocessor", pp_out, widget=self.widget)
         # Single Input
-        self.assertEqual(self.widget.preprocessor.preprocessors, pp_out.preprocessors)
+        self.assertEqual(self.widget.preprocessor.preprocessors[0], pp_out)
         # Preprocessor members match editor model
         pp_from_model = create_preprocessor(self.preproc_widget.preprocessormodel.item(0), None)
-        pp_tile = self.widget.preprocessor.preprocessors[0]
+        pp_tile = self.widget.preprocessor.preprocessors[0].preprocessors[0]
         self.assertIsInstance(pp_tile, type(pp_from_model))
         # MultiInput with OWIntegrate
         self.int_widget = self.create_widget(OWIntegrate)
@@ -99,4 +99,4 @@ class TestTileReaderWidget(WidgetTest):
         pp_out_2 = self.get_output("Preprocessor", widget=self.int_widget)
         self.send_signal("Preprocessor", pp_out_2, 2, widget=self.widget)
         self.assertEqual(self.widget.preprocessor.preprocessors,
-                         pp_out.preprocessors + pp_out_2.preprocessors)
+                         [pp_out, pp_out_2])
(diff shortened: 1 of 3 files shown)


markotoplak:
>       * Coming back to [Integrate: Add "data" output which always contains spectra + results in meta #439](https://github.com/Quasars/orange-spectroscopy/pull/439), we had talked about just getting rid of this checkbox and have the Preprocessor output simply always do `move_metas=True` + `metas=True` (which is the equivalent of multi-method `metas=False` from Preprocess Spectra)

I am nor sure if we should really do that still. I think that what PCA widget does with its two semi-equivalent outputs is confusing. So the problem that #439 would aim to solve is to avoid a select columns (if results were in the metas) or a possible merge widget, if they were not, right?

Do I understand the following correctly: when using the Tile Reader the not-in-metas output is actually preferred because it saves memory when generating the final table? And that is the reason why we still have it?

>     * I found that the WidgetPreview doesn't support multi-input: a bug? I worked around it with some awkward syntax.

Yes, I also remember using the same syntax once. No-one thought of a nicer way, I think.
 
>     * I didn't update the code-copy from OWFile, if this is worth doing I will do in a separate PR

Which copy code do you mean?

stuart-cls:
> So the problem that https://github.com/Quasars/orange-spectroscopy/pull/439 would aim to solve is to avoid a select columns (if results were in the metas) or a possible merge widget, if they were not, right?

Yes, I think that was the idea. Honestly I can't remember the workflows where I was using this very well anymore, so I'm okay to drop it and revisit if we think of a good reason later.

> Do I understand the following correctly: when using the Tile Reader the not-in-metas output is actually preferred because it saves memory when generating the final table? And that is the reason why we still have it?

Yes, that is the entire point of the Tile Reader. The integrate one is tricky because while the `Integrate` preprocessor supports multiple integration regions (this is what comes out of the Preprocess Spectra:Preprocessors output), the **OWIntegrate** widget does not (in order to support sequential independent integration regions). And so we have the `PreprocessorListMoveMetas` work-around. Which is all fine, I was just pointing out that the text representation in the info box in TileFile is now not quite accurate.

I should also say that I often use the not-in-metas output of the Integrate widget anyway, I find it conceptually simpler to think of it as a dimensionality-reduction widget. If I need the original spectra, I just use the nice Select by ID widget.

> Which copy code do you mean?

OWTileFile is just a sub-classed OWFile. But OWFile is not allowed to be sub-classed, so I just copied the code from OWFile. It's not pretty and I think is giving some warnings because it now lags behind the current OWFile implementation.

markotoplak:
> I should also say that I often use the not-in-metas output of the Integrate widget anyway, I find it conceptually simpler to think of it as a dimensionality-reduction widget. If I need the original spectra, I just use the nice Select by ID widget.

I agree that this type of use is conceptually the cleanest. But might be hard to grasp. :)

stuart-cls:
Great! 

I was sure I did have a widget warning for that :D I'm fixing it along with another bug I found, PR shortly

PR #641, opened 2022-11-01
[WIP][FIX] irfft: ramp correction and apodization fix for asymmetric interferograms

stuart-cls:
Fixes an old error in the apodization function generation for asymmetric interferograms, found while adding a ramp correction to avoid double-counting in the symmetric section. Illustration of the issue:

new-apod-ramp

The above figure illustrates how the original apodization code was handling the taper to zero needed in asymmetric interferograms. While it does bring the amplitude to zero at the edge of the short side, it obviously badly violates the symmetry requirements. This would not have been obvious on the original symmetric test interferograms.

### DIFF
--- a/orangecontrib/spectroscopy/irfft.py
+++ b/orangecontrib/spectroscopy/irfft.py
@@ -55,6 +55,12 @@ def find_zpd(ifg, peak_search):
     else:
         raise NotImplementedError
 
+def wing_size(ifg, zpd):
+    """Calculate negative and positive wing size (including zpd point)"""
+    wing_n = zpd + 1
+    wing_p = ifg.shape[-1] - zpd
+    return wing_n, wing_p
+
 def apodize(ifg, zpd, apod_func):
     """
     Perform apodization of asymmetric interferogram using selected apodization
@@ -80,8 +86,7 @@ def apodize(ifg, zpd, apod_func):
     # Calculate negative and positive wing size
     # correcting zpd from 0-based index
     ifg_N = ifg.shape[-1]
-    wing_n = zpd + 1
-    wing_p = ifg_N - zpd
+    wing_n, wing_p = wing_size(ifg, zpd)
     # Use larger wing to define apodization window width
     M = max(wing_n, wing_p)
 
@@ -131,6 +136,18 @@ def apodize(ifg, zpd, apod_func):
         raise ValueError("Apodization function size mismatch: %s" % e)
 
     return ifg_apod
+
+def ramp(ifg, zpd):
+    wing_n, wing_p = wing_size(ifg, zpd)
+    # Use smaller wing to define symmetric subset size
+    M = min(wing_n, wing_p)
+    ramp = np.linspace(0, 1, 2 * M - 1)
+    if wing_n < wing_p:
+        ifg[..., :ramp.shape[0]] *= ramp
+    else:
+        ifg[..., -ramp.shape[0]:] *= ramp[::-1]
+
+    return ifg
 
 def _zero_fill_size(ifg_N, zff):
     # Calculate desired array size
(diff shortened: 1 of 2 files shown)


borondics:
@stuart-cls I think the "original" (on your figure) asymmetric apodization was meant for SNOM IFGs and the intention was to have that shape, where not only the position of the ZPD is offset but also the apodization. In my understanding the reason for this is that the near field signal is contained only on one side of the ZPD and  the other should be down-weighted. There is some literature on this: http://hdl.handle.net/10810/26669

NeaSpec does something very similar, but their calculations are proprietary.

However, I think there is some actual scientific discussion missing... @raul-freitas, @levandoskije, what do you think?

stuart-cls:
@borondics I'm pretty sure this part of the code predates SNOM!

That said, I definitely don't want to break anything correct. I will simply note that for asymmetric interferograms in an FT-IR, downweighting the short side via shrinking the apodization window ("old") is not the same as downweighting the short side via the ramp ("new + ramp"), and the latter is the correct way. 

I'll review the literature you sent, I'm also waiting on other feedback.

If you have any before / after data of what Neaspec is doing, please send it along: we can compare the two approaches. This is basically what I did for the Agilent imaging data to check for correctness.

raul-freitas:
Hello @borondics and @stuart-cls.
I will try to answer you soon with some old data. I will get back to this subject quite soon and I hope to have some direct feedback from the company, as we are processing raw interferograms (.gsf files). I remember playing a bit in the past and comparing my processing with Neaspec processing, I could not exactly reproduce (at least the high frequency features).

Maybe we can plan some synchronized activities soon. I have two new folks that will give some support to Orange-Quasar in here.

borondics:
I was just browsing SciPy for something and realized that many (all?) of the [apodization](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.windows.blackmanharris.html#scipy.signal.windows.blackmanharris) functions are implemented there. Since SciPy is a dependency, should we simply use those functions to decrease the code?

stuart-cls:
@borondics I'm aware of those functions. Originally I wanted to keep the dependencies of `irfft` minimal for easy re-use in other projects. `scipy` is not so annoying to install anymore but I'd still rather keep that. 

Dependencies aside, I looked at how those windows are implemented this time around and they don't work for either of the types of apodization we would like to do.

Which brings me to the summary of the current state of this PR: based on our discussions offline, the thesis above, and feedback from others, I think we will have to distinguish between interferometer types:

1. Symmetric (Michelson-type) interferometer
  a. Fully symmetric interferogram ("Symmetric" in current GUI)
  b. Truncated symmetric interferogram ("Asymmetric" in current GUI)
2. Asymmetric (s-SNOM type) interferometer

where type (1) will use this symmetric + ramp approach and type (2) will use the existing approach of different apodization functions for each wing.

ngergihun:
After @stuart-cls mentioned this existing issue in  #https://github.com/Quasars/orange-spectroscopy/pull/809, probably it is time to revisit this. As discussed above this is a very very important. The asymmetry apodization (which is currently in the widget) introduces serious artifacts.

However, it is necessary for SNOM interferograms. [https://doi.org/10.1364/oe.520793](https://doi.org/10.1364/oe.520793)

There should be an option in the apodization section. Either just a checkbox for asymmetric apodization or a radiobutton, or dropdown to choose.

ngergihun:
@stuart-cls , yes they are still giving different results. Hmmmm.... I will try to send something more useful report in a few hours....
Also, what I noticed, that are is a difference in the number of features, that should not be the case. The ifg I used has 530 points, and my code with zerofilling of 2 gives back 1060 points, while the owfft results in 1025 points.

stuart-cls:
Two outstanding things from @ngergihun observations:

- [ ] Change default in `irfft` to `apod_asym = True` so that the output doesn't change by surprise. Raise a warning if unset to allow porting to the new interface.
- [ ] Figure out what's going on with the zero filling difference.
- [ ] Regression in match to agilent data (phase correction?)

PR #642, opened 2022-11-02
[FIX] Ascii column reader: allow comments starting with ";"

markotoplak:
Such comments are contained in the files exported by the Kazan viewer.

This change does not break previous kind of comments (starting with "#")


markotoplak:
The tests seem to fail on latest Orange, but not because of this PR.

PR #643, opened 2022-11-03
[NOMERGE] Trying to understand crashes on -latest

markotoplak:


markotoplak:
I finally figured out what is happening. We deprecated `get_column_view`. That should not be problematic, we would only get 
warning, but we have `ORANGE_DEPRECATION_ERROR` turned on in our tests. This should cause tests to fail, but no segfault.

Here it gets interesting. We got the deprecation warning, which has now changed to error (exception) within Qt's event handling. Exceptions there in Qt event handling: a segfault.

I am so happy to see a segfault that I can explain.

stuart-cls:
I'm glad you figured it out! This was a tricky one for sure.

PR #644, opened 2022-11-08
On -latest tests, do not crash in case of OrangeDeprecationWarnings

markotoplak:
As described in #378, OrangeDeprecationWarnings that came into Qt's event loop caused segfaults (Qt does not like exceptions).

This PR disables crashes on OrangeDeprecationWarnings for unreleased Orange versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,12 @@ jobs:
       - name: Test with Tox
         run: |
           tox -e ${{ matrix.tox_env }}
+        env:
+          # Raise deprecations as errors in our tests only when testing orange-oldest and orange-released.
+          ORANGE_DEPRECATIONS_ERROR: "${{ matrix.tox_env != 'py-orange-latest' && '1' || '' }}"
+          # Need this otherwise unittest installs a warning filter that overrides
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks @JakaKokosar for writing the actual testing code for this.

stuart-cls:
It seems like this only eliminates the instigation of the problem and doesn't fix the event loop segfaults. Is there any solution to have those crashes visible in a way that is easier to debug? #643 seemed really hard to debug

markotoplak:
Yea, but the real reason is that I was too inexperienced with this kind of errors. If I looked at the trace I could see that this was run as a result of an event and then I could have remembered that this is the default Qt behaviour. :)

We actually wanted crashes, remember https://github.com/Quasars/orange-spectroscopy/pull/559#issuecomment-848300718? That did not crash because pyqtgraph at its import added some guardind code. Then, Aleš removed that workaround so that we can see bugs like this most easily.

I do not see a way to have it both ways currently. Perhaps if there is a way to not make it segfault but only fail tests, but I do not see how...

stuart-cls:
Oh right, good point. Better to take the crashes then!

PR #645, opened 2022-11-08
Replace deprecated get_column_view

markotoplak:
The replacement, `get_column`, also takes care of the conversion into the correct data type.


markotoplak:
@stuart-cls, look, simpler averaging. :) Thanks to biolab/orange3#6058

stuart-cls:
@markotoplak Excellent! Thanks for pointing it out, I reach for the get_column_view often in scripts: this is much nicer to use.

I also love changes that get rid of long explanatory comments :D

PR #646, opened 2022-11-09
[FIX] HyperSpectra now works with integrals with separate baselines

markotoplak:
Thanks to @clsandt for pointing out the bug. At first I just wanted to disable this functionality, but then I decided to actually support it.


markotoplak:
Feri suggested different colours for baseline selectors. I chose orange. Feel free to change it in another PR.

PR #647, opened 2022-11-10
[FIX] OWTileFile warn on preprocessor change

stuart-cls:
Following #638 

[Edit: Left None bug for another day, see #648]

I found a bug where the code that runs the preprocessor() wasn't checking for `PreprocessorList`s with None in them (or empty lists), so `PreprocessorList` tries to call None(). But I kept thinking of more weird edge cases from the multiple inputs, so I decided instead to just not add None in the first place (which is cleaner, I think).

But now when you have a `None` connected (empty Preprocess Spectra), and then change it to a preprocessor (Add an editor to the Preprocess Spectra), it crashes because the index state from the `MultiInput` doesn't match the PreprocessorList length.

@markotoplak  Before I go any further down this rabbit hole, any thoughts?

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owtilefile.py
+++ b/orangecontrib/spectroscopy/widgets/owtilefile.py
@@ -82,6 +82,8 @@ class OWTilefile(widget.OWWidget, RecentPathsWComboMixin):
     class Warning(widget.OWWidget.Warning):
         no_preprocessor = Msg("No preprocessor on input."
                               " Press Reload to load anyway.")
+        new_preprocessor = Msg("Preprocessor has changed."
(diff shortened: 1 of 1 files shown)


markotoplak:
Hmm, to me it seems dangerous to ignore `None`s withing handling inputs. I do not have an informed-enough idea what is happening in the background to argue it. Maybe that would make a problem if doing insert/set with or without nones?

Anyway, I'd rather clean this up later.

stuart-cls:
@markotoplak I guess the status quo crashes fairly quietly in a way that is easy to recover, just add a preprocessor. So I'll split out the warning stuff and leave the None for another day as you suggested

PR #649, opened 2022-11-29
[ENH] PyQt6 and Python 3.10 support

markotoplak:
I tested Orange-Spectroscopy in a more modern environment and this were the minimum required changes for tests to pass.

Some are due to PyQt6 and some are due to Python 3.10.

It would be cool if some developers would start using these more modern versions because our test do not have perfect coverage so other stuff can go wrong too.


markotoplak:
@PrimozGodec, just for context about our discussion today, this is what I had to fix. All minor, but necessary.

PR #653, opened 2022-12-22
[FIX] io.opus: Convert visible image loading Exceptions to warnings

stuart-cls:
This makes a whole class of opusFC bugs less annoying to users.

Also today I learned that warnings are nicely propagated up to the File widget!


markotoplak:
Thanks!

PR #654, opened 2022-12-22
Standard map vars

stuart-cls:
Working on #615 reminded me of a class of little bugs I've been meaning to address: standard "map_x" / "map_y" variables

First commit replaces all "map_x" / "map_y" strings with constants
~Second commit shows how this can be used to reliably plot XY data in Hyperspectra if both these vars are present (otherwise reverts to existing behaviour.~

~If this works, I'll also apply to the other mapping-aware widgets (Bin, Stack, Polar, etc)~


stuart-cls:
Based on @borondics comments in #739, I simplified this PR just to focus on the constants. I'll open another to fix the very annoying bug when Hyperspectra plots vs (clusters, map_x) or something

stuart-cls:
I chose to leave the line-too-long lints.

stuart-cls:
One change at a time :)

PR #657, opened 2023-01-13
Adapt for deprecated colorpalette

markotoplak:


markotoplak:
Following biolab/orange3#6135

PR #660, opened 2023-02-09
[FIX] Multifile: fix a bug when wavenumbers differ for less than 1e-6

### LINKED ISSUE #659: Multifile different size bug
If there are multiple files read with multifile and they are not the same length / number of points, the widget crashes and doesn't give a meaningful Error message.

markotoplak:
Fixes #659 

The problem appears because wavenumbers are compared as numbers at one point and as strings at another (using "%f", so rounded). 

If wavenumbers collide when converting to string, the required precision is computed and used instead. Some wavenumbers will there be more precise than other.

I tried to keep perfect backward compatibility - whatever could be read should be read exactly the same now.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owmultifile.py
+++ b/orangecontrib/spectroscopy/widgets/owmultifile.py
@@ -45,6 +45,13 @@ def numpy_union_keep_order(A, B):
 def decimals_neeeded_for_unique_str(l):
     min_diff = min(np.abs(np.diff(sorted(l))))
     log_diff = math.log(min_diff, 10)
+    # We need to avoid possible degenerate cases like (1.5, 2.5):
+    # the log of the difference (1.0) is 0, but rounded to zero
+    # decimals we would get (2, 2) due to floating point rounding behavior.
+    # Therefore, slightly increase the number of needed decimals.
+    # The number was selected such that (1.5, 2.5) returned 1 decimal,
+    # and (1.49, 2.51) returned 0.
+    log_diff = log_diff - 0.008
     decimals = max(0, math.ceil(-log_diff))
     return decimals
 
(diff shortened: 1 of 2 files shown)


markotoplak:
@borondics, I think I fixed it. could you please test this PR with your data?

markotoplak:
In the future, we could reuse parts of the code in other readers - the same code could help us decrease the number of decimals we use for features. But that could create compatibility problems for old workflows because widgets store feature names, so we'll need to be careful and at least consider that...

orangecontrib/spectroscopy/widgets/owmultifile.py
```diff
@@ -41,6 +42,38 @@ def numpy_union_keep_order(A, B):
     return np.concatenate((A, to_add))
 
 
+def decimals_neeeded_for_unique_str(l):
+    min_diff = min(np.abs(np.diff(sorted(l))))
+    log_diff = math.log(min_diff, 10)
+    decimals = max(0, math.ceil(-log_diff))
```
markotoplak, line 48: I hope this is fine. :) Could anyone find a counterexample?

markotoplak: I woke up with an edge case where this would not work and have fixed the code accordingly.

markotoplak:
The current code should ensure unique wavenumbers. `decimals_neeeded_for_unique_str` should return the number of decimals that would construct unique strings.

Then, an additional decimal is added for increased precision, so floats 1.49 and 2.51 yield strings "1.5" and "2.5" instead of "1" and "3" (without the added decimal).

borondics:
Using [73489ac](https://github.com/Quasars/orange-spectroscopy/commit/73489ac90234887aa5209729b22b8864cce03738) the Multifile reader doesn't fail but the data is still not completely OK. If you check the Data Table, the WNs are still not equal enough probably and the two different length spectra go to different blocks.

image

Would it be possible to use the same columns for the WNs that are common and have "?" only for those that are missing in the shorter dataset?

markotoplak:
This bug was that wns should not be equal but Multifile tried to make them equal at the name level. Now it does not.

What you report is expected and intentional behavior, which was done so to be as general as possible. And it is exactly how the widget would work before the fix.
 
To solve the problem of different blocks we would need to interpolate while reading the files, which for now we do not do. You can interpolate in a separate widget.

To do what you wish, we'd have to make a decision of what to interpolate to, at least, and give users options to use that automatic interpolation or not. I see the benefit of this in terms of speed and memory, but in your particular case Interpolate will suffice.

PR #666, opened 2023-03-09
[ENH] Spectra: waterfall plot

markotoplak:
An option to show spectra in a stacked form. For now we do not have parameters how much to move the spectra. but that can be added later.


stuart-cls:
Sorry to review this after you merged it!

This does not do what I expected. If I take the collagen dataset, average by type and then plot with this on, I get 
image

I would expect to get something like this:
image

Is this just the parameters to move that you mention? 

In any case, I think the default should be more like the second figure, the first is just some kind of weird normalization :)

markotoplak:
I merged to that it gets more testing. :)

Thanks for looking into this. So we should have a bigger step? For now it is, I think (maxy-miny)*0.1, so like a 10% move. I do think this should become a parameter, but what default would you suggest? Feel free to play with the waterfall_constant and report what works well for you.

PR #667, opened 2023-03-16
[FIX] owinterpolated: Call commit.deferred in line edit callbacks

stuart-cls:
I guess I don't use these boxes that often :)

I also noticed searching around for any other instances of this that OWAverage is still using the old commit style. I didn't touch it in this PR.


markotoplak:
Whoops, this one is on me. Thanks for the fix!

Issue #547, opened 2021-03-25
Missing parts of window(s)...


borondics:
On my mac I don't see the Menu window of Spectra properly. The key shortcuts should appear on the right of the options but they are missing. :~(
![image](https://user-images.githubusercontent.com/6304794/112466446-30cf3a80-8d66-11eb-96cc-90991323ce09.png)

PR #673, opened 2023-04-11
[ENH] Improved the context menu display on Mac

borondics:
Previously, the context menu of OWSpectra didn't display keyboard shortcuts on Mac (Ventura) and the window edges were very close to the elements. 

image

Now this is changed to the following:

image

Needs testing on Linux and Windows.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owhyper.py
+++ b/orangecontrib/spectroscopy/widgets/owhyper.py
@@ -677,7 +677,7 @@ class ImagePlot(QWidget, OWComponent, SelectionGroupMixin,
 
         choose_xy = QWidgetAction(self)
         box = gui.vBox(self)
-        box.setContentsMargins(20, 0, 10, 0)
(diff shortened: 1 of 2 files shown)


borondics:
Should fix #547

markotoplak:
The added spacing on the left makes HyperSpectra looks worse on Linux. The problem is that different platforms use different spacing to draw things so we should never hardcode it. 

image

To make spacing symmetric, I recommend using `box.setContentsMargins(10, 0, 10, 0)` or even `5, 0, 5, 0`.

Interesting that vertical spacing is so strange on Macs...

borondics:
Can we hunt down the parameters and reference them dynamically? I don't know what type of spaces are added to different elements in that Menu window.

markotoplak:
> Can we hunt down the parameters and reference them dynamically? I don't know what type of spaces are added to different elements in that Menu window.

Maybe, but just maybe. Who knows, what we'll forget. We are abusing the menus here anyway.

PR #678, opened 2023-06-29
A utility function that can easily replace wavenumbers (x)

markotoplak:
Renames are saved into the domain so that also new data can be transformed in the same way.

This was Feri's idea. He asked how to transform X and I figured it would be, for starter, nice to have a utility function that does that.


borondics:
Seems to work great!

Thanks a lot!!!

image

PR #682, opened 2023-08-31
update macos runner

markotoplak:
The old machine is not available anymore.


markotoplak:
Falling tests are due to yesterday's pandas update...

PR #683, opened 2023-09-05
[ENH] Spectra: individual display in its own thread

markotoplak:
This prevents blocking canvas with Dask data sets.

This change is independent of dask and can be tested with normal data sets. This PR changes `CurvePlot`, the basic display components, thus all widgets using it need to be tested.


borondics:
Seems to work fine for me.

PR #687, opened 2023-09-15
HyperSpectra: handle degenerate (all nan) coordinates and data

### LINKED ISSUE #686: Hyperspectra doesn't like to work with the SNR widget
The output of the SNR widget is not well handled in Hyperspectra. It might be related to the number of rows in the data table.

markotoplak:
Fixes #686

Handle (and also test) degenerate inputs.

#686 described errors that happened for the SNR widget because the widget is not meant to be used on normal data (without data point repetitions). HyperSpectra now does not crash but in the (before crashing) cases does not do anything useful either.


borondics:
The workflow was
image

using the "liver spectroscopy dataset"

markotoplak:
@borondics, this was not merged yet. Could you check if it works with this branch?

markotoplak:
Ok, I am just merging this into master, and test there. :)

PR #695, opened 2023-10-20
Test why peakfit started crashing on Windows and Macs

markotoplak:


markotoplak:
Closing because it no longer applies (tests now pass)

PR #699, opened 2023-12-04
Unify cuts Issue#661

nicholsonbt:


markotoplak:
Thanks for the PR!

Failing tests on Win/Mac are not the problem with this PR.

To avoid silent wrongly read new setting on older version of Quasar, I suggest also renaming `orangecontrib.spectroscopy.cut`. Then, users will be warned when opening a too-new workflow.

nicholsonbt:
The name is changed from `orangecontrib.infrared.cut`, but instead of warning the user that a too-new workflow is being opened, it's silently removing the new editors from the preprocessor widget.

Is there an existing function to catch any editors with unknown names and output a warning?

markotoplak:
Thanks. I'll make a separate PR which will make out-of-date installations in the future explicit, but can not see anything that can be done here.

PR #703, opened 2024-01-17
Fix errors caused by NaN when computing hashes

markotoplak:
Hashing NaN in python can yield different results. Therefore NaNs are replaced with None before hashing in __hash__ functions of transformations.


markotoplak:
All iks well again.

Issue #706, opened 2024-02-01
HyperSpectra: invalid contexts can be matched to a data 


markotoplak:
HyperSpectra has comboboxes to choose feature. Some of the accept all features, and some are limited to metas and classes.

All these settings are linked to a `DomainContextHandler`. Therefore, old settings could cause newly connected data to crash because of #6721.

This was originally reported by @clsandt.

PR #707, opened 2024-02-05
[FIX] Avoid some context crashes: exclude_attributes=True

markotoplak:
Some contexts match even though feature-selection combo boxes do not show the feature. Problematic settings are those where combo boxes are limited to specific parts of the table (attributes, class_vars, or metas).

This partial solution follows from https://github.com/biolab/orange3/issues/6721#issuecomment-1923826563. This does not solve the problem completely, but if problem appears, adding a new widget instance allows the user to continue.

Also, there is another minor issue: exclude_attributes won't allow users to reuse contexts where class vars were used in the fields (current implementation handles attributes and class vars in the same way), but saving and loading works properly (because of perfect matches).

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owhyper.py
+++ b/orangecontrib/spectroscopy/tests/test_owhyper.py
@@ -386,12 +386,14 @@ class TestOWHyper(WidgetTest):
         self.assertTrue(self.widget.Information.not_shown.is_shown())
 
     def test_migrate_context_feature_color(self):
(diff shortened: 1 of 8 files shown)


markotoplak:
Partial solution to #706.

markotoplak:
This does not fix crashes that appear if the user chose a class variable in the boxes (I tried by coloring Iris data in Spectra), because attributes and classes are handled in a package.

That would have to be fixed by making a new context handler.

PR #713, opened 2024-03-26
[FIX] Disable __eq__ and __hash__ because it can be slow

markotoplak:
It can be so slow that it appears Orange has crashed.

With @borondics we had a workshop in Berlin and preprocessing appeared to crash for most students. We thus downgraded to a previous version.

After debugging I saw that a series of preprocessors causes domain comparisons to be very long (for example, after normalization, each feature in turn references the whole original domain, and then equality checking just multiplies).


markotoplak:
I intend to merge this and re-release.

PR #716, opened 2024-04-25
[FIX] Peak fit: support lmfit 1.3.1

markotoplak:
Remove constant hack. It is no longer necessary and was failing because that hack was incompatible with 1.3.


stuart-cls (approved):
🥳 So nice to remove cruft! Looks good to me, works for a brief test.

PR #718, opened 2024-05-03
Extend shift spectra

matyson:
Add a scale factor to the CurveShift preprocessor in order to enable scaling the spectra amplitude, now results in a form:
res = scale * spectra + amount.  Also, rename it into LinearTransform for coherence.

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_preprocess.py
+++ b/orangecontrib/spectroscopy/tests/test_preprocess.py
@@ -11,7 +11,7 @@ from orangecontrib.spectroscopy.data import getx
 from orangecontrib.spectroscopy.preprocess import Absorbance, Transmittance, \
     Integrate, Interpolate, Cut, SavitzkyGolayFiltering, \
     GaussianSmoothing, PCADenoising, RubberbandBaseline, \
-    Normalize, LinearBaseline, LinearTransform, EMSC, MissingReferenceException, \
+    Normalize, LinearBaseline, ShiftAndScale, EMSC, MissingReferenceException, \
     WrongReferenceException, NormalizeReference, XASnormalization, ExtractEXAFS, \
     PreprocessException, NormalizePhaseReference, Despike, SpSubtract
 from orangecontrib.spectroscopy.preprocess.als import ALSP, ARPLS, AIRPLS
@@ -58,7 +58,7 @@ PREPROCESSORS_INDEPENDENT_SAMPLES = [
     Normalize(method=Normalize.Vector),
     Normalize(method=Normalize.Area, int_method=Integrate.PeakMax, lower=0, upper=10000),
     Normalize(method=Normalize.MinMax),
-    LinearTransform(1, 2),
+    ShiftAndScale(1, 2),
     Despike(threshold=5, cutoff=60, dis=5),
     ALSP(lam=100E+6, itermax=5, p=0.5),
     ARPLS(lam=100E+5, itermax=5, ratio=0.5),
@@ -588,11 +588,11 @@ class TestPCADenoising(unittest.TestCase):
                                         [4.75015528, 3.15366444, 1.46254138, 0.23693223]])
 
 
-class TestLinearTransform(unittest.TestCase):
+class TestShiftAndScale(unittest.TestCase):
 
     def test_simple(self):
         data = Table.from_numpy(None, [[1.0, 2.0, 3.0, 4.0]])
-        f = LinearTransform(amount=1.1, scale=2.)
+        f = ShiftAndScale(offset=1.1, scale=2.)
         fdata = f(data)
         np.testing.assert_almost_equal(fdata.X,
                                        [[3.1, 5.1, 7.1, 9.1]])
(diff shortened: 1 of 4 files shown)


orangecontrib/spectroscopy/widgets/preprocessors/misc.py
```diff
@@ -295,34 +295,40 @@ def createinstance(cls, params):
...
-    Editor for CurveShift
+    Editor for LinearTransform
     """
     # TODO: the layout changes when I click the area of the preprocessor
     #       EFFECT: the sidebar snaps in
 
-    name = "Shift Spectra"
-    qualname = "orangecontrib.infrared.curveshift"
+    name = "Linear Transformation"
+    qualname = "orangecontrib.infrared.lineartransform"
```
stuart-cls, line 306: Since we are renaming anyway, this is a good time to drop the legacy `infrared` namespace: use `orangecontrib.spectroscopy.lineartransform` instead.

stuart-cls (dismissed):
Thanks! Looks great. 

Since you renamed the preprocessor you will need to do a settings migration, see https://github.com/Quasars/orange-spectroscopy/blob/9d7856ceacae27e028ff72b738ce9fae0ba2a5af/orangecontrib/spectroscopy/widgets/owpreprocess.py#L920-L923 for an example.

orangecontrib/spectroscopy/widgets/preprocessors/misc.py
matyson: okay, i think i resolved these changes in the next 2 commits

stuart-cls:
Thanks! Backwards compatible on my machine now and tests past locally, just trying to get a sense of what's going on with the CI tests.

orangecontrib/spectroscopy/tests/test_preprocess.py
```diff
@@ -588,14 +588,14 @@ def test_iris(self):
                                         [4.75015528, 3.15366444, 1.46254138, 0.23693223]])
 
 
-class TestCurveShift(unittest.TestCase):
+class TestLinearTransform(unittest.TestCase):
 
     def test_simple(self):
         data = Table.from_numpy(None, [[1.0, 2.0, 3.0, 4.0]])
-        f = CurveShift(amount=1.1)
+        f = LinearTransform(amount=1.1, scale=2.)
```
markotoplak, line 595: Here we'd need better names for "amount". When there was just a shift, that was obvious.

markotoplak:
This seems a nice addition. I am not sure about the preprocessor name in the widget, though. 

So, in the widget, we'd have to present the preprocessor to the user with a name they would expect. Maybe "linear transformation" is too technical. Also, to me, "linear transformation" would imply something linear in the wavenumber range, like we'd achieve with EMSC with 2 parameters. To me this seems more like "stretch and move". What do other tools use? 

I am not against "linear transformation"  if that is something expected, but in this case, I'd keep the old one, too. Adds almost zero code because we can make a subclass. Also, "Shift spectra" is a bad name because it does not imply direction. :)

And the award for the worst name we already have would go to "Savitzy golay filter". I think we'd need to add a "Derivatives" preprocessor that opens (almost) exactly the same interface.

What do you think, @borondics and @stuart-cls?

markotoplak:
@borondics, @stuart-cls, could you decide on how to show this to the user? As I wrote above, I would propose having both Shift and Linear transformations because it makes sense for discoverability. Also, let's decide on naming classes and arguments so that we can finally merge this one.

raul-freitas:
Hello guys, it is great that we are ending this PR. As far as I understood the last issue is a semantic one. I think we can speed this up as for me the most important part is done. So, let us decide these names. As a suggestion and keeping in mind a general and consistent approach, I would suggest "Vertical Offset" for the existent one (taking the opportunity that Marko is already unhappy with the current name) and "Vertical Scaling" for the new one. The choice is in phase with Marko's comments about the direction.

Please let us know what you want to do next.

markotoplak:
We discussed this with @borondics. We decided that replacement with a single preprocessor is natural if we keep the name simple. We decided on the following name:
"Linear Transformation" -> "Shift and scale"

We decided on the following naming:
- the argument `amount` -> `offset`
- "Shift Amount" -> "Vertical offset" within the settings
- "Scale Factor" -> "Vertical scaling" within the settings

Any complaints? :)

raul-freitas:
> We discussed this with @borondics. We decided that replacement with a single preprocessor is natural if we keep the name simple. We decided on the following name: "Linear Transformation" -> "Shift and scale"
> 
> We decided on the following naming:
> 
> * the argument `amount` -> `offset`
> * "Shift Amount" -> "Vertical offset" within the settings
> * "Scale Factor" -> "Vertical scaling" within the settings
> 
> Any complaints? :)

Sounds good to me!

borondics:
@raul-freitas great! @matyson, could you please make the changes and then we merge?

matyson:
> We discussed this with @borondics. We decided that replacement with a single preprocessor is natural if we keep the name simple. We decided on the following name: "Linear Transformation" -> "Shift and scale"
> 
> We decided on the following naming:
> 
> * the argument `amount` -> `offset`
> * "Shift Amount" -> "Vertical offset" within the settings
> * "Scale Factor" -> "Vertical scaling" within the settings
> 
> Any complaints? :)

@markotoplak i think i've made all the suggested changes, could you review it?

PR #727, opened 2024-07-16
Add Phase unwrap preprocessor

matyson:
Add a preprocessor to unwrap the phase of a spectra using numpy's `unwrap` function. I did not include the proper widget implementation since it was implied from our last meeting that this widget better fits in the [Orange SNOM](https://github.com/Quasars/orange-snom) add-on. 

When looking this other repo, though, i've noticed that in order to extend the Preprocessor widget from the SNOM add-on i should first add this processing function at the Spectroscopy repo and then import it there, but if i'm wrong please let me know how to proper do this PR.

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -758,6 +758,40 @@ class CurveShift(Preprocess):
         return data.from_table(domain, data)
 
 
+class _PhaseUnwrapCommon(CommonDomain):
+
+    def __init__(self, unwrap, domain):
+        super().__init__(domain)
+        self.unwrap = unwrap
+
+    def transformed(self, data):
+        return np.unwrap(data.X)
+
+
+class PhaseUnwrap(Preprocess):
+    """
+    Unwrap phase values using numpy.unwrap defaults or bypass the data.
+
+    Parameters
+    ----------
+    unwrap    : toggle unwrap/bypass (boolean)
+    """
+
+    def __init__(self, unwrap=True):
+        self.unwrap = unwrap
+
+    def __call__(self, data):
+        if self.unwrap:
+            common = _PhaseUnwrapCommon(self.unwrap, data.domain)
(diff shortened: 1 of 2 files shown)


markotoplak:
@matyson, I just merged Quasars/orange-snom/pull/3, which shows how to write preprocessors in the SNOM add-on. There, [orangecontrib/snom/widgets/preprocessors/registration_example.py](https://github.com/Quasars/orange-snom/pull/3/files#diff-08c57d019c4df5af82482cec70e6d956d2d86e4d056575addcbee859de4ac4b1) indeed imports a preprocessor from `spectroscopy`, but that was done only for demonstration. It does not matter where the "computation" code is - we could have also defined it in the SNOM addon.

So I think this preprocessor code fits the SNOM add-on better too. I suggest you defined it in file `orangecontrib/snom/preprocess/phase_unwrap.py` or in something a bit more general if you plan to create some more related preprocessors too.

orangecontrib/spectroscopy/preprocess/__init__.py
```diff
@@ -758,6 +758,40 @@ def __call__(self, data):
...
+class PhaseUnwrap(Preprocess):
+    """
+    Unwrap phase values using numpy.unwrap defaults or bypass the data.
+
+    Parameters
+    ----------
+    unwrap    : toggle unwrap/bypass (boolean)
+    """
+
+    def __init__(self, unwrap=True):
```
markotoplak, line 780: Just a curiosity: why does this preprocessor has a bypass switch?

If this for testing or do you have a real usecase?

matyson: Actually, i don't quite remember why, i think i was inspired by another widget.

matyson:
@markotoplak, i will this PR to the SNOM add-on and follow your suggestions. 

For the tests, should i create a `orangecontrib/snom/tests` directory too? There is one `orangecontrib/snom/widgets/tests` but i am not sure if it applies to computation testing as well.

markotoplak:
Yes, please, create the appropriate test folder.

Issue #446, opened 2020-06-01
Submenus in Preprocess Spectra


borondics:
More and more method-specific preprocessors will be coming to Orange Spectroscopy and therefore I think it would be good to have the possibility of adding submenus for those to the list of preprocessors in order to keep it structured and easy to read.

borondics:
We would need this very much I think.

Requests for different preprocessors are coming in, for example new baseline correction methods, and simply extending the currently existing preprocessors is not always feasible. 

Even now, we are simply disabling some of the controls for the Rubberband and Linear baseline preprocessors because they only relevant for one of them. Ideally these should be two different GUIs entirely so we either solve to build the preprocessor GUI on the fly depending on which one is selected or make a submenu in the main preprocessor selector for baselines and then we can add new methods easier.

I don' think I can implement this especially for the reason of backward compatibility.

@stuart-cls, @markotoplak thoughts?

stuart-cls:
I agree that the baseline preprocessors should be updated and brought up to our current standards.

It's not too hard to add some dynamic GUI to a preprocessor "Editor" panel: the old Integrate interface does it for example. Taking Normalize as an example: if you think there are too many radio buttons (I guess that is the motivation?) we could change to a combobox and then any required subsettings can be popped in below without too much trouble.

Or maybe what I describe is not what you had in mind :)

PR #728, opened 2024-07-19
[ENH] Preprocess Spectra: supports categories of preprocessors

markotoplak:
Implements #446 with simple submenus.

Each editor can belong to multiple categories.


markotoplak:
We discussed this with @borondics today, and had parts of code ready from last year. It would be prudent if someone else tested this too (also, this could create problems in Integrate and Peak Fit).

stuart-cls:
I tried it, it seems to work. It doesn't impact Integrate or Peak Fit from what I can see, since it gracefully handles "preprocessors" items without the categories attribute.

From a UI perspective, I don't love it. It takes an already sub-optimal pop-up menu and adds nested menus, which are a real pain on some platforms (i.e. the little menus in Spectra/Hyperspectra). But that would be a much bigger UI change and also I don't have a better suggestion :D

PR #730, opened 2024-07-19
[ENH] File reader for 4D mapping stacks from the SGM beamline at CLS

stuart-cls:
Loads 4D stacks from the SGM beamline at the CLS. Primary energy axis (features) is the XRF energy (MCA).

Notes:
 * No test file, will try to get one
 * These are Nexus-compatible files, so future work is to make a more generic nexus reader and subclass that 
 * Missing piece is tool to rotate primary energy axis after ROI integration, I'll do that as a separate PR


markotoplak:
Merging despite the missing test files. Please add them when you'll get a chance.

PR #738, opened 2024-08-19
[MNT] Refactor preprocess tests

markotoplak:
Testing common preprocessor functionality - also lovingly called torture tests - relied on putting preprocessors into a special list in `test_preprocess.py`. This prevented reusing tests in other repositories.

An example of a PR that could benefit them: Quasars/orange-snom/pull/4

This PR refactors preprocessor tests so that torture tests can now be mixed into the test class.


markotoplak:
So this helped me find a bug in the other repo even though I honestly though that nothing can be wrong there... Tests are useful.

markotoplak:
This PR should not break anything (it is only tests). I'll review the code again and then merge it.

PR #739, opened 2024-09-06
[ENH] Neaspec multichannel reader

brnovasco:
This PR is related to the discussion in https://github.com/Quasars/orange-spectroscopy/pull/717.

Following up on the development of a reader for the Neaspec data containing multiple channels with separated data for phase and amplitude in multiple runs, rows and columns. 

We are creating a reader that performs a interpolation of the data to be able to have all interferograms in the same domain related to the raw position sensor reading instead of its array index. 

This imply changes in the OWFFT widget so we avoid duplicate code for the complex fft calculation. We then standardize the way the widget performs the fft for the .gsf and .txt data from Neaspec simplifying the logic on the FFT widget and improving the NeaReaderGSF class.


stuart-cls:
Thanks!
This seems quite clear, @raul-freitas explained to me about the interpolation part but I won't comment on that part.

> This imply changes in the OWFFT widget so we avoid duplicate code for the complex fft calculation. We then standardize the way the widget performs the fft for the .gsf and .txt data from Neaspec simplifying the logic on the FFT widget and improving the NeaReaderGSF class.

This is very nice, thank you.

From the FFT side, this PR seems good. Perhaps @markotoplak can put a second set of eyes on the reader side.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -310,4 +332,298 @@ def handle_data(self, data):
...
+        self.resampled_df = self._create_resampled_polar_df_from_cartesian(
+            self.cartesian_df,
+            self.cartesian_channel_names,
+            self.domain,
+            padding=padding,
+        )
+        # clean up the cartesian DataFrame
+        self.cartesian_df = pd.DataFrame()
+
+    def read(self):
```
stuart-cls, line 583: Are you re-implementing `build_spec_table` (and not subclassing `SpectralFileFormat`) for this reader just to change the number of decimals in the feature names? Just curious.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -310,4 +332,298 @@ def handle_data(self, data):
...
+        out_data = df.drop(columns=["Row", "Column", "Run", "Channel"]).values
+        out_data = out_data.astype(np.float64)
+        # formatting metas
+        meta_domain = [
+            Orange.data.ContinuousVariable.make("column"),
+            Orange.data.ContinuousVariable.make("row"),
+            Orange.data.ContinuousVariable.make("run"),
+            Orange.data.StringVariable.make("channel"),
+        ]
+        out_meta = df[["Column", "Row", "Run", "Channel"]].values
```
stuart-cls, line 617: Are "Column" and "Row" the XY map positions or something else? Generally we use `map_x` and `map_y` but it will work anyway.

borondics: we should start using the same x/y notation for position names, so i suggest renaming these to map_x and map_y

I made the same mistake in the "Reshape" widget, but I will change that to conform this nomenclature too

orangecontrib/spectroscopy/io/neaspec.py
brnovasco: Hi, @stuart-cls ! Thanks for reviewing. 

Before re-implementing build_spec_table, I bumped on an error related to finding the string representation of some value, which seemed connected to domain creation. While I wasn’t completely sure, I agree that it might be related to the number of decimals. However, I couldn’t see a straightforward way to resolve this using build_spec_table.

The other 'Nea' readers use the array index as the data domain, but in this file, we have actual positions, as @raul-freitas mentioned. The scale of these positions could indeed be causing the issue. I considered changing the scale (e.g., from meters to micrometers), but this would obscure that information from the user, even if we added it to the metadata. Do you think adjusting the scale would be a better approach?

orangecontrib/spectroscopy/io/neaspec.py
borondics: Thanks in advance! There were no conventions until now, but I think it is a good time to introduce them... @markotoplak and @stuart-cls, what do you think?

orangecontrib/spectroscopy/io/neaspec.py
stuart-cls: Regarding the scale, I don't know what SNOM users expect but I imagine that micrometers would be more readable. Certainly in the OPUS/Agilent readers we use micrometers for map coordinates.

orangecontrib/spectroscopy/datasets/NeaReaderMultichannel_test/Test_Au_Fourier_Scan_Synchrotron.txt
markotoplak, line 1: I have only one complaint: the test file (12MB) included is too large and will increase the repository size substantially.

It should be made much smaller, but so that it still tests most of the reader functionality. Perhaps limit the number of columns and runs to ~3.

The commit that changes the test file should then be rebased into the commit when the file was added so that we do not increase the repository history size. I can do that part too if you make a commit with a small test file.

markotoplak (changes requested):
I only have complaints about the test file size and this is the only thing I will enforce.

I do not deal with such files ofter therefore I am not the one to judge this from usability point of view. Also, someone else has to decide what to do with many Nea readers we know have - are there any to remove or merge together?

markotoplak:
@borondics, @stuart-cls, @brnovasco, please tell me when this is ready - I am not a good judge of file formats. If I understood the above comments, row and column should be changed to map_x and map_y, and then this is it?

Please do not merge yourselves because the test file should be made smaller and the big one rebased out.

borondics:
My only wish here is to have "map_x" and "map_y" as meta names.

Maybe we could make a definition file that all orange-spectroscopy can import and define these there?

In some other readers the final step is a helper function that constructs the data table from the different parts and that has "map_x" and "map_y". So, we could unify things that way too.

Also, in this particular case it should be row -> map_x and column -> map_y because the measurement method is row-by-row, which introduces special need for the data analysis.

As for test files, it should be really easy to just delete all the data except for a few lines, then it is a really dramatic space saving.

orangecontrib/spectroscopy/io/neaspec.py
stuart-cls: I updated that PR to solve just this issue.

stuart-cls:
Here's my understanding:

- [ ] Test file too big
- [ ] standard map vars would be nice
- [ ] `.txt` extension is overloaded
- [ ] decide on meters vs microns

orangecontrib/spectroscopy/io/neaspec.py
raul-freitas: Hello @brnovasco and @stuart-cls, thanks for advancing this. About the scale, microns is an appropriate unit for the M column. This column has the position readings of the stage encoder in microns.

brnovasco: I believe the issue with scaling the domain from meters to micrometers is related to the formatting of the values used as column names in the table. The problem could arise from rounding, leading to multiple columns with the same name, and also from generating excessively long strings for column names.

raul-freitas: > Hello @brnovasco and @stuart-cls, thanks for advancing this. About the scale, METER is an appropriate unit for the M column. This column has the position readings of the stage encoder in meters. For example, for a typical mirror travel of 500 µm and 1024 pts, the step is ~0.49 µm. The M column will show 4.9E-7, therefore, meters.

Guys, just to make it clear, meters are the original unit of the data. However, converting it to microns would be more comfortable.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -310,4 +332,298 @@ def handle_data(self, data):
 
     def _gsf_reader(self, path):
         X, _, _ = reader_gsf(path)
-        return np.asarray(X)
\ No newline at end of file
+        return np.asarray(X)
+
+
+class NeaReaderMultiChannel(FileFormat):
+    EXTENSIONS = (".txt",)
```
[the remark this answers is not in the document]
brnovasco: Is this a matter of setting PRIORITY on the new widget with a value greater than the value from NeaReader?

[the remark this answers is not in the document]
stuart-cls: I think based on @raul-freitas comment in #747 the correct answer is to merge the two readers (or make a meta-reader). Let's not make this PR any busier. It works fine, users just have to select the correct reader at the moment.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -604,26 +588,29 @@ def read_spectra(self):
...
-        self.original_df = []
+        # format data
         out_data = df.drop(columns=["Row", "Column", "Run", "Channel"]).values
         out_data = out_data.astype(np.float64)
         # formatting metas
         meta_domain = [
-            Orange.data.ContinuousVariable.make("column"),
-            Orange.data.ContinuousVariable.make("row"),
+            Orange.data.ContinuousVariable.make("map_x"),
+            Orange.data.ContinuousVariable.make("map_y"),
```
stuart-cls, line 597: When you rebase to get rid of the large file commit, please use `from orangecontrib.spectroscopy.utils import MAP_X_VAR, MAP_Y_VAR`  from #654

brnovasco:
Hi, @markotoplak and @stuart-cls. I think its ready with the requested changes. Can you check?

stuart-cls (approved):
Looks good to me. Thank you for this nice PR.

markotoplak (approved):
Thank you for the PR. Also thank you for rebasing the big file out.

PR #741, opened 2024-09-16
[MNT] Support for Preprocess images from the SNOM addon

markotoplak:
A more general preprocessor widget that allows other kinds of previews.

This is support work for Quasars/orange-snom/pull/7


markotoplak:
This should not :) break anything, so merging so that testing the other PR is easier.

PR #750, opened 2024-09-27
[RFC][WIP] Auto map vars

stuart-cls:
Continuing from #654 (I'll rebase when ready)

Users get very confused (and I find it annoying) when Hyperspectra plots vs (clusters, map_x) or something instead of just (map_x, map_y) when they are present.

I added a XYDomainModel which reliably plots XY data in Hyperspectra if both these vars are present (otherwise reverts to existing behaviour).

If this works, I'll also apply to the other mapping-aware widgets (Bin, Stack, Polar, etc)

### DIFF
--- a/orangecontrib/spectroscopy/io/agilent.py
+++ b/orangecontrib/spectroscopy/io/agilent.py
@@ -9,6 +9,7 @@ from agilent_format import agilentImage, agilentImageIFG, agilentMosaic, agilent
     agilentMosaicTiles
 from orangecontrib.spectroscopy.io.util import SpectralFileFormat, _spectra_from_image, \
     TileFileFormat, ConstantBytesVisibleImage
+from orangecontrib.spectroscopy.utils import MAP_X_VAR, MAP_Y_VAR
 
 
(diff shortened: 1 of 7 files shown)


stuart-cls:
@markotoplak I saw your comment in https://github.com/Quasars/orange-spectroscopy/pull/765#issuecomment-2500882810 and wondered if my approach here with a custom `XYDomainModel` would also suffer from the settings bug you reference?

markotoplak:
Yes, it could IF that model is connected to something that is changing a context setting AND that context setting is not limited in the same way as the model.

stuart-cls:
I see some tests are failing now :D I'll dig more

stuart-cls:
Indeed this triggers https://github.com/biolab/orange3/issues/6721 in `TestOWHyper.test_few_samples`
It also hits a new bug since I wrote it due to ae1df1eadd7a34155fb5bf1ef7cf78296d24795e since `PyListModel.Seperator` doesn't have a `.name` attribute.
And it's a bit weird that the map_x/map_y variables end up duplicated. 

So I'm starting to think it's not worth the complexity, especially now that the attr_x/y combo boxes aren't hidden.

PR #751, opened 2024-10-01
[ENH] Single NeaGSF reader with measurement.signaltype attribute

ngergihun:
A new gsf reader that we discussed. Only reading the selected file using `reader_gsf` from utils and adding datatype to the attributes.

### DIFF
--- a/orangecontrib/spectroscopy/io/neaspec.py
+++ b/orangecontrib/spectroscopy/io/neaspec.py
@@ -361,7 +361,7 @@ class NeaImageGSF(FileFormat, SpectralFileFormat):
         X, XRr, YRr = reader_gsf(self.filename)
         features, final_data, meta_data = _spectra_from_image(X, np.array([1]), XRr, YRr)
 
-        meta_data.attributes["signal_type"] = signal_type
+        meta_data.attributes["measurement.signaltype"] = signal_type
 
         return features, final_data, meta_data
 
--- a/orangecontrib/spectroscopy/io/neaspec.py
+++ b/orangecontrib/spectroscopy/io/neaspec.py
@@ -8,7 +8,7 @@ from Orange.data import FileFormat, Table
 from scipy.interpolate import interp1d
 
 from orangecontrib.spectroscopy.io.gsf import reader_gsf
-from orangecontrib.spectroscopy.io.util import SpectralFileFormat
+from orangecontrib.spectroscopy.io.util import SpectralFileFormat, _spectra_from_image
 from orangecontrib.spectroscopy.utils import MAP_X_VAR, MAP_Y_VAR
 
 class NeaReader(FileFormat, SpectralFileFormat):
@@ -226,6 +226,7 @@ class NeaReaderGSF(FileFormat, SpectralFileFormat):
             file_html = folder_file + ".html"
 
         data_gsf_a = self._gsf_reader(file_gsf_a)
+        print(np.shape(data_gsf_a))
         data_gsf_p = self._gsf_reader(file_gsf_p)
         info = self._html_reader(file_html)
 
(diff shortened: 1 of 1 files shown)


markotoplak:
Why another reader for gsf? What does this do that the older does not? Why did you not just add the datatype to the old one?

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -310,4 +313,32 @@ def handle_data(self, data):
...
+            datatype = "Amplitude"
+        elif 'Z' in channel_name:
+            datatype = "Topography"
+        else:
+            datatype = "None"
+
+        X, XRr, YRr = reader_gsf(self.filename)
+        features, final_data, meta_data = _spectra_from_image(X, np.array([1]), XRr, YRr)
+
+        meta_data.attributes["datatype"] = datatype
```
markotoplak, line 341: datatype is too general. We need a better name for this.

ngergihun:
@markotoplak Yes, we agree, we have too many readers that do almost the same. We definitely want to eliminate and merge them. For the other reader, namely `NeaReaderGSF` requires that the user always have 3 specific files. We wanted to eliminate this by only reading one file. But I did not want the delete the other one. I also did not want to modify the `GSFReader` from `gsf.py` since we have a very specific pattern that we search for in the filename to auto-recognize the `datatype`.

markotoplak:
OK, I understand if this makes the workshop easier.

Please think of a better name and rebase it too.

ngergihun:
@markotoplak Sure, I'll do it, thanks!

ngergihun:
@markotoplak I rebased it to the new master and changed the name to mark that this reader is for images. Maybe it is still not the best name. We are going to come up with a unified nomenclature for the neaspec things after discussion with the others.

orangecontrib/spectroscopy/io/neaspec.py
borondics: I think we should keep it the meta.attribute name as "datatype" or at least I can't come up with a better name for now. The purpose is to be general because in the future we should try to add and use this kind of information with the data. As an example, IR and Raman spectra have some specifics and we could use this information elsewhere, like automatically setting the axes labels on the plots. 

In this particular case I was thinking to change it to "neadatatype" but this would not allow us to use this particular key with non-neaspec instruments. Then I thought "snomdatatype" but neaspec instruments can measure other stuff, not only snom... So I think a general name is the best here and gives us the flexibility for the future.

@markotoplak, what do you think with these arguments is it OK to merge?

markotoplak: I was also not able to work out a better name. But do we at least have a replacement for "data", that could imply that we have amplitude and phase in question? I can't think of a better word. Maybe `signal`? No idea.

Or maybe some prefix, like `spectra.datatype` or something better?

borondics: **I think "signal_type" is a good idea!**

The main point is to expand beyond the amplitude/phase box so that we can use the same meta.attribute later for other things like Raman or Mirage (which started to have many types of data coming out from it).

Then we could add later another meta.attribute for "method" that would distinguish between instrumental differences such as Raman vs. IR vs. SNOM but this is a discussion for later.

ngergihun: Ok, then I change this to `signal_type` and take care that image preprocessor search for the new name.

markotoplak: @ngergihun, wait. :) I would also prefer some prefix, like `spectroscopy.` or something smarter, so like `spectroscopy.signal_type` or maybe `measurement.signal_type`. @borondics, @ngergihun, if you think something similar applies, good, otherwise, just `signal_type` is good enough.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -226,6 +228,7 @@ def read_spectra(self):
             file_html = folder_file + ".html"
 
         data_gsf_a = self._gsf_reader(file_gsf_a)
+        print(np.shape(data_gsf_a))
```
markotoplak, line 231: Remove the print, please.

orangecontrib/spectroscopy/io/neaspec.py
ngergihun: @markotoplak okay, let's use `measurement.signaltype` :D :D

PR #753, opened 2024-10-03
[ENH] HyperSpectra: show value traces of selected lines

markotoplak:
tr

### DIFF
--- a/orangecontrib/spectroscopy/widgets/line_geometry.py
+++ b/orangecontrib/spectroscopy/widgets/line_geometry.py
@@ -71,7 +71,7 @@ def intersect_curves(x, ys, q1, q2):
     :param x: x values of curves (they have to be sorted).
     :param ys: y values of multiple curves sharing x values.
     :param q1: point of the line segment (x, y)
-    :param q2: point of the line segnemt (x, y)
+    :param q2: point of the line segment (x, y)
     :return:
     """
 
--- a/orangecontrib/spectroscopy/widgets/owhyper.py
+++ b/orangecontrib/spectroscopy/widgets/owhyper.py
@@ -576,6 +576,15 @@ class ImageSelectionMixin:
         if menu:
             menu.addAction(select_polygon)
 
+        line = QAction(
+            "Trace line", self, triggered=self.plot.vb.set_mode_select,
+        )
+        line.setShortcuts([Qt.Key_L])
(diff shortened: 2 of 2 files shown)


ngergihun:
WOW, so cool, it works for me nicely!

stuart-cls:
I had to try it, it's a popular request :)

Using Hair from Datasets, then "From spectra: Integral from 0" (basically the default Hyperspectra settings) it crashes after I draw a line:

```
  File "/home/reads/src/orange-spectroscopy/venv/lib/python3.11/site-packages/Orange/widgets/utils/concurrent.py", line 549, in _on_task_done
    self.on_done(future.result())
  File "/home/reads/src/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owspectra.py", line 499, in on_done
    master.set_curve_pens()
  File "/home/reads/src/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owspectra.py", line 1504, in set_curve_pens
    self.set_curve_pen(i)
  File "/home/reads/src/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owspectra.py", line 1497, in set_curve_pen
    self.curves_cont.objs[idc].setZValue(int(insubset) + 2*int(inselected))
                                                           ^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```
with local variables
```
OrderedDict([('color_var', None),
             ('have_subset', False),
             ('idc', 0),
             ('idcdata', 0),
             ('inselected', None),
             ('insubset', False),
             ('self',
              <orangecontrib.spectroscopy.widgets.owhyper.CurvePlotHyper object at 0x7f934959ae60>),
             ('thispen',
              defaultdict(<function CurvePlot.__init__.<locals>.<lambda> at 0x7f934956d4e0>,
                          {None: <PyQt5.QtGui.QPen object at 0x7f9349590350>})),
             ('value', None)])
```

The "Selected" output is correctly set.

markotoplak:
@stuart-cls, thanks, I'll look into this tomorrow. Just from the error I have no idea what is happening.

markotoplak:
@stuart-cls, thanks for finding the issue. I had problems reproducing it, but then finally figured out this code path happens when disabling "Show averages" in the trace plot. I only tested it with on...

The problem was the I initialized CurvePlot wrongly (see the commit for fix). :)

PR #754, opened 2024-10-04
HyperSpectra: change the default integration method

markotoplak:
This one also works for "short" spectra, like images with 1 column.

The problem is that if you open an image that has only one "spectral" column, you see nothing useful because the default integral is not working (=returns all the same values). I propose changing it.

I did not want to do any smart detection because then whatever would be detected would also be saved as a default setting. We could perhaps do detection if these settings were "schema only".

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owhyper.py
+++ b/orangecontrib/spectroscopy/tests/test_owhyper.py
@@ -22,7 +22,8 @@ from Orange.util import OrangeDeprecationWarning
 
 from orangecontrib.spectroscopy.data import _spectra_from_image, build_spec_table
 from orangecontrib.spectroscopy.io.util import VisibleImage
-from orangecontrib.spectroscopy.preprocess.integrate import IntegrateFeaturePeakSimple, Integrate
+from orangecontrib.spectroscopy.preprocess.integrate import IntegrateFeaturePeakSimple, \
+    Integrate, IntegrateFeatureSimple
 from orangecontrib.spectroscopy.widgets import owhyper
 from orangecontrib.spectroscopy.widgets.owhyper import \
     OWHyper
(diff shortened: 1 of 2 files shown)


borondics:
I was thinking about this a bit. The previous default, when we did the whole integral should also work for one column, no? If that is OK or we can make it work, we can keep the old default and later come up with a nicer clever solution that handles single column and multicolumn data as well.

markotoplak:
Integral of a function that defines one non-zero value on the real axis is zero, so we can not make that option work if we do not want to do some very artificial exceptions.

I would change the default, because the old one (integral on a the whole data width) rarely made much sense.

borondics:
What do we select then for hyperspectral datasets? Having the first value in the spectral domain makes also little sense... At least with the full integral we immediately saw a nice signal-to-noise image that actually says something about the measurement as a whole.

Can we make a hybrid version with logic (one col vs. multiple cols)? This can also come later, doesn't need to be now. Actually, many of the plotting variants don't work for one column so we should make them grey when there is only one column or remove them from the list? 

Ahh... life is never simple, is it? :)

markotoplak:
It is the same issue as with the color selector. If I do something smart, I break the settings unless I also modify them to schema_only or something.

But here I think that this a really nice simple default.

markotoplak:
Oh, in Hyperspectral you then get the middle wavelength.

stuart-cls:
> Oh, for hyperspectral data you then get the middle wavelength.

This is what Agilent's software does. It can be misleading about the hyperspectral cube as @borondics says. But it is fast (no integral calculation) which is nice for large datasets, I usually leave mine set to "closest value" (as a setting) when working with large datasets.

borondics:
Well, I am totally for speed in general! But then, let's have some fun and pick the wavelength randomly. :D What do you guys think?

markotoplak:
No, @borondics, we are not doing it random. :)

So, do I put this into the release with the scaling factors?

borondics:
Sure, let's have the middle point (which is also random, but the random number generator is the user, so I'm happy).

PR #755, opened 2024-10-04
Real space coordinates for GSF images

borondics:
Instead of simply having the number of pixels in the map we have added the calculations for the real coordinates. Data comes from the GSF header and was read previously.


markotoplak:
So two tests now fail because reader returns something different. I do agree that actual positions are much better than just indices, so, if you ensured that the values are correct; here don't trust your imagination, use some software for comparison, I'll fix the test so that they correspond to this.

borondics:
We checked the coordinates, they are good.

markotoplak:
Let's hope you checked well.

borondics:
Thanks!

PR #757, opened 2024-10-04
HyperSpectra: use a preselected colormap for some SNOM data

markotoplak:
So, @borondics, this is ALMOST what you wanted: two specific kinds of data get their own color maps.

So what is wrong with it? For those two data types users can not then change colors. If I allowed that and used the datatype as a hint in changing the actual settings, that setting would also (it is how settings work) influence the future HyperSpectra instances. I could make some more complex logic around it so that the setting would not be saved unless the user then manually changed it, but that is more complex and if not done carefully, could introduce more problems than it would solve.

This is it for for now. Take it or leave it...

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owhyper.py
+++ b/orangecontrib/spectroscopy/widgets/owhyper.py
@@ -439,8 +439,19 @@ class ImageColorSettingMixin:
                 self.img.setLookupTable(dat.colors)
                 return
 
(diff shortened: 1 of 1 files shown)


borondics:
Hi @markotoplak, thanks for trying to add this quickly. I think with this condition that currently it is not possible to change the colormap for those files we should just leave it in the original version, so no preselection of the colormap.

PR #758, opened 2024-10-04
Nea .gsf files not always include Interferometer Center/Distance

markotoplak:
#739 broke the existing Nea gsf reader. At least some of the files I see lack `"Interferometer Center/Distance"` and therefore it crashes.

### DIFF
--- a/orangecontrib/spectroscopy/io/neaspec.py
+++ b/orangecontrib/spectroscopy/io/neaspec.py
@@ -283,13 +283,18 @@ class NeaReaderGSF(FileFormat, SpectralFileFormat):
 
         # calculate datapoint spacing in cm for the fft widget
         number_of_points = px_z
-        scan_size = float(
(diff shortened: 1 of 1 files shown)


markotoplak:
@brnovasco, I see you added the part that was crashing from @borondics's file.

Why did you expect that this always appears in the files? It is in the specification? Or is it version dependant?

markotoplak:
Ok, I understand that this can happen, but we have to ensure that the common utilities run for all the variants.

brnovasco:
> @brnovasco, I see you added the part that was crashing from @borondics's file.
> 
> Why did you expect that this always appears in the files? It is in the specification? Or is it version dependant?

I was replicating the behavior that was written in the owfft widget, that assumes that if the reader is 'NeaReaderGSF' it expects it to have the 'Interferometer Center/Distance' field. Sorry I missed this error.

PR #759, opened 2024-10-04
NeaReaderMultiChannel - Apply wavenumber scaling to Δx

brnovasco:
Changes on the multichannel interferogram reader (NeaReaderMultiChannel) so it can check for wavenumber scaling factor metadata and apply it as a corresponding scaling factor to Δx (used by the fft). 

I think @raul-freitas can explain better how this is factor comes from the device generating the data, but in short it is calculated comparing the experimental data to reference values (i.e. water vapor lines) during the data collection. 

We discussed how to implement this in https://github.com/Quasars/orange-spectroscopy/pull/717 and now @raul-freitas validated the necessity of using this factor in the Δx calculated from the positional reference (M) we are using for this type of file.

### DIFF
--- a/orangecontrib/spectroscopy/io/neaspec.py
+++ b/orangecontrib/spectroscopy/io/neaspec.py
@@ -582,7 +582,20 @@ class NeaReaderMultiChannel(FileFormat, SpectralFileFormat):
                              M maximum values are outside the expected range."
             )
         self.domain = np.arange(d_start, d_end, dm)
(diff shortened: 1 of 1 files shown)


markotoplak:
This does not seem as if could break anything, so I am OK (even though I have no idea what it does :) ).

Thanks!

borondics:
It works on my mac and I could also build a workflow to convert the IFGs into a spectrum.

PR #764, opened 2024-10-23
[FIX] HyperSpectra: fix changing data with active line trace

markotoplak:
Solves the following error reported by @borondics. When you have the hyperspectra widget and make a line selection, everything is OK. However, if you load a different file into the same workflow, there is an error:

image


stuart-cls:
Fixes it for me.

PR #773, opened 2024-11-12
Speedup for the 2D COS widget

borondics:
50x speedup, numerically more stable and less than 1e-10 difference.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owcos.py
+++ b/orangecontrib/spectroscopy/widgets/owcos.py
@@ -41,11 +41,9 @@ def calc_cos(table1, table2):
     sync = series1.T @ series2 / (len(series1) - 1)
 
     # Hilbert-Noda transformation matrix
-    hn = np.zeros((len(series1), len(series1)))
-    for i in range(len(series1)):
-        for j in range(len(series1)):
-            if i != j:
-                hn[i, j] = 1 / np.pi / (j - i)
+    i, j = np.ogrid[:len(series1), :len(series1)]
+    with np.errstate(divide='ignore'):
+        hn2 = np.where(i != j, 1 / (np.pi * (j - i)), 0)
 
     # asynchronous correlation
     asyn = series1.T @ hn @ series2 / (len(series1) - 1)


borondics:
This is an example how to parallelize isocurves, but we figured out that the slow part is actually the plotting, so I am removing this.

orangecontrib/spectroscopy/widgets/owcos.py
```diff
@@ -41,11 +41,9 @@ def calc_cos(table1, table2):
     sync = series1.T @ series2 / (len(series1) - 1)
 
     # Hilbert-Noda transformation matrix
-    hn = np.zeros((len(series1), len(series1)))
-    for i in range(len(series1)):
-        for j in range(len(series1)):
-            if i != j:
-                hn[i, j] = 1 / np.pi / (j - i)
+    i, j = np.ogrid[:len(series1), :len(series1)]
+    with np.errstate(divide='ignore'):
```
markotoplak, line 45: Do not do these error state ignore because they are global (=can create problems if running multiple threads).

borondics: I thought it ignores it only for the calculation in the "with" statement...?

markotoplak: That would be correct if the assumption that only one thread is running holds. But this is not true in general...

borondics: So, then we just swallow the errors? Or is it possible to catch them in another way?

orangecontrib/spectroscopy/widgets/owcos.py
```diff
@@ -417,18 +417,26 @@ def plotCOS(self):
...
+                self.left_plot.addItem(pt, ignoreBounds=True)
 
             self.left_plot.plot(leftSP.mean(axis=0), leftSPwn, pen=p)
             self.left_plot.addItem(self.left_hLine)
 
-            for s in topSP:
-                self.top_plot.plot(topSPwn, s, pen=pg.mkPen(color=(50, 50, 50), width=0.5))
+            self.left_plot.setXRange(np.min(leftSP), np.max(leftSP))
+
+            top_indices = np.linspace(0, len(topSP)-1, 100, dtype=int)
```
markotoplak, line 430: skip indices if you have <100 spectra, now this would repeat them.

borondics: Good point, thanks!

orangecontrib/spectroscopy/widgets/owcos.py
markotoplak: Swallow. :(

borondics: Apparently starting from numpy 2.0 this would be [thread-safe](https://numpy.org/doc/stable/reference/generated/numpy.errstate.html)... Maybe when the numpy restriction in Orange is lifted we can revisit this. For now I'll remove it and make a note of this in the code.

orangecontrib/spectroscopy/widgets/owcos.py
```diff
@@ -442,18 +419,34 @@ def plotCOS(self):
             self.cbarCOS.set_range(-1 * np.nanmax(np.absolute(cosmat)), np.nanmax(np.absolute(cosmat)))
             self.cbarCOS.set_colors(np.array(colorcet.diverging_bwr_40_95_c42) * 255)
 
-            for s in leftSP:
-                self.left_plot.plot(s, leftSPwn, pen=pg.mkPen(color=(50, 50, 50), width=0.5))
+            if len(leftSP) > 100:
+                left_indices = np.linspace(0, len(leftSP)-1, 100, dtype=int)
```
markotoplak, line 423: Don't generate indices with linspace, it is complex (see the exception) and just not made for it. To me, having to error round to get indices, always smells. Rather use: `random.Random(0).sample(len(whatever), 100)` to generate indices which are only needed if you have too many spectra. There should be no else

orangecontrib/spectroscopy/widgets/owcos.py
```diff
@@ -15,7 +15,7 @@
 from orangewidget.utils.visual_settings_dlg import VisualSettingsDialog
 from orangecontrib.spectroscopy.data import getx
 from orangecontrib.spectroscopy.widgets.owhyper import ImageColorLegend
-from orangecontrib.spectroscopy.widgets.owspectra import InteractiveViewBox
```
markotoplak, line 18: Remove this unneeded import (found it in the lint)

orangecontrib/spectroscopy/widgets/owcos.py
borondics: I am using this because we should do an equidistant sampling of the spectra (usually these experiments describe something changing as a function of time/temperature/etc), random is not a good choice in this case...

markotoplak: Ok, I understand your argument. If being equidistant is important, then this is fine.

Random of 100 is almost equidistant though. :)

markotoplak:
I simplify the code a bit and intend to merge if tests pass. Thanks!

PR #774, opened 2024-11-13
Perkin Elmer file reader

borondics:
Based on existing code from [specio](https://github.com/paris-saclay-cds/specio) and from colleagues at the Diamond synchrotron comes a new reader for Perkin Elmer files.


orangecontrib/spectroscopy/io/__init__.py
```diff
@@ -23,3 +23,5 @@
 from .maxiv import HDRReader_STXM
 from .soleil import SelectColumnReader, HDF5Reader_HERMES, HDF5Reader_ROCK
 from .cls import HDF5Reader_SGM
+
+from .perkinelmer import PerkinElmerReader
```
stuart-cls, line 27: Let's put this in the `# Instrument-specific readers` section

orangecontrib/spectroscopy/io/perkinelmer.py
```diff
@@ -0,0 +1,471 @@
...
+from Orange.data import Table, Domain, FileFormat
+
+from orangecontrib.spectroscopy.io.util import SpectralFileFormat, _spectra_from_image_2d
+
+
+### This file is based on the development chain indicated below:
+
+## The below file readers are based on file readers written by Specio
+## (https://github.com/paris-saclay-cds/specio) with the following
+## license:
```
stuart-cls, line 15: I guess it's the `BlockReader` class that is borrowed from `specio`? Let's put that class in `orangecontrib.spectroscopy.utils.specio` with the other code-copies.

orangecontrib/spectroscopy/io/perkinelmer.py
```diff
@@ -0,0 +1,471 @@
...
+        domain = Domain([], None)
+        meta_data = Table.from_numpy(domain,
+                                     X=np.zeros((len(datavals), 0)))
+
+        meta_data.attributes = meta
+
+        return datavals, wavenumbers, meta_data
+
+    def read_spectra(self):
+        if self.filename[-2:] == "sp":
```
stuart-cls, line 453: We haven't done this before (2 readers in one) but I don't have a strong preference.

Better to use `os.path.splitext[1]` than string mangling I think.

orangecontrib/spectroscopy/io/perkinelmer.py
```diff
@@ -0,0 +1,471 @@
...
+            xpositions = m.attributes['x_init'] + m.attributes['x_delta'] * np.arange(m.attributes['n_x'])
+            ypositions = m.attributes['y_init'] + m.attributes['y_delta'] * np.arange(m.attributes['n_y'])
+
+            # TODO this is not nice - we need new helpers to build the coordinates
+            y_loc = np.repeat(np.arange(m.attributes['n_y']), m.attributes['n_x'])
+            x_loc = np.tile(np.arange(m.attributes['n_x']), m.attributes['n_y'])
+
+            features, final_data, locs_table = _spectra_from_image_2d(intensities, wn,
+                                                                      xpositions[x_loc],
+                                                                      ypositions[y_loc])
```
stuart-cls, line 469: If it is an image (not a list of arbitrary positions), then yes, you should use `_spectra_from_image` not `_spectra_from_image_2d`. It has the repeat/tile stuff already. 

Then you can just return the result of that function directly.

stuart-cls (changes requested):
Great!

You need to add the test files to `MANIFEST.in` before we can find out if the tests pass :)

Generally, I don't love the structure of the reader: 

1. It's two readers jammed into one for no reason I can see.
2. I don't really like the API boundary between the `specio` code and the reader code. Looking at `specio`, I can see why it was done this way. I would prefer to keep the code-copies in the `utils` folder and the Orange-specific stuff only in the reader section. 
 
I understand if you don't have much appetite for pedantic refactoring :) I don't think it will affect maintainability so skip it if you'd rather not.

orangecontrib/spectroscopy/io/perkinelmer.py
markotoplak: So, by joining two readers, you make a list of readers smaller, which does make sense. But what you lose is the ability to enforce a reader if a file perhaps has a wrong extension. If that is not a problem, go forward with the code, but even in this case I would only enforce one of the readers, so have `if ... else` instead of `if .. elif`.

One thing to fix. What i find problematic is that this string comparison respects letter case. You must do `.lower()` before comparing.it.

orangecontrib/spectroscopy/io/__init__.py
borondics: done

orangecontrib/spectroscopy/io/perkinelmer.py
borondics: Yeah, I think we need to unify readers as much as possible. See https://github.com/biolab/orange3/issues/6929 -  I don't know what is the best way for doing this but currently we are going towards producing lists that don't fit a screen anymore.

I did the changes for the logic but a bit differently... Let me know how you like it.

orangecontrib/spectroscopy/io/perkinelmer.py
borondics: I made a new file to store this part of the code. I think the guys from Diamond refactored the specio code a bit.

borondics:
Thanks for the comments guys, I tried to improve things! Have a look...

orangecontrib/spectroscopy/io/perkinelmer.py
stuart-cls: Yes, I saw that after I made this comment.

Great, can you put the `specio` file in `orangecontrib/spectroscopy/utils/specio`? That's where the other code copies are (pymca, etc)

orangecontrib/spectroscopy/io/perkinelmer.py
stuart-cls: I don't see how making the list shorter will help any use case (except the gui problem). Ideally the user never even looks at that list, auto-matching works. The only time they need the list is when they need to override the auto-matching and pick a specific reader, in which case they need the whole list!

orangecontrib/spectroscopy/io/perkinelmer.py
stuart-cls: From the code, there's no way the current reader can support arbitrary lists! It generates the coordinates from `x_init` / `x_delta` / 'n_x`. Are you sure `.fsm` files can have arbitrary lists?

If you make one helper, you then have to send it a variable to tell it what the data is. Better to have tools that do one thing well as we currently do. Maybe the names are slightly confusing, but they are well described in the docstrings.

orangecontrib/spectroscopy/io/perkinelmer.py
borondics: Yes, it is a GUI problem. Unfortunately the automatic recognition is not working all that well. Especially for general extensions (csv, txt, dat, etc. <-we should make this a filename too) and extensions where there can be multiple formats inside the file (NeaSPEC txt files). So I think unifying the readers by categories that make sense would be a good thing. Ideally I would like to see instrument or company names only once in the list and from there handle things well for each group. As we hopefully will support more and more file reading the list will become long in any case, but with this direction we can keep it more reasonable I think.

stuart-cls: Okay I think I understand. Some kind of middle ground, using more of the meta-reader infrastructure to shorten the list. And of course continuing to improve the automatic recognition!

If I was doing this one from scratch with your suggestion, I would do
PerkinElmerMeta -> PerkinElmerSP / PerkinElmerFSM but I think we've discussed the colour of this shed long enough :)

stuart-cls (approved):
This is good from my side, don't forget @markotoplak comment about line endings and squashing the commits.

borondics:
Then, it is good for me too... @markotoplak, would you merge it?

orangecontrib/spectroscopy/io/perkinelmer.py
borondics: Let's see what we can work out in the future. I'll open a new issue for the meta reader discussion.

markotoplak:
Tests are bad. Test should not skip complex parts of the code, like image coordinates, which are not tested at all. This should be improved.

markotoplak:
I am merging this as is, but tests should really be made better.

borondics:
Thanks for merging! I added map coordinate tests in #776

PR #776, opened 2024-11-20
Added tests for map coordinates.

borondics:


markotoplak:
I am fine with this precise checking if it passes. Otherwise, it this case, where they should be mostly rounded to the integer, assertAlmostEqual with the correct parameters would be preferred.

What I do not like is that these tests only test two pixels in the first line or column (do not know which is which). The more interesting cases are the ones somewhere in the middle of the data.

borondics:
This test file is a 4x4 map so we are actually testing the middle pixels.

markotoplak:
Yes, you are testing middle pixels on the first row/column. Please test middle pixes on middle row/column. :)

borondics:
How about a test like this? All the coordinates... :)

borondics:
Your solution is much nicer! 

However, while confusing, I think the int() or round() difference is not important as what you are testing is code stability, not real things, right? In any case, thanks for the improvement.

PR #777, opened 2024-11-25
[ENH] HyperSpectra: move image properties to control area

markotoplak:
HyperSpectra has too many options in the menu, and some of them are as important as the ones in the control area. I suggest to unify this.

image

I have been thinking about this for a while, especially when doing workshops and students do not understand, why do their images look so funny, when they only had strangely set axes. MY final motivation was #765, which adds additional options. Having less often used options in control panel and more common ones (especially axes) in the menu seemed off to me. Therefore, I tried this and to me it seems fine. 

Asking for comments, especially @stuart-cls and @borondics.

In #765 I also made visual image pane smaller if disabled.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owhyper.py
+++ b/orangecontrib/spectroscopy/widgets/owhyper.py
@@ -324,7 +324,8 @@ class ImageColorSettingMixin:
         box = gui.vBox(self)
         box.setContentsMargins(0, 0, 0, 5)
         self.color_cb = gui.comboBox(box, self, "palette_index", label="Color:",
-                                     labelWidth=50, orientation=Qt.Horizontal)
+                                     labelWidth=50, orientation=Qt.Horizontal,
+                                     contentsLength=10)
         self.color_cb.setIconSize(QSize(64, 16))
         palettes = _color_palettes
         model = color_palette_model(palettes, self.color_cb.iconSize())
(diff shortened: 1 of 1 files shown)


borondics:
Hi @markotoplak, I think you are right about this, the feature is too hidden in the menu, so this change is quite timely. If we go with this though I think we should be starting to enable collapsing those boxes because soon we will need to scroll the left panel, especially with smaller windows or monitors.

In the long term I would really like to have it working by interacting with the colorbar itself, not from a menu or panel. 
Here is an example of a pyqtgraph colorbar:
image

As I see now we have our own colorbar implementation - is this a must or we can default to pyqtgraph here?

I guess in any case adding double click and right click functionalities would not be too hard, although one could argue that these are also hidden features, but I think they would be also intuitive.

markotoplak:
I am open to nicer color choices in the future. The limit sliders are exactly what are you showing here are could be replaceable.

But do we go forward with this for now?

borondics:
I think we should go ahead with this change for now.

Would it be easy to add collapse/open functionality to the sections?

markotoplak:
It is probably easy to add, but I would not (no other widgets use something similar).

For now, we'll collapse open vector plots and visual images because they are rarely used.

If the UI is too cluttered, we could hide inactive settings in Image Values.

stuart-cls:
:100:  :100: :100: from me. 

The hidden axis thing has bitten so many users (and also me). It also matches other widgets like scatterplot etc better.

Thanks @markotoplak !

stuart-cls:
Ah, the "Image colors" box should probably disappear on RGB mode. It looks strange just empty:

image

stuart-cls:
Should we do this to **Spectral Series** as well? I think it's the only other widget with this type of view (at least, in this repo).

PR #779, opened 2024-12-09
[ENH] SoftiMax HDF5 reader

borondics:
This PR contains a change for the HERMES reader to ensure that the file is closed after reading and a new HDF5 reader for the Max-IV SoftiMax beamline that has almost the same entries as the HERMES (counter instead of Counter). Instead of doing some string manipulation magic I added a new class. Eventually it would be nice to unify these readers though.

Also, I will change the order of `x_locs` and `y_locs` in the HERMES reader because the current code will crash with non-square maps.

To come: small SoftiMax test file and reader test.

### DIFF
--- a/orangecontrib/spectroscopy/io/soleil.py
+++ b/orangecontrib/spectroscopy/io/soleil.py
@@ -43,16 +43,36 @@ class HDF5Reader_HERMES(FileFormat, SpectralFileFormat):
 
     def read_spectra(self):
         import h5py
-        hdf5_file = h5py.File(self.filename)
-        if 'entry1/collection/beamline' in hdf5_file and \
-                hdf5_file['entry1/collection/beamline'][()].astype('str') == 'Hermes':
-            x_locs = np.array(hdf5_file['entry1/Counter0/sample_x'])
-            y_locs = np.array(hdf5_file['entry1/Counter0/sample_y'])
-            energy = np.array(hdf5_file['entry1/Counter0/energy'])
-            intensities = np.array(hdf5_file['entry1/Counter0/data']).T
-            return _spectra_from_image(intensities, energy, x_locs, y_locs)
-        else:
-            raise IOError("Not an HDF5 HERMES file")
+        with h5py.File(self.filename, 'r') as hdf5_file:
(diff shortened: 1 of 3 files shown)


orangecontrib/spectroscopy/io/soleil.py
```diff
@@ -43,16 +43,36 @@ class HDF5Reader_HERMES(FileFormat, SpectralFileFormat):
...
+                y_locs = np.array(hdf5_file['entry1/Counter0/sample_y'])
+                energy = np.array(hdf5_file['entry1/Counter0/energy'])
+                intensities = np.array(hdf5_file['entry1/Counter0/data']).T
+                return _spectra_from_image(intensities, energy, x_locs, y_locs)
+            else:
+                raise IOError("Not an HDF5 HERMES file")
+
+
+class HDF5Reader_SoftiMAX(FileFormat, SpectralFileFormat):
+    """ A very case specific reader for HDF5 files from the SoftiMAX beamline in MAX-IV"""
```
stuart-cls, line 59: Can you put it in `maxiv.py instead?`

orangecontrib/spectroscopy/io/soleil.py
```diff
@@ -43,16 +43,36 @@ class HDF5Reader_HERMES(FileFormat, SpectralFileFormat):
...
+    EXTENSIONS = ('.hdf5',)
+    DESCRIPTION = 'HDF5 file @SoftiMAX/MAX-IV'
+
+    def read_spectra(self):
+        print("SoftiMax")
```
stuart-cls, line 64: Leftover print

stuart-cls (commented):
Looks good! I like this approach, better to get it in and combine later. Example file / tests would be good.

I would guess these are actually `NXstxm` files and we can use that logic in the future.

>Also, I will change the order of x_locs and y_locs in the HERMES reader because the current code will crash with non-square maps. 

I don't understand this comment.

markotoplak:
I'll merge this for now so that it is released, but tests would be appreciated.

borondics:
Thanks guys! I will write tests.

PR #785, opened 2025-01-08
[ENH] Re-enable equivalence of spectral tranformations

markotoplak:
Now that orange3/pull/6764 is released in Orange 3.38, we can revert #713.

So this brings #625 back.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,13 +126,13 @@ if __name__ == '__main__':
         data_files=DATA_FILES,
         install_requires=[
             'setuptools>=51.0.0',  # same as the last one needed for Orange3
-            'pip>=19.3',  # same as for Orange 3.37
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -126,13 +126,13 @@ def include_documentation(local_dir, install_dir):
         data_files=DATA_FILES,
         install_requires=[
             'setuptools>=51.0.0',  # same as the last one needed for Orange3
-            'pip>=19.3',  # same as for Orange 3.37
+            'pip>=19.3',  # same as for Orange 3.38
```
stuart-cls, line 129: Why do we have pip in the requirements?

stuart-cls (approved):
LGTM and the tests are passing :rocket:

setup.py
markotoplak: No idea. :)

PR #786, opened 2025-01-28
FFT widget: use WidgetPreview

borondics:
This might be a bad fix, maybe AnyQt should handle this, but it has implications for orange3 as well?

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owfft.py
+++ b/orangecontrib/spectroscopy/widgets/owfft.py
@@ -691,7 +691,7 @@ def main(argv=sys.argv):
 
     ow.set_data(dataset)
     ow.handleNewSignals()
-    app.exec_()
+    app.exec()
     ow.set_data(None)
     ow.handleNewSignals()
     return 0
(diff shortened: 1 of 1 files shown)


stuart-cls:
@borondics What problem does this solve? I'm asking because our tests are passing on PyQt6 already so we should capture this failure as well as fixing it.

borondics:
@stuart-cls with app.exec_() the widget doesn't start on it's own when testing it manually from Pycharm. Maybe it is a config problem on my computer or a Mac issue?

stuart-cls:
Ah, I see. I guess we don't test that. The problem is the FFT widget is using a really old way of firing up the test widget. I'll update this PR to the new way.

As an aside, somewhere along the way the default test file loaded was switched to `NeaReaderGSF_test/NeaReaderGSF_test O2A raw.gsf` from `IFG_single.dpt`. Is this still the best test file? Should we make it a script option?

borondics:
@markotoplak if you look for app.exec_() in the orange3 codebase you will find multiple locations where it is still used...

markotoplak:
@borondics, no worries, it will get adapted when someone will be running it directly.

PR #788, opened 2025-02-05
[ENH] PTIR loader: add images

kaiyang-tu:
Update to PTIR loader to search for Camera / Fluorescence images based on PTIR-studio v4.5 data format.

In addition it will pull focus information for each spectra and add it as meta information


markotoplak:
Thanks for taking care of the review, @stuart-cls. I would mainly like the try: except: to be shorter and have a tiny test file, if possible.

orangecontrib/spectroscopy/io/ptir.py
```diff
@@ -65,6 +67,42 @@ def read_spectra(self):
...
+        for img_name in image_channels:
+            if 'Image' in img_name:
+                img = hdf5_file['Images'][img_name]
+                im = Image.fromarray(img[:], 'RGBA')
+            elif 'Heightmap' in img_name:
+                img = hdf5_file['Heightmaps'][img_name]
+                im = Image.fromarray((img[:]/np.max(img[:]) - np.min(img[:])/np.max(img[:])) * 255, 'F')
+                im = im.convert('L')
+            else:
+                continue
```
stuart-cls, line 95: If we do a `warnings.warn(f"Unsupported visible image type {img_name}")`  future unsupported images types will get surfaced. Unless there's already some of these that you are intentionally hiding.

orangecontrib/spectroscopy/io/ptir.py
```diff
@@ -161,23 +201,42 @@ def read_spectra(self):
...
+            new_attributes.append(ContinuousVariable.make('z-focus'))
+            new_columns.append(np.full((len(z_locs),), z_locs))
+
+            domain = Domain(additional_table.domain.attributes,
+                            additional_table.domain.class_vars,
+                            additional_table.domain.metas + tuple(new_attributes))
+            data = additional_table.transform(domain)
+            with data.unlocked():
+                data[:, new_attributes] = np.asarray(new_columns).T
+            return features, spectra, data
```
stuart-cls, line 224: The `if hyperspectra` block will never include `visible_images` returning here. Instead, put the following `# locations` block under an `else:` and then have a common section after that which adds visible images and returns features, spectra, data

orangecontrib/spectroscopy/io/ptir.py
```diff
@@ -65,6 +67,42 @@ def read_spectra(self):
         wavenumbers = []
         x_locs = []
         y_locs = []
+        # for including focus information
+        z_locs = []
+
+        # map checked images in file
+        image_channels = []
+        if 'Heightmaps' in keys:
+            for img_name in [i for i in hdf5_file['Heightmaps'].keys()]:
```
stuart-cls, line 76:
```suggestion
            for img_name in list(hdf5_file['Heightmaps'].keys()):
```

orangecontrib/spectroscopy/io/ptir.py
```diff
@@ -65,6 +67,42 @@ def read_spectra(self):
...
+                if hdf5_file['Heightmaps'][img_name].attrs['Checked'][0] == 1:
+                    image_channels.append(img_name)
+        if 'Images' in keys:
+            for img_name in [i for i in hdf5_file['Images'].keys()]:
```
stuart-cls, line 80:
```suggestion
            for img_name in list(hdf5_file['Images'].keys()):
```

orangecontrib/spectroscopy/io/ptir.py
```diff
@@ -161,23 +201,42 @@ def read_spectra(self):
...
+            domain = Domain(additional_table.domain.attributes,
+                            additional_table.domain.class_vars,
+                            additional_table.domain.metas + tuple(new_attributes))
+            data = additional_table.transform(domain)
+            with data.unlocked():
+                data[:, new_attributes] = np.asarray(new_columns).T
+            return features, spectra, data
+
+        # locations
+        x_loc = y_loc = z_loc = np.arange(spectra.shape[0])
```
stuart-cls, line 227: These three variables (`x_loc`, `y_loc`, `z_loc`) are all the same index array, and only used once. Call it `i_loc` or something and use that

stuart-cls: Actually, unless `x_locs` and friends can be longer than `spectra` (I don't know, can they?) why not just replace both of these lines with 
`metas = np.vstack((x_locs, y_locs, z_locs)).T`

stuart-cls (commented):
Looks good @kaiyang-tu ! Just a few small changes. The rest of the lint stuff is coming from the original code, I'm not sure why it's triggering the lint diff.

orangecontrib/spectroscopy/io/ptir.py
kaiyang-tu: Thanks for mention here. It did cause one issue with it images in `hyperspectra` files. Changed to common `return features, spectra, data` at the end.

orangecontrib/spectroscopy/io/ptir.py
kaiyang-tu: Based on my understanding, images as either `'Heightmaps'` with `[x,y]` array or `'Image'` as `[x,y,4]` array.   
Maybe better way to account for unsupported image types is take the image conversion out of this `if else` and do a follow up with an `img.shape` check for what conversion or `warnings.warn(f"Unsupported visible image type {img_name}")` is needed?

stuart-cls: That sounds good. We can also just leave it since now I realize that the `image_channels` list is only populated from the `Heightmaps` and `Images` keys in the previous block so the `else:` block here is probably not reachable.

orangecontrib/spectroscopy/io/ptir.py
```diff
@@ -65,6 +67,42 @@ def read_spectra(self):
...
+        # map checked images in file
+        image_channels = []
+        if 'Heightmaps' in keys:
+            for img_name in [i for i in hdf5_file['Heightmaps'].keys()]:
+                if hdf5_file['Heightmaps'][img_name].attrs['Checked'][0] == 1:
+                    image_channels.append(img_name)
+        if 'Images' in keys:
+            for img_name in [i for i in hdf5_file['Images'].keys()]:
+                if hdf5_file['Images'][img_name].attrs['Checked'][0] == 1:
+                    image_channels.append(img_name)
```
kaiyang-tu, line 82: Ran into a case, where image in PTIR files does not have `'Checked'` attribute. To account for this issue:
```suggestion
                img_attrs = hdf5_file['Images'][img_name].attrs
                if img_attrs.keys().__contains__('Checked') and not img_attrs['Checked'][0]:
                    image_channels.append(img_name)
```
Or better to put into `try` to catch future cases?
```suggestion             
                try:
                    if hdf5_file['Images'][img_name].attrs['Checked'][0] == 1:
                        image_channels.append(img_name)
                except AttributeError:
                    warnings.warn(f"Unsupported visible image entry {img_name}")
```
Same would be needed earlier for `'Heightmaps'`

stuart-cls: The first is fine, but it should be `if "checked" in img_attrs.keys() and img_attrs['Checked'][0] == 1:`

(You changed the boolean logic from the original, was that intentional?)

Or maybe just `if hdf5_file['Images'][img_name].attrs.get('Checked', [None])[0] == 1:`

What is the "checked" parameter anyway?

kaiyang-tu: Ok thanks, will change accordingly.

Checked is parameter in the PTIR studio software that allows you to deselect spectra/images that should not be displayed. It can be useful to cut down on spectra/images you have used for finding the area of interest.

stuart-cls: Hmm, it seems strange to me in a file loader to respect gui selections from another software? Perhaps it would be more appropriate to load them all and then put the unchecked ones at the bottom of the list or something. But I'll defer to your experience on this one :)

stuart-cls: We discussed this offline and @kaiyang-tu preferred to leave it as is. Can be changed in future PR if needed.

stuart-cls:
@kaiyang-tu looks good to me now. Sorry I missed your updates. I can't tell which lint issues you added and which were already there but I think we got the worst ones.

@markotoplak if you have time to take a quick look that would be great otherwise I will merge this later this week.

stuart-cls (approved):

PR #792, opened 2025-02-25
[FIX] NeaImageGSF fix: works without channel in filename

ngergihun:
Previously the NeaImageGSF reader crashed if the filename did not contain the proper data channel string pattern. With the fix, instead of crashing, it reads the file and considers it a Topography type measurement.

### DIFF
--- a/orangecontrib/spectroscopy/io/neaspec.py
+++ b/orangecontrib/spectroscopy/io/neaspec.py
@@ -345,29 +345,33 @@ class NeaImageGSF(FileFormat, SpectralFileFormat):
     EXTENSIONS = (".gsf",)
     DESCRIPTION = 'NeaSPEC single image'
 
-    def read_spectra(self):
+    @staticmethod
+    def detect_signal_type(filename):
 
         channel_strings = ['M(.?)A', 'M(.?)P', 'O(.?)A', 'O(.?)P', 'Z C', 'Z raw']
         channel_name = None
 
         for pattern in channel_strings:
-            if re.search(pattern, self.filename) is not None:
-                channel_name = re.search(pattern, self.filename)[0]
+            if re.search(pattern, filename) is not None:
+                channel_name = re.search(pattern, filename)[0]
 
         if channel_name is None:
             signal_type = 'Topography'
+        elif 'P' in channel_name:
+            signal_type = "Phase"
+        elif 'A' in channel_name:
+            signal_type = "Amplitude"
         else:
-            if 'P' in channel_name:
-                signal_type = "Phase"
-            elif 'A' in channel_name:
-                signal_type = "Amplitude"
-            elif 'Z' in channel_name:
-                signal_type = "Topography"
-            else:
-                signal_type = "Topography"
+            signal_type = "Topography"
+
+        return signal_type
+
+    def read_spectra(self):
 
         X, XRr, YRr = reader_gsf(self.filename)
         features, final_data, meta_data = _spectra_from_image(X, np.array([1]), XRr, YRr)
+
+        signal_type = self.detect_signal_type(self.filename)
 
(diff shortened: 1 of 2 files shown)


orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -351,15 +351,20 @@ def read_spectra(self):
...
-        if 'P' in channel_name:
-            signal_type = "Phase"
-        elif 'A' in channel_name:
-            signal_type = "Amplitude"
-        elif 'Z' in channel_name:
-            signal_type = "Topography"
+            else:
+                channel_name = None
+
+        if type(channel_name)==str:
```
markotoplak, line 357: `if channel_name is not None` (preferred here), alternatively also `isinstance(channel_name, str)`. `type` and comparisons are discouraged.

markotoplak: I would actually also flatten this out, you can start wtih `if channel_name is None`, say it is topography, and then continue with elifs.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -351,15 +351,20 @@ def read_spectra(self):
...
                 channel_name = re.search(pattern, self.filename)[0]
-
-        if 'P' in channel_name:
-            signal_type = "Phase"
-        elif 'A' in channel_name:
-            signal_type = "Amplitude"
-        elif 'Z' in channel_name:
-            signal_type = "Topography"
+            else:
+                channel_name = None
```
markotoplak, line 355: So what does this mean? This sets channel_name to None if only the last pattern is not found. Is this the correct logic?

If it should be None if it was not found in any of the channel_strings, then `channel_name = None` should got in front of the for loop.

ngergihun: Yes, true it is the logic. Sure, if it is before the loop, it does not have to be set, just ones to None.

orangecontrib/spectroscopy/io/neaspec.py
ngergihun: Yes, makes sense, thx for the suggestion. Just a dumb question: Why are type comparisons not favored?

markotoplak: `type` does not handle subclasses, `isinstance` does.

It is also what I am used seeing in most codebases, and directly tells the reader that there is type comparison there without even needing to parse the expression.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -348,13 +348,15 @@ class NeaImageGSF(FileFormat, SpectralFileFormat):
...
         for pattern in channel_strings:
             if re.search(pattern, self.filename) is not None:
                 channel_name = re.search(pattern, self.filename)[0]
-            else:
-                channel_name = None
 
-        if type(channel_name)==str:
+        if channel_name is None:
+            signal_type = 'Topography'
+        else:
```
markotoplak, line 359: Thiis else: and then if could be just elif.

markotoplak:
Thanks for the fixes, now it seems simpler. But I see that the behavior of this functionality can be quite complex (even your first commit had an error). So I suggest having tests. For that:

1) Make this a separate function, perhaps `detect_signal_type(filename)`.
2) Test it well, with carefully made-up or real filenames.

The tests then define what the behavior should be (and will be when someone edits the code further).

If you like it, put a test on it

ngergihun:
Hi Marko! Sure, should the test run with the Reader tests? Or should the function go to io utils and the test to the corresponding tests?

ngergihun:
OK. In the meanwhile, I realized that we don't even have a test for this reader at all.

markotoplak:
Looks good, thanks!

PR #795, opened 2025-03-06
OWFileBrowser widget

ngergihun:
We wanted to have a quick and easy way to browse through lots of measurements quickly. To achieve we implemented a new widget that has a treeview-based file selection enabling much easier file browsing and loading. We think this it is a useful addition that makes the life of an end user easier and also enables the use of this widget in standalone apps.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owfilebrowser.py
+++ b/orangecontrib/spectroscopy/widgets/owfilebrowser.py
@@ -10,7 +10,6 @@ from AnyQt.QtWidgets import (
     QHBoxLayout,
     QVBoxLayout,
     QFileDialog,
-    QFileSystemModel,
     QLabel,
     QWidget,
     QLineEdit,
@@ -19,6 +18,14 @@ from AnyQt.QtWidgets import (
     QHeaderView,
     QSizePolicy as Policy,
 )
+
+# QFileSystemModel model moved to QtGui in PyQt6;
+# use the following until AnyQt unifies it
+try:
+    from AnyQt.QtGui import QFileSystemModel
+except ImportError:
+    from AnyQt.QtWidgets import QFileSystemModel
+
 from AnyQt.QtCore import QSize, QDir, QPoint, Qt, Signal, QSortFilterProxyModel
 
 from orangewidget.utils.filedialogs import format_filter


borondics:
Thanks for this! I think it will be nice.

ngergihun:
No, but maybe the sheet could be saved. Is it with: `sheet = Setting(None, schema_only=True)` ?

markotoplak:
The last loaded file path should also be saved. Ideally, a RecentPath object is used for that setting. To keep the widget the as similar as OWFile, a list of RecentPath's with just one element could be used.

RecentPath already includes the reader and sheet setting.

ngergihun:
Okay, I'll make the changes to use RecentPath

markotoplak:
On my Linux with Qt 6.8.1 I get:

```
File "/home/marko/dev/orange-spectroscopy/orangecontrib/spectroscopy/widgets/owfilebrowser.py", line 8, in <module>
    from AnyQt.QtWidgets import (
ImportError: cannot import name 'QFileSystemModel' from 'AnyQt.QtWidgets' (/home/marko/venv312/lib/python3.12/site-packages/AnyQt/QtWidgets.py)
```

I added a commit that fixes the issue.

markotoplak:
General comments:
- File type chooser should go under the file pickler.
- Opening on click is a strong operation. Choosing a file should not make anything time consuming run by default. I'd add a checkbox where this can be enabled / disabled. By default it should be double-click.
- How does this work on Windows where there are also different drives (C:, D:)?
- Pathname display on Linux is wrong (I guess also on Macs). For me, it shows double slashes "/ /" for the start. (I do like how well does it interact though)

markotoplak:
Here I see many functions from OWFile copied and maybe slightly edited. This makes code review needlessly hard.

I therefore suggest this widget is redone as a OWFile subclass so that there is no need to copy the shared function. For now, at least. This will also bring in all recent_path functionality.

ngergihun:
@markotoplak, thx for the suggestions. I'll work on a new version and come back with that for review.

ngergihun:
Hi @markotoplak ! I started the rewrite of the widget by subclassing OWFile. I try to reduce the code as much as possible, but I still need to reimplement some functions a bit differently due to the different components and interactions with the widget.

Also, when subclassing I get a warning, saying:
`RuntimeWarning: subclassing of widget classes is deprecated and will be disabled in the future.
Extract code from OWFile or explicitly open it by adding `openclass=True` to class definition.
  class OWFileBrowser(OWFile):`

Also, OWFile has UI elements that we don't have and most of its functions refer to them, I have to create 'dummy' UI elements with the same name that's makes a bit of a mess.

If it is okay with the mentioned limitations, then I can continue.

stuart-cls:
I had the same problem with TileFile. I think in that widget I just ended up putting in a code-copy which is not great.

Really we need this: https://github.com/biolab/orange3/issues/5376

ngergihun:
@stuart-cls, yes, I came across your OWTileFile when I was looking for inspiration and suspected that you copy-modified some of the code for this reason.

markotoplak:
Please make it a subclass for now. For ease of the review, mainly. Even though Orange is warning against it.

PR #796, opened 2025-03-13
[FIX] PCA Denoising: set unused components to zero

### LINKED ISSUE #793: PCA Denoising never sets unused components to zero
Specifically https://github.com/Quasars/orange-spectroscopy/blob/12724370b2f7902f5b09cd215c36af06bc38bc5e/orangecontrib/spectroscopy/preprocess/__init__.py#L124

is never True.

The observed influence of changing the "components" setting in the widget is only due to varying the number of components in the PCA model.

stuart-cls:
Although this fixes #793, it has no mathematical impact on the output. 

I'm going to add the ability to select components but probably not expose that in the PreprocessorEditor


stuart-cls:
I was hoping to at least try caching the PCAModel but Preprocess widget throws the preprocessor away every time you change the components value so that doesn't gain anything at this point. Kind of a disappointing PR!

markotoplak:
> PreprocessorEditor. Currently the components are 1-based (rather than 0-based) to match the output of PCA widget (PCA1, PCA2, etc). Thoughts?

Obligatory EWD: https://www.cs.utexas.edu/~EWD/ewd08xx/EWD831.PDF

Let's keep the core code 0-based. The UI, if we allow it, can be 1-based.

stuart-cls:
I thought about how to explain it more clearly, here I go:

The previous code changed the number of components in the PCA model, and then always reconstructed from the full (but truncated model):

| PC1 | PC2 |
| --- | ---|
| 1.4 | 1.5 |
| 0.9 | 0.4 |

The new code always (mostly) calculates the same PCA model, and then truncates the number of components with non-zero eigenvalues.

| PC1 | PC2 | PC3 | ... | PCMAX |
| --- | ---| --- | --- | --- |
| 1.4 | 1.5 | ~0.5~ 0 | ... | ~1.34~ 0 |
| 0.9 | 0.4 | ~0.9~ 0 | ... | ~0.45~ 0 |

Since PCA always returns the same components in the same order, PC1 and PC2 are equal in both ways of calculating and so is the reconstructed matrix.

As far as I can see, the second approach only helps with two things: 
1. caching the PCA model calculation (which I couldn't see how to do in this case) and
2. Selecting a specific sub-set of components (not just the ones with the highest variance)

Hopefully that is clear (and not super wrong :sweat_smile: )

markotoplak:
> Hopefully that is clear (and not super wrong 😅 )

Thanks. Sorry, I figured this out myself yesterday and I also though I sent that message here, but I seemed to have forgotten.

PR #810, opened 2025-05-23
[ENH] OWFFT: Simplify auto dx handling

stuart-cls:
Following @ngergihun comment in https://github.com/Quasars/orange-spectroscopy/pull/809#issuecomment-2888918610 took the opportunity to simplify our (auto) dx handling.

 - Checkbox is now `dx_auto` which allows user to override any default value
 - `dx` default value is set from data table _if_ `dx_auto` is set, otherwise it's left as a user setting.
 - If no `dx` is present in the dataset _and_ `dx_auto` is set, the value falls back to our old `DEFAULT_HENE`

This gets rid of a bunch of weirdness and generally lets users do what they want.

Still outstanding:

- [ ] Document what dx is ( yes, it's the pathlength step size)
- [ ] Add a settings transistion
- [ ] A little more standardization of the Info box
- [ ] Rebase once #809 is merged

Comments welcome.

### DIFF
--- a/orangecontrib/spectroscopy/tests/test_owfft.py
+++ b/orangecontrib/spectroscopy/tests/test_owfft.py
@@ -7,7 +7,7 @@ from Orange.widgets.tests.base import WidgetTest
 from orangecontrib.spectroscopy.data import getx
 from orangecontrib.spectroscopy.io.neaspec import NeaReaderGSF
 from orangecontrib.spectroscopy import irfft
-from orangecontrib.spectroscopy.widgets.owfft import OWFFT, CHUNK_SIZE
+from orangecontrib.spectroscopy.widgets.owfft import OWFFT, CHUNK_SIZE, DEFAULT_HENE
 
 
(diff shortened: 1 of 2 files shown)


stuart-cls:
@ngergihun Please take another look.

ngergihun:
Hi @stuart-cls !
I think the info is much more descriptive and better for following what is going on.
I have got an error for the settings migration:
image

I guess this is because the dx_HeNe was removed and is not in the settings anymore.

markotoplak:
@ngergihun, did you perhaps test migration with one the workflows that was saved from this branch but before migration PR was added? That would explain the error. Otherwise, @stuart-cls, I do not understand why that code would crash.

ngergihun:
> @ngergihun, did you perhaps test migration with one the workflows that was saved from this branch but before migration PR was added? That would explain the error.

Ohh, yes, sorry. This was the problem. Reset widget settings solved the problem. Sorry for the false alarm.

Issue #747, opened 2024-09-25
Neaspec readers


borondics:
We seem to have duplicate code but there might be a reason for it. 
Specifically, I am referring to `gsf.py` and `neaspec.py` where the functionality of `GSFReader` and `NeaReaderGSF` are very close. @stuart-cls, the log says both were made by you on the same day but I think this is due to the refactoring of the directory tree when we moved the readers in separate files. I vaguely remember working on these but no idea why we have two types.

I like `GSFReader` more because it uses the `_spectra_from_image` helper that will enforce column naming to _map_x_ and _map_y_. 

However, `NeaReaderGSF` reads both phase and amplitude values, which are sometimes required for calculations, so I don't just want to delete those and we also need a meta column in future calculations that contains the phase/amplitude information.

My ideal solution would be:
Unify the two classes into one, where the new `Orange.data.Table` will contain
```
metas: map_x, map_y, type 
data
```

The `type` meta column would contain the phase/amplitude information.

Any suggestions?

raul-freitas:
The Multifile reader has these 2 formats related to Neaspec:
1. NeaSPEC (*.nea *.txt)
2. NeaSPEC raw files (*.gsf)

*.nea files were the very first format of files, long ago discontinued. The *.txt format in 1) is still active as it reads spectra data saved as *.txt. Finally, the *.gsf reader was dedicated to read single optical channel raw interferograms, however, the company no longer saves them in this format. It is important to highlight that 2) has nothing to do with SPM imaging data.

My suggestion is to merge the new multichannel interferograms reader (*.txt) and a unique flag could distinguish between spectra.txt or interferograms.txt. Remember that only interferograms files have the M column (maybe this could be a flag...).

ngergihun:
[...]

PR #813, opened 2025-06-29
[ENH] Unified NeaSpec readers

ngergihun:
Regarding personal discussions and Issue: Neaspec readers #747 , I created a reader that can load all types of NeaSpec data.

.GSF files:
- from imaging measurements

.NEA files:
- old neaspec fileformat for ifg and spectra

.TXT files:
- all kinds of spectra and IFG files, including PTE and nanoRaman

New functionality:
-Sheets: The user can select which channel to load. "All" is the default.

Works for point spectra, linescans, and hyperspectral scans.

It can replace all readers, except @raul-freitas reader, because that has additional processing implemented. Thus, to be backward compatible with old workflows, I did not remove `NeaReaderMultiChannel` yet. If you could move the processing part from `NeaReaderMultiChannel` to `OWFFT` widget than we can also remove this and there is going to be only One Reader, to read them all!

Please check it out and test with your files.


markotoplak:
Can you please clean up commits? With big PRs, 1 commit should be 1 functionality so that these commits, if the PR is long, can be reviewed commit-by-commit. So if you'd like to have your code reviewed in a sensible time frame, it makes sense to think how to ease the review. :)

So commits have two functions: (1) for the developer to have a history and (2) for the reviewer to understand the code. So these are usually, when things are non-trivial, incompatible. So what I tend to do with big stuff in Orange3: I do ensure that it works first, but then I go through my commits and perhaps join some of them, or recommit the code in different chunks so that it makes more sense for the reviewer. I understand this is difficult if you did not think about this in advance (very atomic commits would help), so no need to do it here, but I'd appreciate it in further PRs.

If you go there-and-back a lot, often just making it a single commit is just fine. So maybe squashing it is sufficient. Or maybe squashing a set of commits.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,199 +11,275 @@
...
-                version = 2
-        if version == 1:
-            return self.read_v1()
-        else:
-            return self.read_v2()
+        return waveN, M, meta_data
 
 
 class NeaReaderGSF(FileFormat, SpectralFileFormat):
-
```
markotoplak, line 200: Please take all the linting that is not connected with the PR out.

markotoplak:
So, I see you also edited (linted) some code that is not directly connected to the PR. I want that out, so that it makes review easier.

Perhaps the easiest think to do is to open a new PR and then just copy the parts you want.

ngergihun:
@markotoplak , okay I will condense the things. Indeed, we did a lot of back and forth while trying to finalize what we wanted, and it is the raw thing. I agree to make a new PR with a squashed commit history, and I will try to "delint". :)

PR #816, opened 2025-07-09
[ENH] Add Image Overlay: new widget

ngergihun:
### Functionality: 

The user can choose a feature from the "Overlay Data" input, which is added to the attributes of the "Data" input table as a "visible image". This way, it is possible to add a visible image overlay even if the reader cannot read it from the original data file, or if it is stored separately. 

The widget creates an RGB overlay image from the displayed image. Thus, changing the display parameters directly affects the assigned visible overlay image.

### Outstanding issues:
- [ ] New icon
- [x] Weird test problem

One of the tests does not seem to run correctly, even though in a real application, the same thing works. Somehow, during `test_add_visible_image`, the image acquired from the imageplot is None, because `im = self.imageplot.img.qimage` in `set_vimage` of the OWOverlay. However, everything runs correctly when the widget is used. 

@markotoplak : Could you please assist us with this? What could cause this weird issue with the tests?

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owoverlay.py
+++ b/orangecontrib/spectroscopy/widgets/owoverlay.py
@@ -1,3 +1,6 @@
+import time
+from typing import Optional
+
 import numpy as np
 
 from AnyQt.QtCore import Qt, QBuffer
@@ -8,13 +11,14 @@ from Orange.widgets.settings import DomainContextHandler
 from Orange.widgets.utils.itemmodels import DomainModel
 from Orange.widgets.widget import OWWidget, Input, Output, Msg
 from Orange.widgets import gui, settings
+from Orange.widgets.utils.concurrent import TaskState, ConcurrentWidgetMixin
 
 from orangewidget.settings import SettingProvider, ContextSetting
 
 import io
 from orangecontrib.spectroscopy.io.util import ConstantBytesVisibleImage
 from PIL import Image
-from orangecontrib.spectroscopy.utils import get_ndim_hyperspec
+from orangecontrib.spectroscopy.utils import get_ndim_hyperspec, InvalidAxisException
 
 from orangecontrib.spectroscopy.widgets.owhyper import BasicImagePlot
 
@@ -28,6 +32,14 @@ class AImagePlot(BasicImagePlot):
         super().__init__(parent)
         self.axes_settings_box.hide()
         self.rgb_settings_box.hide()
+        self.image_updated.connect(self.drawing_done)
+
+    def update_view(self):
+        self.drawn = False
+        super().update_view()
+
+    def drawing_done(self):
+        self.drawn = True
 
     def add_selection_actions(self, _):
         pass
@@ -36,19 +48,11 @@ class AImagePlot(BasicImagePlot):
         pass
 
 
-class ImagePreview:
-    imageplot = SettingProvider(AImagePlot)
-
-    value_type = 1
-
-    def __init__(self):
-        self.imageplot = AImagePlot(self)
-
-    def shutdown(self):
-        self.imageplot.shutdown()
+class InterruptException(Exception):
+    pass
 
 
-class OWOverlay(OWWidget):
+class OWOverlay(OWWidget, ConcurrentWidgetMixin):
     name = "Add Image Overlay"
     description = "Add an image that can be displayed in Hyper Spectra to the dataset."
     icon = "icons/bin.svg"
@@ -93,7 +97,8 @@ class OWOverlay(OWWidget):
         return None
 
     def __init__(self):
-        super().__init__()
+        OWWidget.__init__(self)
+        ConcurrentWidgetMixin.__init__(self)
 
         self.maindata = None
         self.data = None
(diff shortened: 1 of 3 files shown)


borondics:
The functionality is great for me. This is what we discussed with @markotoplak.

Here is an icon suggestion.

OWOverlay

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def __init__(self):
+        self.imageplot = AImagePlot(self)
+
+    def shutdown(self):
+        self.imageplot.shutdown()
+
+
+class OWOverlay(OWWidget):
+    name = "Add Visible"
```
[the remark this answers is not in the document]
borondics: Yes, we need to make this general. "Add Image Overlay" works for me.

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -0,0 +1,281 @@
...
+                                        pos_x=posx-xres/2,
+                                        pos_y=posy-yres/2,
+                                        size_x=width+xres,
+                                        size_y=height+yres,
+                                        image_bytes=img_bytes,
+                                        )
+                
+                # Assign it to the datatable attributes
+                if self.maindata and vimage is not None:
+                    self.maindata.attributes['visible_images'] = [vimage]
```
markotoplak, line 234: We should NEVER modify the input data in any way. We really have to ensure that the widgets only work forward, not edit the input data.

markotoplak: Here, copying the data, and (maybe also) that dictionary, is needed. Also, there should be a test that ensures that input data dict did not change at all.

markotoplak: Also, it is wrong that visible images are replaced (this one should be added).

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -0,0 +1,281 @@
...
+                buffer.open(QBuffer.ReadWrite)
+                im.save(buffer, "PNG")
+                pil_im = Image.open(io.BytesIO(buffer.data()))
+                pil_im = pil_im.transpose(Image.FLIP_TOP_BOTTOM)
+                img_bytes = io.BytesIO()
+                pil_im.save(img_bytes, format='PNG')
+
+                # Need to modify the position and the scale since visual imageplot 
+                # place the corner of the pixel to the given position
+                vimage = ConstantBytesVisibleImage(name="External Image",
```
markotoplak, line 224: Because images should be added, an unused name should be chosen here (Orange has functions for choosing new names)

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -0,0 +1,281 @@
...
+    def _update_attrs(self):
+        self.imageplot.attr_x = self.attr_x
+        self.imageplot.attr_y = self.attr_y
+        self.redraw_data()
+    
+    def redraw_data(self):
+        self.imageplot.update_view()
+        self.commit.deferred()
+    
+    def set_vimage(self, data):
```
markotoplak, line 193: The reason that tests are failing is that this function is sloppy. This function should be asynchronous (threaded), because it relies on painting that is itself asynchronous. The check `if im is not None` should not be used for avoiding outputs if the image is not yet draw. Image should always exist here unless there was an error.

ngergihun:
@markotoplak Thanks for the suggestions. I tried to correct some parts, and the tests now seemed to run. I was unsure which native Orange functions you were referring to for choosing new names, so I've added some basic incremental overlay image name changes and included a test for this. This can be changed, of course.

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -246,9 +248,14 @@ def set_vimage(self, data):
                     image_bytes=img_bytes,
                 )
 
-                # Assign it to the datatable attributes
-                if self.maindata and vimage is not None:
-                    self.maindata.attributes["visible_images"] = [vimage]
+                # # Assign it to the datatable attributes
+                if newmaindata and vimage is not None:
+                    if "visible_images" in list(newmaindata.attributes):
```
markotoplak, line 269: `.attributes` is a dict, so conversion to list is not needed (and it actually makes it, O(n), while dict lookup is O(1); though that is not important here). So: `if "visible_images" in newmaindata.attributes:`

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -0,0 +1,317 @@
...
+                    return f"{base_name} {i}"
+
+                # Update name
+                allnames = []
+                if "visible_images" in list(newmaindata.attributes):
+                    allnames = [
+                        im.name for im in newmaindata.attributes["visible_images"]
+                    ]
+                basename = "Overlay Image"
+                name = get_unused_name(allnames, basename)
```
markotoplak, line 255: @ngergihun, `get_unique_names` from `Orange.data.util` :)

orangecontrib/spectroscopy/widgets/owoverlay.py
```diff
@@ -0,0 +1,317 @@
...
+                    if base_name not in existing_names:
+                        return base_name
+                    i = 1
+                    while f"{base_name} {i}" in existing_names:
+                        i += 1
+                    return f"{base_name} {i}"
+
+                # Update name
+                allnames = []
+                if "visible_images" in list(newmaindata.attributes):
```
markotoplak, line 250: Don't convert to `list`

markotoplak:
> Here is an icon suggestion.

@borondics, I like the look of icon, but apart from in the browser it does not render well for me. I could also not open it in inkspace nicely to reconvert (it shows some gibberish).

ngergihun:
@markotoplak Thanks for the reviews and suggestions. I changed to `get_unique_names` and corrected my mistake with the list conversion. Originally I wanted to write `list(newmaindata.attributes.keys())` but just `newmaindata.attributes` is much nicer, thanks.

markotoplak:
I think the new icon is wonderful. Thanks!

PR #817, opened 2025-07-20
[ENH] Unified NeaSpec readers (breaking change for .nea files)

ngergihun:
### Description:
Updated pull request for unifying neaspec readers.

**Updated:** NeaReader
**Removed:** NeaImageGSF

### Changed functionality:
NeaReder: Can read all the following data formats from neaspec.

.GSF files: from imaging measurements

.NEA files: old neaspec fileformat for ifg and spectra

.TXT files: all kinds of spectra and IFG files, including PTE and nanoRaman

### New functionality:
NeaReader:

-Sheets: The user can select which channel to load. "All" is the default.

### Notes:
Based on personal discussions and Issue: Neaspec readers https://github.com/Quasars/orange-spectroscopy/issues/747 , I created a reader that can load all types of NeaSpec data.

### DIFF
--- a/orangecontrib/spectroscopy/io/neaspec.py
+++ b/orangecontrib/spectroscopy/io/neaspec.py
@@ -260,6 +260,24 @@ class NeaReader(FileFormat, SpectralFileFormat):
         domain = Orange.data.Domain([], None, metas=metas)
         meta_data = Table.from_numpy(domain, X=np.zeros((len(M), 0)), metas=Meta_data)
         # Add measurement parameters to attributes
+
+        if is_ifg:
+            # Taken from old NeaReaderMultiChannel
+            # calculate datapoint spacing in cm for the fft widget as the optical path
+            dx = 2 * np.mean(np.diff(new_maxis)) * 1e2  # convert [m] to [cm]
+            # check file headers for wavenumber scaling factor
+            # and apply it to the calculated spacing
+            try:
+                wavenumber_scaling = measparams["WavenumberScaling"]
+                wavenumber_scaling = float(wavenumber_scaling)
+                dx = dx / wavenumber_scaling
+            except KeyError:
+                pass
+            # register the calculated spacing in the metadata
+            measparams["Calculated Datapoint Spacing (Δx)"] = ["[cm]", dx]
+            measparams["Domain Units"] = "[µm]"
+            measparams["Channel Data Type"] = "Polar", "i.e. Amplitude and Phase separated"
+
         meta_data.attributes = measparams
 
         return waveN, M, meta_data
--- a/orangecontrib/spectroscopy/tests/test_readers.py
+++ b/orangecontrib/spectroscopy/tests/test_readers.py
@@ -479,6 +479,18 @@ class TestNea(unittest.TestCase):
         self.assertEqual("O0A", data.metas[2][3])
         self.assertEqual("O0P", data.metas[3][3])
 
+    def test_ifg_read_info(self):
+        fn = 'NeaReaderMultichannel_test/Test_Au_Fourier_Scan_Synchrotron.txt'
+        absolute_filename = FileFormat.locate(fn, dataset_dirs)
+        data = NeaReader(absolute_filename).read()
+        self.assertEqual(len(data), 30)
+        self.assertEqual("channel", data.domain.metas[3].name)
+        self.assertEqual("O0A", data.metas[2][3]) # New reader has more channels
+        self.assertEqual("O0P", data.metas[3][3])
+        self.assertEqual(data.attributes['Channel Data Type'][0], 'Polar')
+        self.assertEqual(data.attributes['Calculated Datapoint Spacing (Δx)'][0], '[cm]')
+        self.assertEqual(data.attributes['Scan'], 'Fourier Scan')
+        check_attributes(data)
 
 class TestNeaGSF(unittest.TestCase):
 


orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -13,7 +13,7 @@
 
 from pySNOM import readers
 
-class GeneralNeaReader(FileFormat, SpectralFileFormat):
+class NeaReader(FileFormat, SpectralFileFormat):
```
markotoplak, line 17: TODO: Check if this renaming breaks saved workflows.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -340,45 +430,6 @@ def _gsf_reader(self, path):
         return np.asarray(X)
 
 
-class NeaImageGSF(FileFormat, SpectralFileFormat):
```
markotoplak, line 343: Check if removing this causes issues with saved workflows.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,195 +11,285 @@
 from orangecontrib.spectroscopy.io.util import SpectralFileFormat, _spectra_from_image
 from orangecontrib.spectroscopy.utils import MAP_X_VAR, MAP_Y_VAR
 
+from pySNOM import readers
```
markotoplak, line 14: This live is in wrong position (see lint).

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,195 +11,285 @@
...
+
+        if self.filename.endswith(".nea"):
+            Max_omega = int(measparams["PixelArea"][2])
+        else:
+            # Neaspec have several the naming
+            # for the indepedent variable index
+            if "Depth" in list(data.keys()):
+                Max_omega = int(np.max(data["Depth"]) + 1)
+            elif "Index" in list(data.keys()):
+                Max_omega = int(np.max(data["Index"]) + 1)
```
markotoplak, line 126: Why int(n+1) ? I saw these were present before, but why?

If for rounding, this is a particularly bad way to do it. The reason deserves a comment in code.

markotoplak: Also, the code is full of needless conversion to list when searching in a dictionary.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,195 +11,285 @@
...
-        waveN = file[0:int(Max_omega), 3]
-        metas = [Orange.data.ContinuousVariable.make("row"),
-                 Orange.data.ContinuousVariable.make("column"),
-                 Orange.data.StringVariable.make("channel")]
+        # Neaspec tends to name the independent variable differently all the time
+        # Try to prepare for all the names we know so far
+        if "Run" in list(data.keys()):
+            waveN = data["M"][0 : int(Max_omega)] * 1e6
+            metas = [
+                Orange.data.ContinuousVariable.make(MAP_X_VAR),
```
markotoplak, line 262: Variable.make is not needed anymore. Also, these code .

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,195 +11,285 @@
...
+                Orange.data.StringVariable.make("channel"),
+            ]
+        elif "Wavelength" in list(data.keys()):
+            waveN = data["Wavelength"][0 : int(Max_omega)]
+            metas = [
+                Orange.data.ContinuousVariable.make(MAP_X_VAR),
+                Orange.data.ContinuousVariable.make(MAP_Y_VAR),
+                Orange.data.StringVariable.make("channel"),
+            ]
+        else:
```
markotoplak, line 274: Last two repeat a lot. Could be unified with a small exception.

Code here repeats a lot and could be shorter. At least first part of metas, with positions, can be created once before the ifs.

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,195 +11,285 @@
...
+            chn = self.sheets[0]
+            self.sheet = chn
+
+        if chn == "All":
+            N_chn = len(self.sheets) - 1
+        else:
+            N_chn = 1
+
+        # There are strored in the measparams
+        Max_row = int(np.max(data["Row"]) + 1)
```
markotoplak, line 109: Also here, why int() and +1?

If this was an int value anyway, and +1 is added to get a count, then name is wrong. Then it should be `n_rows` or something similar.

markotoplak:
@ngergihun, I did a coarse reading of the PR and wrote some comments. I can make the code prettier (=shorter) myself, so doing that should not be a priority.

But I am really interested why is the int(something.max() + 1) pattern in there... That I can not answer myself. :)

orangecontrib/spectroscopy/io/neaspec.py
```diff
@@ -11,195 +11,285 @@
...
+            height = Max_col
+        if "ScannerCenterPosition" in list(measparams.keys()):
+            xoff = measparams["ScannerCenterPosition"][0]
+            yoff = measparams["ScannerCenterPosition"][1]
+        else:
+            xoff = 0.0
+            yoff = 0.0
+
+        # Create the list of points centered to the origo
+        # x = np.linspace(-width / 2, width / 2, Max_row)
```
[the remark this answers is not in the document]
ngergihun: Mistake, forgot to remove. x and y are already in
`X, Y = np.meshgrid(
            np.linspace(-width / 2, width / 2, Max_row),
            np.linspace(-height / 2, height / 2, Max_col),
        )`

ngergihun:
> Interpolation is not used anymore in the new version. Why was it needed before? If it is not needed anymore, the import that lint complains about should also be removed.

Yes, before it was doing data interpolation, but we discussed with Feri that the reader should only read. Import can be removed.

markotoplak:
> @markotoplak, thanks for checking the PR. Yes, my eyes had already ignored them; I had forgotten it so long ago. There was a mistake in the pysnom header reader, and instead solving it there right away I just quickly put a conversion here and of course it was stupidly copied again without paying attention.

But what exactly is the issue? What I see is that these numbers tend to be used as counts in the code. In this case, they are just named badly.

> The "list(data.keys())" mistake comes from my lack of knowledge that its so slow.
Not only it is slow, it also copies a  objects into a new data structure (and thus getting and releasing new memory). All type conversion should be always looked at as questionable.

ngergihun:
> > @markotoplak, thanks for checking the PR. Yes, my eyes had already ignored them; I had forgotten it so long ago. There was a mistake in the pysnom header reader, and instead solving it there right away I just quickly put a conversion here and of course it was stupidly copied again without paying attention.
> 
> But what exactly is the issue? What I see is that these numbers tend to be used as counts in the code. In this case, they are just named badly.
> 

There is no issue now, it was a very long time ago. The `data["whatever"]` values were floats. This was fixed already. 
However, at the same time, I noticed that data["whatever"] are numpy ints and the measparams["whatever"] are native. I don't know if this will matter at some point.

> > The "list(data.keys())" mistake comes from my lack of knowledge that its so slow.
> 
> Not only it is slow, it also copies a objects into a new data structure (and thus getting and releasing new memory). All type conversion should be always looked at as questionable.

Got it, makes sense, thanks for the explanation.

markotoplak:
> > But what exactly is the issue? What I see is that these numbers tend to be used as counts in the code. In this case, they are just named badly.
> 
> There is no issue now, it was a very long time ago. The `data["whatever"]` and some `measparams["whatever"]` values were floats. This was fixed already. However, at the same time, I noticed that `data["whatever"]` are numpy ints and the `measparams["whatever"]` are native ints. I don't know if this will matter at some point.

OK, there is no type issue. But I still see a possible naming issue. Please state which one of two following option describes the current code:
1. `Max_row` is the maximum value of "Row"
2. `Max_row` is number of different "Row" values (or a "Row" count)

If (2), then naming is misleading.

markotoplak:
> How is this handled in the new version? Is there any tests case ensuring this is done correctly?

OK, I'll answer myself: I am quite certain it is handled wrongly and the tests that was there to ensure correctness was just modified to confirm to the new implementation. **Tests are there to make us think what the results should be and not just to make them pass.** I'll comment more in that particular test.

ngergihun:
@markotoplak, you are right, I assumed all the same, which is actually true for the spectrum files, because there are no runs, but it is not correct for the interferograms, where M is different for sure. We have to think about it, because it is not trivial how to do it the best way. It can change the data in an unexpected way, creating spectral artifacts.

orangecontrib/spectroscopy/tests/test_readers.py
```diff
@@ -442,25 +442,43 @@ class TestNea(unittest.TestCase):
...
-        self.assertEqual(len(data), 12)
-        self.assertEqual("channel", data.domain.metas[2].name)
-        np.testing.assert_almost_equal(getx(data), [0.000295, 0.00092])
-        self.assertEqual("O0A", data.metas[0][2])
-        np.testing.assert_almost_equal(data.X[0, 0], 10.2608052)  # O0A
-        self.assertEqual("O0P", data.metas[6][2])
-        np.testing.assert_almost_equal(data.X[6, 0], 0)  # O0P
+        self.assertEqual(len(data), 20)
+        self.assertEqual("run", data.domain.metas[2].name)
+        self.assertEqual("channel", data.domain.metas[3].name)
```
markotoplak, line 447: Also, I printed out the table, and "Row" and "Column" are read as (-0.5, -0.5), even though there is only (0, 0) in the file. Why did this get changed?

The fixed version of this tests should also test if "Row" and "Column" were read correctly.

markotoplak:
> We have to think about it, because it is not trivial how to do it the best way. It can change the data in an unexpected way, creating spectral artifacts.

Conceptually, I see two options:
1. We can interpolate as we used to, which of course can also introduce certain artifacts.
2. We just return a table of what was read (with lots unknowns), thus outputting all possible data and not make any assumptions when reading it, and leave this to the user to manage.

Option (2) makes the data harder to use and if none of the widgets have the flexibility to handle that additional information, users will need to interpolate anyway and we get something equivalent to (1), just with more effort. :) But if some downstream widgets could handle that effectively and smarter than just interpolating in that space, well, then (2) could be useful.

markotoplak:
@ngergihun, thank you!

borondics:
This was a big step! Thanks guys!!!

PR #820, opened 2025-07-22
[ENH] PTIR Loader: load raw discrete frequency data

callumgassner:
Introduces the ability to load the raw data for discrete frequency maps (O-PTIR, DC, Phase). Selection of full spectra and df maps is achieved through the "Sheets" drop down in File and Multifile.

I had to make some adjustments to my original implementation due to changes from #788 (particularly the inclusion of z-focus). I have tested these changes using a number of PTIR Studio files containing a various mix of data/measurement types and have not come across any issues. However, I haven't been able to test e.g. 1 file containing 2 maps of different sizes (if that is even possible for these files? I don't have much experience with them).

### DIFF
--- a/orangecontrib/spectroscopy/io/ptir.py
+++ b/orangecontrib/spectroscopy/io/ptir.py
@@ -41,6 +41,21 @@ class PTIRFileReader(FileFormat, SpectralFileFormat):
                         channel_map[signal] = label
                 except:
                     pass
+
+        for discrete_maps in filter(lambda s: s == 'Heightmaps', keys):
+            hdf5_meas = hdf5_file[discrete_maps]
+            meas_keys = list(hdf5_meas.keys())
+            meas_attrs = hdf5_meas.attrs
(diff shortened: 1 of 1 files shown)


stuart-cls:
I will take a look and @kaiyang-tu has some good test files.

kaiyang-tu:
@callumgassner this is a nice add. Regarding your question, yes a file can contain two maps of different size (see attached file [discrete-OPTIR-test.zip](https://github.com/user-attachments/files/21497471/discrete-OPTIR-test.zip)).

Short chat with @stuart-cls, if more than one discrete maps exist in single file, it might be good to give them a meta information (e.g. _Region 1, Region 2_ to differentiate maps. Related, z-focus should be consistent for each wavenumber when collecting a single discrete map. This could be bundle as one z-focus value per _Region_. Something along the lines of: 

meta region | meta z-focus | features
_Region 1_ | z-focus value | map1 values
_Region 2_ | z-focus value | map2 values

PR #822, opened 2025-07-31
[FIX] owhyper: Show RGB limits settings in color box when selected

stuart-cls:
Following up #777 to show RGB limits in the "Image colors" group in the controls area.

image

This also fixes a weird bug where the RGB limits were visible in the inset Menu only on old installs, but not if you ran OWHyper stand-alone or started Quasar with `--clear-widget-settings`. Settings, amirite


borondics:
Hi Stuart,

This is a great addition. However it made me think of another issue I had
and now it is amplified.

We are using the side menu more and more and it is getting long... We
should find a nice and visually pleasing l way of compressing it. One
solution could be making these units like RGB an expandable section but we
could also come up with a different control than just a list of labels and
text boxes for every function.

Cheers,
Feri

markotoplak:
Thank you, this looks good, sorry for the oversight when we moved the options.

Issue #640, opened 2022-10-30
Integrator ideas


borondics:
Here is a screenshot for inetgrator ideas. We already have a few implemented and I don't think we should implement all, but a few are still missing.

<img width="488" alt="image" src="https://user-images.githubusercontent.com/6304794/198902200-be5eb5b6-6497-4ba2-a81b-92a8f1f6fff2.png">

stuart-cls:
See also #82

Issue #549, opened 2021-03-25
New area integration with independent baseline 


clsandt:
Hello
it would be necessary to modify the Integrate spectra and Preprocess spectra widgets with new types of peak integration.

The first type would be a modified 'Intergrate from baseline' area integration with a baseline independent from the peak edges as in th picture below:
![image](https://user-images.githubusercontent.com/20243766/112501988-fbd4df00-8d89-11eb-84b8-498a720fcf40.png)

The second type would also be a modified 'Integrate from baseline' area integration with an independent baseline but instead of defining two set points for the baseline, it should be possible to define two ranges where the algorithm will search for the lowest or highest or average point to define the X position of the baseline point. This is illustrated in the next picture:
 
![image](https://user-images.githubusercontent.com/20243766/112502382-51a98700-8d8a-11eb-9ca7-fe349bad4071.png)

Thanks

Christophe

stuart-cls:
Thanks Christophe, these are great suggestions!

Issue #82, opened 2017-07-11
Preprocessor widget or new Peak measurement widget


clsandt:
It would be very useful to add new tools in the preprocessor widget or to create a new widget to extract parameters from peaks such as
- position of the maximum
- position of the center of symmetry
- integrated area (possibly with different baseline options) I know such integration already exists in the Preprocess spectra widget
-ratio of integrated areas or heights
- full width at half height

stuart-cls:
These are great suggestions. Since they are a similar type of "transformation" of the data as Integration[1], I wonder if the new Integrate widget could be generalized into a "Peak Analysis" or similar?

[1] which already includes things that aren't integration, such as peak max and closest value

clsandt:
A single "Peak analysis" widget with centralized options would be great.
Maybe a noise measurement (RMS and/or P2P) could be integrated in the same widget.

Issue #26, opened 2016-07-13
Integrate: Select integration method using figure


stuart-cls:
Replace unclear method text with figures illustrating the integration method to be used.

PR #823, opened 2025-09-10
[ENH] Add "Absolute integral from baseline" peak analysis option

graceflaman:
Hi All,

I've been adding some preprocessors to the Integrate Spectra widget with help from @stuart-cls. I'm starting this PR to track these developments. I added a little bit of love to the documentation for this widget -- I think having an explanation/visual/equation of each integrator type might help. This is along the lines of issue #26 .

I am working towards adding a few more preprocessors referencing some open issues #640 #549 and #82 . Here is a list of some of the other preprocessors I'm working on/would like to add:
- [ ] Opus-like RMS noise
- [ ] Allan deviation
- [ ] Standard deviation
- [ ] Integral ratio
- [ ] Peak-to-peak ratio
- [ ] signal-to-noise ratio

Any input and suggestions are much appreciated!

Thanks,

Grace Flaman

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/integrate.py
+++ b/orangecontrib/spectroscopy/preprocess/integrate.py
@@ -323,16 +323,17 @@ class Integrate(Preprocess):
 
     INTEGRALS = [IntegrateFeatureSimple,
                  IntegrateFeatureEdgeBaseline,
-                 IntegrateFeatureEdgeBaselineAbsolute,
                  IntegrateFeaturePeakSimple,
                  IntegrateFeaturePeakEdgeBaseline,
                  IntegrateFeatureAtPeak,
                  IntegrateFeaturePeakXSimple,
                  IntegrateFeaturePeakXEdgeBaseline,
-                 IntegrateFeatureSeparateBaseline]
+                 IntegrateFeatureSeparateBaseline,
+                 IntegrateFeatureEdgeBaselineAbsolute,
+                 ]
 
     # Integration methods
-    Simple, Baseline, BaselineAbsolute, PeakMax, PeakBaseline, PeakAt, PeakX, PeakXBaseline, Separate = INTEGRALS
+    Simple, Baseline, PeakMax, PeakBaseline, PeakAt, PeakX, PeakXBaseline, Separate, BaselineAbsolute = INTEGRALS
 
     def __init__(self, methods=Baseline, limits=None, names=None, metas=False):
         self.methods = methods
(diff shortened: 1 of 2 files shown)


graceflaman:
Hi @markotoplak - thanks for checking in! I have time to address this and have now rebased. I will check in with @stuart-cls for help if I need.

- [ ] new integral into `scipi.integrate.trapezoid`

orangecontrib/spectroscopy/preprocess/integrate.py
```diff
@@ -304,6 +323,7 @@ class Integrate(Preprocess):
 
     INTEGRALS = [IntegrateFeatureSimple,
                  IntegrateFeatureEdgeBaseline,
+                 IntegrateFeatureEdgeBaselineAbsolute,
```
stuart-cls: It makes sense to add this here, logical grouping and order. And it works fine in the **Integrate** widget.

However, we use this `INTEGRALS` list in two other places (Integrate Preprocessor, and HyperSpectra) and these ones rely on the ordering of the list being fixed:

```integrate.py
Integrators_classes = Integrate.INTEGRALS[:-1]  # without SeparateBaseline
```

```owhyper.py
integration_methods = Integrate.INTEGRALS
```

And so when you open an old workflow with option `IntegrateFeaturePeakSimple` saved, you instead get this new preprocessor.

[grace823-before.ows.xml](https://github.com/user-attachments/files/25025182/grace823-before.ows.xml)

@markotoplak Any opinion on how to handle this? Either way we will have to touch at least the preprocessor, since we currently exclude SeparateBaseline there by `[:-1]` indexing.

stuart-cls:
@graceflaman Except for the issue noted above, this looks good to me. Tests pass locally (not sure what's going on with the CI). The only outstanding lint issue is a new line-too-long, let's leave it.

orangecontrib/spectroscopy/preprocess/integrate.py
stuart-cls: Okay that approach sounds good.

Regarding the meta-question of the `Integrate` preprocessor, I guess you mean that each method would be a single preprocessor? Is there an example in our repo or in Orange3 that would be the gold standard?

stuart-cls:
I just changed the order to fix the settings issue. I'm not too worried about the deprecated Preprocess Spectra editor, but at least it's not broken now.

@graceflaman Do you think this value has any use in the list in hyperspectra? Currently it is there, at the bottom. But maybe we should curate that interactive list as @markotoplak suggested.

stuart-cls:
Test failures unrelated. Thanks @graceflaman !

markotoplak:
Thank you, @graceflaman and @stuart-cls.

PR #825, opened 2025-10-14
CurvePlot: Markings count as Items (for bounds corrections) 

### LINKED ISSUE #592: Show everything in Peak Fit
Currently the Peak Fit widget automatic zoom behaves the same as Spectra, i.e., it rescales considering only the spectra.
image

Instead of this behavior could we have the autoscale include all the fitted curves as well? Similar to this:
image

stuart-cls:
Fixes #592

Originally I was going to only change the `ignoreBounds` kwarg for PeakFit, but actually I can't find any reason why not to do it in general. I tested on all the widgets that use markings (Preprocess, Hyper, Integrate) and they all seem to do the right thing?

Of course now if your initial guesses in the fit are way off, that will be shown and the only way to get back to your data is to zoom in.

@markotoplak the override of `ignoreBounds` seems to be quite old, perhaps the reason for it is no longer relevant?

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owspectra.py
+++ b/orangecontrib/spectroscopy/widgets/owspectra.py
@@ -1506,7 +1506,7 @@ class CurvePlot(QWidget, OWComponent, SelectionGroupMixin):
 
     def add_marking(self, item):
         self.markings.append(item)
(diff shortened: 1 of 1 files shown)


markotoplak:
I forgot what was the reason to ignoreBounds. Maybe there is none anymore...

PR #828, opened 2025-10-30
[ENH] OWStackAlign: optionally use reference images

ngergihun:
**Description:**

This modification intends to improve the stackalign widget by introducing a reference input. This way, a tracking image set can be used to calculate the image drifts. The calculated drifts are used to correct and crop the images from the input stack.

**Motivation:**

Images from optical hyperspectral data show different contrast in the image plane depending on the wavelength. At the wavelengths matching the absorption of some components in the sample, we experience high optical contrast, but at off-resonant wavelengths, the image contrast changes drastically. This confuses image registration algorithms. As a consequence, the calculated image drifts are incorrect and lead to false results.

**Solution:**

The tracking problem can be solved by using images with a fixed image feature contrast for all wavelengths. I scanning probe optical microscopy, multiple measurement channels are recorded at the same time, for example, topography, or tip oscillation amplitude. Since these data are recorded simultaneously, they can be used as tracking images to extract the image drifts.

**Implementation:**

The `alignstack` function is now split into two parts:
1. `calculate_stack_shifts`: calculates the image shifts from the given stack of images
2. `alignstack_with_shifts`: correct the data image set using the shift calculated in 1.

The usage of the reference input is optional: the "Use reference images" checkbox enables the method. When the reference data is connected but not used, a warning notifies the user.

The number of images in both datasets has to be the same; thus, an error pops up when it is not correct. The users have to know their datasets and keep in mind that the order of the data has to be the same in both data tables.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owstackalign.py
+++ b/orangecontrib/spectroscopy/widgets/owstackalign.py
@@ -294,9 +294,8 @@ class OWStackAlign(OWWidget):
     def handleNewSignals(self):
         self.Warning.missing_reference.clear()
 
-        if self.data is not None:
-            self.closeContext()
-            self.openContext(self.data)
+        self.closeContext()
+        self.openContext(self.data)
 
         if self.refdata is not None or self.data is not None:
             self._sanitize_ref_frame()


orangecontrib/spectroscopy/widgets/owstackalign.py
```diff
@@ -286,31 +286,30 @@ def _update_attr(self):
...
-            self._sanitize_ref_frame()
+
+        if self.refdata is not None:
+                self.openContext(self.refdata)
+                self._sanitize_ref_frame()
         else:
             self.refdata = None
+
+        if self.data is not None:
+                self.openContext(self.data)
```
markotoplak: `openContext` is something that can only be called once per connections.

orangecontrib/spectroscopy/widgets/owstackalign.py
```diff
@@ -286,31 +286,30 @@ def _update_attr(self):
...
-        self.openContext(refdataset)
-        if refdataset is not None:
-            self.refdata = refdataset
-            self._check_use_ref()
-            self._sanitize_ref_frame()
+
+        if self.refdata is not None:
+                self.openContext(self.refdata)
+                self._sanitize_ref_frame()
         else:
```
markotoplak: This should not set self.data to None (that will fight the input and the state will depend on how things are connected. So the inputs have to set something that needs to stay as such, but of course we can have another layer of attributes that has the active ones.

orangecontrib/spectroscopy/widgets/owstackalign.py
ngergihun: Thanks for the comment. Yes, and I realized that I only need it for the data.

PR #829, opened 2025-11-12
io.ascii: Try to skip header rows without a comment delimiter

stuart-cls:
From the workshop, a spectrometer CSV export type with format

```
Wavenumber,Intensity
650.4205,39.928503
651.3523,40.086846
```

Fails to load since the header is not commented out in any way. 

I added an arbitrary `skiprows` range of 3, but 1 would be sufficient for the specific example.
I also used `.dpt` for the test to avoid the annoying `.csv` collision with Orange Table format.


stuart-cls:
Yes, the `delimiter=None` setting defaults to whitespace in `np.loadtxt`

PR #830, opened 2025-11-27
[ENH] Spectral Series: visual parameter adjustment

borondics:
This has been missing since now - now I copy-pasted it from `Spectra`. I think we probably said that we will make it work at some point.

- [ ] Do we want to use `InteractiveViewBox` from the `Spectra` widget to provide more interactive zooming?


borondics:
If #839 is merged, we can delete this.

PR #831, opened 2025-12-09
[ENH] Hyper Spectra: show as scattter plot

markotoplak:
image

This is a very-hacked-together attempt.

What is missing:
- [ ] Ensure this does not crash the image display
- [x] Reuse colors for the Scatter Plot smartly (now it is too slow)
- [ ] Point size setting (probably in visual plot settings)
- [ ] Smarter redrawing (each property change now redraws the scatterplot multiple times)
- [ ] Tests


markotoplak:
> Hey Guys! I also tested with some of my data. Looks very cool and seems to work very well, although I only went up to 35k spectra. :) The selections seem to work as well, except for the polygon selection.

What was wrong with the polygon selection? Which mode, what happened?

Issue #832, opened 2025-12-11
[TEST] LogisticRegressionLearner behaves differently in scikit-learn 1.8


stuart-cls:
This is causing some tests to fail:

`TestConversion.test_predict_different_domain` and friends which use this learner as a proxy for testing small changes.

PR #833, opened 2025-12-11
[MNT] Hold scikit-learn<1.8 until #832 can be resolved

stuart-cls:


stuart-cls:
Passes now, except for failures on orange-latest from #834

PR #835, opened 2025-12-11
[MNT] Tests: replace Logistic Regression with Random Forest

### LINKED ISSUE #832: [TEST] LogisticRegressionLearner behaves differently in scikit-learn 1.8

markotoplak:
The problem is that LR from scikit-learn 1.8 is incompatible with Orange3 <= 3.39.0.

LR was chosen as the learner of choice for these because it was supposed to be the most stable and also was supposed to be theoretically guaranteed to converge to the global optimum. Oh well... Someday we can get it back.

Fixes #832

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -39,7 +39,7 @@ requirements:
     - h5py
     - extranormal3 >=0.0.3
     - renishawWiRE >=0.1.8
-    - pillow >=9.0.0
+    - pillow >=9.2.0
     - lmfit >=1.3.3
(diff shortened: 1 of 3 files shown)


stuart-cls:
I uploaded opusFC 3.14, so it fails for the right reasons now :) I only re-ran the macOS 3.14 one but you can see the errors there.

PR #837, opened 2025-12-12
try testing without opusfc

markotoplak:


markotoplak:
I have no idea why Windows 3.14 tests fail. I suspected something was wrong with opusFC - but that is not the case.

PR #839, opened 2026-01-15
[ENH] Spectra: line width as plot property

### LINKED ISSUE #544: Visual features for the Spectra widget

markotoplak:
Fixes #544

Not automatic, but plot property.

The disadvantage of this PR is that newly saved workflows where line thickness was changed will fail to load on old versions. Not solvable, but probably not a too big of an issue.

### DIFF
--- a/orangecontrib/spectroscopy/widgets/owspectra.py
+++ b/orangecontrib/spectroscopy/widgets/owspectra.py
@@ -89,6 +89,9 @@ def selection_modifiers():
 class ParameterSetter(CommonParameterSetter):
 
     VIEW_RANGE_BOX = "View Range"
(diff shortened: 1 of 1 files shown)


borondics:
Well, don't we all love how macs are different than other stuff? :) Speed depends on age... my oldie is slow but those fancy people with new computers. :D

PR #842, opened 2026-02-03
[MAINT] Switch to pyproject.toml, enable ruff linting / formating, setup pre-commit

stuart-cls:
A few more changes than I suggested, but I think it's worth the churn so we can stop fighting the differential pylint

Still todo:
 - [x] pre-commit
 - [x] ruff format

If we like it
 - [ ] update contributing.md , README.md


markotoplak:
Thanks! I changed the author for the big ruff commit to `Ruff <ruff@formatter.worked` so that a git blame will immediately point out the reason for the changes. The rest remained the same.

Issue #841, opened 2026-01-24
Add offset correction based on minimal or percentile value in preprocessors


pisarik:
Hi!

### Problem

My spectra are shifted by different values. I would like to find out the individual shift for each spectrum and subtract it, so that the spectra are laying on the zero line.

### Closest existing solutions

At the moment, the closest solutions I could find are:
1. Baseline correction, but this one has only "Linear" and "Rubber band" options
2. Shift and scale, but this one takes only a constant value inputted manually
3. EMSC with polynomial order 0 and Scaling ticked off. This is very close, but requires a reference spectrum

### Suggested solution

The simplest would be to subtract the min value, but due to the noise it is often practical to subtract a "soft minimum", e.g. 3rd percentile. In addition, it is often practical to search this value in a particular region of a spectrum. So, I would suggest something like this:

```python
# Pseudocode
def minimum_offset_correction(spectra, percentile, reg_low, reg_high):
    region = spectra[reg_low: reg_high]
    # percentile can be 0 for the minimum value
    offsets = np.percentile(region, percentile, axis=-1, keepdims=True)
    corrected = spectra - offsets
    return corrected
```

### Implementation

Add an additional option in the Baseline correction: "Minimum subtraction", that has 2 parameters:
1. Percentile (default: 0, which means minimal value)
2. Region in which the widget should search (default: whole spectrum)

### Contribution

If I didn't miss any existing solutions, I would be really willing to contribute this solution :) 
Please, tell if you think this is worthwhile and if you have additional suggestions for the implementation.

pisarik:
HI @stuart-cls,  I think this is a great suggestion! But the question is the eternal one: _flexible but complex_ vs _simple but inflexible_ 😄 

From my experience of tutoring at biospec summer school, where we use Orange:
* students without engineering background often struggle with the flexible solutions. For them, this additional step requires yet another memorization (I need to use **Integrate**), so it becomes a cognitive load. 
* On the other hand, advanced users often come up with some ad-hoc solutions and for them it is nice to be flexible when needed.

I would say that "Minimum subtraction" is such a basic option for baseline correction that I would expect it to be easily available. While theoretically I would definitely appreciate also opportunity to use an existing feature (why not?), but practically it is difficult to imagine what else function (other than `min`/`percentile`) I could possibly need to justify this additional flexibility. So maybe from this point of view there is no real need for the "existing feature" option in the case of baseline correction. 

[...]

PR #843, opened 2026-02-05
[ENH] Baseline shift correction

stuart-cls:
In-use prototype from @clsandt in #841 . I took the files there and applied as a patch.

Still needs 
 - [ ] Docs
 - [ ] Tests
 - [ ] Preprocessor could move to its own file
 - [ ] Visual indicators for value selection

@pisarik Do you have time to iterate on this? 

@markotoplak Is this another one of the pattern we were discussing in #823 for preprocessors?

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -974,3 +974,89 @@ class SpSubtract(Preprocess):
                                     data.domain.metas)
         return data.from_table(domain, data)
 
+class BaselineShiftFeature(SelectColumn):
+    InheritEq = True
+
+
+class _BaselineShiftCommon(CommonDomainOrderUnknowns):
+    """
+    Vectorized baseline shift transformation logic.
+    """
+
+    def __init__(self, method, wn=None, wn_range=None, domain=None):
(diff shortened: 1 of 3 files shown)


clsandt:
Personnally, I think it should go in the baseline corrections.
This is a type of baseline corrections in other spectroscopy softwares.

stuart-cls:
I agree with @pisarik , let's keep things at a high level (baseline subtraction, normalization/scaling) and not bother users with the implementation detail. So something like

 * Baseline correction
   * Constant value (static)
   * Constant value (calculated) [this PR]
   * Linear
   * Rubber band
 * Normalize Spectra
   * Constant value (static)
   * Vector
   * Min/max
   * Area
   * Attribute
   * SNV

We can write a settings migration to replace past usages of the shift-scale editor.

Just to stir the pot a little, I think of _spectrum_ subtraction and normalization to be a different class of thing. Currently we do one of each :D


@markotoplak my other question was about the `Preprocessor` implementation (and _not_ the editor presentation to users). I I see now that `Baseline` is of the style one-class-per-method, which I think is the architecture we want on the code side. Correct? 
`Normalize` is the older style like `Integrate`.

PR #846, opened 2026-02-18
[CI] Bump -oldest pandas version to 1.5.3

### LINKED ISSUE #845: [CI] -oldest tests fail
Looks like the culprit is the switch to python 3.11 means pandas 1.4.* wheel is not available (and will not build).

I'll put a PR to bump the -oldest to match orange3

stuart-cls:
Fixes #845 , only changed pandas since the rest already match the oldest from when orange3==3.38.0 

(maybe the logic is unclear here?)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -136,6 +136,7 @@ if __name__ == '__main__':
             'spectral>=0.22.3,!=0.23',
             'serverfiles>=0.2',
             'AnyQt>=0.2.0',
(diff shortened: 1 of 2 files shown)


markotoplak:
Thanks. Pandas should also be a direct dependency in setup.py, could you also add it there? Also, probably better to use 1.5.3 - it should be 3.11 compatible (well, available) and i always like to keep versions as free as possible.

PR #848, opened 2026-02-26
[ENH] Allow reading Unspecified type files in WiREReader

borondics:
Added the Unspecified option to the WiREReaders that allows for readi…ng a file with a single spectrum. 

No testing yet.


orangecontrib/spectroscopy/io/wire.py
```diff
@@ -21,6 +21,11 @@ def read_spectra(self):
                 table = self.series_reader(wdf_file)
             elif wdf_file.measurement_type == 3: # line scan
                 table = self.map_reader(wdf_file)
+            elif wdf_file.measurement_type == 0: # Unspecified
+                try:
```
markotoplak: This try ... except is fine for debugging, but for real code, there are two options:
- don't even look for exceptions: remove try ... catch
- instead of printing an exception, raise a specific exception (if nothing specific comes to mind, do the previous option)

The principle is that exceptions should not be just silently ignored (what having only printouts effectively means).

borondics:
The problem (I think) is that the unspecified category could be anything. 

Should I try to extend this to try all the readers? With several tries or a nested try-try-try?

markotoplak:
No, don't speculate, just handle what we know there is... So just this for now, and if there is a file where the reader handles and error, well, then we handle that.

borondics:
So, I found another "unspecified" file that fails for another reason - it seems we have no control over what is in those files. Therefore I came up with this new solution to loop through all readers and check if they can return the structure we expect (at least in case of the single_file for now) or raise some meaningful errors.

markotoplak:
It is great that you found the file, but still I'd look only use the loop when the type is undefined. Relying on exceptions could be potentially problematic.

I'd have a dict `{ measurement_type: reading function }`, and if the type is defined, just take the correct one, otherwise loop through the values of that dict, as you've done, with exceptions. This would make the could short.

PR #851, opened 2026-04-22
[ENH] Support Photothermal v5 .ptir files with ptir5 library

stuart-cls:
I added support for the ptir v5 files using the library from Photothermal: https://github.com/photothermal/PTIR5-py

The test files from that library aren't included in the package, so I added them here with a note about their source.

A few details of note:
 - It will take different wavenumber ranges from a single file and concatenate them (like OWMultiFile). But I'm not sure what happens if the wavenumber steps are not matched (I don't have a file to test this)
 - Single Wavenumber Images: If no point spectra or hypercubes are present, they are unpacked into a coordinate grid of 1-point spectra. However, if other spectral data is present, they are converted to 8-bit visible images and attached purely as background maps. I think this is consistent with what we were doing.

@borondics @kaiyang-tu can you test this? I did have one test file from Lund in addition to the example files in the library and it worked fine.

Edit: Note that until the `ptir5` package is released (https://github.com/photothermal/PTIR5-py/issues/4) you will need to manually install it from the repo to test.


stuart-cls:
Ah, I see the discrete maps -> data case for v4 was being discussed in #820 and not actually merged.

borondics:
@stuart-cls this will be super nice!!! Thank you so much! I will test it tonight.

borondics:
I think it is working but the file format is becoming more complex and we would need to support it somehow. @stuart-cls and @markotoplak I would like to re-discuss the HDF5 browser...

Some files I could open, others I couldn't and the visible images are also not available, so there could be a few improvements.

borondics:
So, the main problem is with the mixed files, when we have all kinds of entries. We discussed with @markotoplak and he agrees that the only solution is making a browser for the File widget, which would be quite a project. Let's see if we can manage this within some reasonable time, it would benefit everyone a lot. What do you think @stuart-cls?

stuart-cls:
Yes, @kaiyang-tu gave me similar feedback. I made a few changes (not quite tidied up) that should handle a few more cases, but I agree that the general browser / loader is needed. There's some nice conceptual models of datashapes in this library that match well with some of our existing helper functions.

That said, I think we can continue with this as a 80% solution for now.

stuart-cls:
@kaiyang-tu suggested we skip the labels (since they are unreliable/user-edited) and load the different types by user request (Spectra, Images, Hyperspectra). Only gotcha is you have to filter the DC/OPTIR labels later. Most mixed files should work now.

stuart-cls:
`ptir5` packages are released now and the CI passes

markotoplak:
@stuart-cls, what is the state of this one? could we merge this?

stuart-cls:
I was waiting for some feedback from @kaiyang-tu  and @borondics about this last change:

> skip the labels (since they are unreliable/user-edited) and load the different types by user request (Spectra, Images, Hyperspectra). Only gotcha is you have to filter the DC/OPTIR labels later.

It adds some complexity (which we usually avoid) in exchange for some flexibility in files that can be loaded.

PR #854, opened 2026-07-13
Add gating widget

csbrasnett:
This PR adds a gating widget, originally developed at DLS by Ben Nicholson. 

The widget allows for a 'physical' 'gate' to start/stop the flow of data between widgets in complex pipelines, so that the user has some more control over the processing. While this somewhat duplicates the flow breaking that can also be done, feedback from users has suggested that this is a helpful addition particularly for long pipelines.

Tests pass, ruff linting checked, docs tbd.


csbrasnett:
@markotoplak I think the failures here and in #856 are not directly related to the PR, but more general 3.14 support, is that right?

markotoplak:
The `MultiFile` test crash is unrelated to the PR, yes. It is not related to Python 3.14 though - it is caused by some changes in yet-unreleased Orange. The aim of these `orange-latest` tests is to warn us when something goes wrong so that we prepare for the future (either fix it there, if the compatibility broke in an ugly way, or adapt it here).

markotoplak:
@csbrasnett, I think I fixed most of the issues of the master branch, so you could rebase to master and push if you'd like more passing tests.

markotoplak:
@csbrasnett, please do not do merge commits in PRs, do a "rebase". This makes changelog cleaner (but does not infuence tests).

csbrasnett:
@markotoplak apologies, I think I've managed that on #854 now

borondics:
Hi @csbrasnett,

First of all, big thanks for being so active! 

We discussed this with @markotoplak and he likes the idea of the Gate widget, but I was suggesting to have this functionality implemented as an orange-canvas feature. Many other projects using Orange would benefit from it and I think the canvas would look nicer too... Maybe @janezd could add his thoughts?

Cheers!

csbrasnett:
Thanks @borondics, just to reiterate, these are all widgets that were previously developed but needed some bringing up to date and proper integration into quasar/orange, so I haven't been developing anything really novel yet. But nice to have some time to familiarise myself with the code base and how you guys like to run the show :)

Anyway, I don't particularly mind where any of these end up in the end, so if preferred I can close this PR and open one on the main Orange3 repo instead, just let me know!

orangecontrib/spectroscopy/widgets/owgate.py
```diff
@@ -0,0 +1,110 @@
...
+
+    return True
+
+class OWGate(OWWidget):
+    name = "Gate"
+    description = "A gate to control the flow of data."
+    icon = "icons/gate.svg"
+    id = "orangecontrib.spectroscopy.widgets.owgate"
+    priority = 10
+
```
markotoplak, line 13: Here there should be a line
`replaces = "orangecontrib.whatever.the.old.name.was"`
so that the old workflows could still open

csbrasnett: Does this matter even if the package hasn't been published anywhere?

markotoplak: It only matters for you - if you or your users had workflows using this widget saved, then, with this, they would still work. Whatever you do it is fine with me.

markotoplak:
Could you perhaps remove input comparison? That would perhaps help in the short term, but could mask upstream bugs. The previous widget should be careful not to give the same thing twice.

csbrasnett:
Hi @markotoplak, just wondering if this change is what you intended for this widget? I'll have a look at why the linting test fails as well

PR #855, opened 2026-07-14
[ENH Iterative Concave Ruberband Baseline Correction

samuelpinilla:
# Concave Rubberband Baseline Correction Preprocess

## 📌 Summary
This pull request introduces a  preprocessing method: **Iterative Concave Rubberband Baseline Correction**. The implementation is designed as an `Orange3` `Preprocess` subclass and integrates with the spectroscopy add-on pipeline.

The algorithm is from existing work (Samuel Pinilla at DLS) and refactored for compatibility with the Orange preprocessing framework, with a clear separation between computation and GUI layers.

---

## ✨ Key Features
- ✅ Implements **iterative concave rubberband baseline correction**
- ✅ Supports:
  - Baseline subtraction (default)
  - Baseline visualization mode
- ✅ Handles edge cases such as flat or degenerate spectra
- ✅ Efficient processing:
  - Sequential execution for small datasets
  - Multi-threaded execution for larger datasets
- ✅ Fully compatible with Orange’s domain transformation system

---

## 📖 Usage Example
```python
from orangecontrib.spectroscopy.preprocess import ConcaveRubberbandBaseline

pp = ConcaveRubberbandBaseline(n_iter=3)
processed_data = pp(data)

### DIFF
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -45,6 +45,8 @@ from orangecontrib.spectroscopy.preprocess.utils import (
     linear_baseline,
 )
 
+from orangecontrib.spectroscopy.preprocess.concaverubberband import ConcaveRubberbandBaseline  # noqa: E402
+
 
 class MNFDenoising(Preprocess):
     def __init__(self, components=None):
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -45,9 +45,6 @@ from orangecontrib.spectroscopy.preprocess.utils import (
     linear_baseline,
 )
 
-from orangecontrib.spectroscopy.preprocess.concaverubberband import ConcaveRubberbandBaseline  # noqa: E402
-
-
 class MNFDenoising(Preprocess):
     def __init__(self, components=None):
         self.components = components
--- a/orangecontrib/spectroscopy/preprocess/__init__.py
+++ b/orangecontrib/spectroscopy/preprocess/__init__.py
@@ -45,8 +45,6 @@ from orangecontrib.spectroscopy.preprocess.utils import (
     linear_baseline,
 )
 
-from orangecontrib.spectroscopy.preprocess.concaverubberband import ConcaveRubberbandBaseline  # noqa: E402
-
 
 class MNFDenoising(Preprocess):
     def __init__(self, components=None):
--- a/orangecontrib/spectroscopy/preprocess/concaverubberband.py
+++ b/orangecontrib/spectroscopy/preprocess/concaverubberband.py
@@ -3,6 +3,27 @@ Iterative concave rubberband baseline correction as a Preprocess subclass.
 
 Algorithm ported from algorithm.py (Samuel Pinilla). No Orange/Qt imports
 at the algorithm layer; GUI integration is handled in baseline.py.
+
+The iterative concave rubberband baseline correction is a multi-pass refinement
+of the classical rubberband algorithm. In the classical approach, the lower
+convex hull of the spectrum is computed once and its vertices are connected by
+straight-line segments to form a piecewise-linear baseline. The concave variant
+extends this in two ways. First, for wide spans between hull vertices — where a
+linear segment would either miss a broad background hump or cut into spectral
+features — it fits a downward-opening parabola that is tangent to the spectrum
+from below, producing a smoother and more conservative baseline estimate. Second,
+rather than applying this correction in a single pass, the algorithm proceeds
+iteratively: after each pass the estimated baseline is subtracted from the
+spectrum, and the hull and parabolic fitting are recomputed on this residual.
+Because the residual is progressively flatter, each subsequent iteration captures
+background contributions that were not fully resolved in the previous pass,
+tightening the baseline toward the true background floor. The Iterations parameter
+controls how many of these refinement passes are applied. A small value (1-3)
+applies a gentle correction suitable for spectra with a mild or nearly linear
+background; higher values (5-10) are appropriate for spectra with pronounced,
+broadly curved baselines. Excessive iterations should be avoided, as they risk
+pulling the estimated baseline into genuine spectral features, leading to
+overcorrection of the signal.
 """
 
 import os
--- a/orangecontrib/spectroscopy/preprocess/concaverubberband.py
+++ b/orangecontrib/spectroscopy/preprocess/concaverubberband.py
@@ -1,0 +1,206 @@
+"""
+Iterative concave rubberband baseline correction as a Preprocess subclass.
+
+Algorithm ported from algorithm.py (Samuel Pinilla). No Orange/Qt imports
+at the algorithm layer; GUI integration is handled in baseline.py.
+"""
+
+import os
+import threading
+import numpy as np
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from scipy.spatial import ConvexHull, QhullError
+
+import Orange.data
+from Orange.preprocess.preprocess import Preprocess
+
+from orangecontrib.spectroscopy.preprocess.utils import (
+    SelectColumn,
+    CommonDomainOrderUnknowns,
+)
+
+_PARALLEL_THRESHOLD = 32  # below this, skip thread-pool dispatch overhead
+_task_state = threading.local()
+# owpreprocess.run_task sets .value = state before pp(data); cleared in finally.
+
+
+def _correct_spectrum(
+    spectrum: np.ndarray, n_iter: int
+) -> tuple[np.ndarray, np.ndarray]:
+    """Apply concave rubberband correction to one spectrum.
+
+    :param spectrum: 1-D spectral intensities, shape (N,).
+    :param n_iter: number of hull-correction iterations (>= 1).
+    :returns: (corrected, baseline), both shape (N,) float64.
+    """
+    spectrum = np.ravel(spectrum)
+    N = len(spectrum)
+
+    baseline = np.zeros(N)
+    s = spectrum.copy()
+
+    for _ in range(n_iter):
+        points = np.column_stack((np.arange(1, N + 1), s))
+        try:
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: orangecontrib/spectroscopy/preprocess/__init__.py [hunk 1]
```diff
@@ -45,6 +45,8 @@
     linear_baseline,
 )
 
+from orangecontrib.spectroscopy.preprocess.concaverubberband import ConcaveRubberbandBaseline  # noqa: E402
```
markotoplak: Please remove this import. The others were imported for backward compatibility when some classes were restructured. Do it in a separate commit, so that it will be easy to rebase it out (to keep history as clean as possible).

orangecontrib/spectroscopy/preprocess/concaverubberband.py
```diff
@@ -0,0 +1,207 @@
...
+                             for chunk, off in zip(chunks, offsets)}
+            for fut in as_completed(future_to_off):
+                off = future_to_off[fut]
+                c, b = fut.result()
+                out[off: off + len(c)] = c
+                baselines[off: off + len(b)] = b
+                done += len(c)
+                if state is not None:
+                    state.set_progress_value(100 * done / M)
+    except Exception:
```
markotoplak: Why are catching exceptions here? Why would the upper code die? This looks like it would just mask errors. If there is a good reason, it should be more explicit.

orangecontrib/spectroscopy/widgets/preprocessors/baseline.py
```diff
@@ -4,7 +4,7 @@
 from AnyQt.QtWidgets import QVBoxLayout, QFormLayout, QPushButton, QApplication, QStyle
 
 from Orange.widgets import gui
-from orangecontrib.spectroscopy.preprocess import LinearBaseline, RubberbandBaseline
+from orangecontrib.spectroscopy.preprocess import LinearBaseline, RubberbandBaseline, ConcaveRubberbandBaseline
```
markotoplak: Here you could import from the new file directly (in regards to a previous comment).

orangecontrib/spectroscopy/widgets/owpreprocess.py
```diff
@@ -1050,7 +1053,11 @@ def progress_interrupt(i: float):
             pp = create_preprocessor(item, reference)
             plist.append(pp)
             if data is not None:
-                data = pp(data)
```
markotoplak, line 1053: This is not cool, it is a hacked-in intrusion into a base class.

orangecontrib/spectroscopy/preprocess/__init__.py
samuelpinilla: Thank you for the suggestion. This is solved in the latest commit. Please let me know what you think.

orangecontrib/spectroscopy/preprocess/concaverubberband.py
samuelpinilla: Thank you for the suggestion. This is solved in the latest commit. Please let me know what you think.

orangecontrib/spectroscopy/widgets/owpreprocess.py
samuelpinilla: Thank you for the suggestion. This is solved in the latest commit. Please let me know what you think.

markotoplak:
Thanks for the fixes.

So, that original progress-bar solution was a bit problematic (especially if they were multiple preprocessors), but I agree with your wish to include a nicer grained progress bar. I think the main preprocessing loop should do something similar to [what the Learner base class does](https://github.com/biolab/orange3/blob/86f208ce4434ce4f6925ecef397f2ccc1d0b3657/Orange/base.py#L120) - then individual preprocessors could report their status. I suggest we do that in another PR.

markotoplak:
Please also add a test similar to current tests for [Rubberband](https://github.com/Quasars/orange-spectroscopy/blob/b3db6ba5ac4eb90f716f8d88bd6346983db194f9/orangecontrib/spectroscopy/tests/test_preprocess.py#L394). You can just copy that class and adapt it to your preprocessors. That will then subject your preprocessors to various kinds of strange data (and domain conversion) tests.

samuelpinilla:
> Please also add a test similar to current tests for [Rubberband](https://github.com/Quasars/orange-spectroscopy/blob/b3db6ba5ac4eb90f716f8d88bd6346983db194f9/orangecontrib/spectroscopy/tests/test_preprocess.py#L394). You can just copy that class and adapt it to your preprocessors. That will then subject your preprocessors to various kinds of strange data (and domain conversion) tests.

Thank you. The last commit I added this. Please let me know.

orangecontrib/spectroscopy/widgets/preprocessors/baseline.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, parent=None, **kwargs):
             self,
             "baseline_type",
             items=["Linear", "Rubber band", "Concave Rubberband"],
-            callback=self.edited.emit,
+            callback=self._on_baseline_type_changed,
```
markotoplak: Why did you do this explicitely here? Why not leave self.edited.emit()? For me, the old version also works because emitting edited also causes the UI adaptation or is there some other difference?

samuelpinilla: Apologies for my delay. Took me some time, but I did some changes following your comments. Please see commit "small changes to hide unused parameters". Please let me know what you think. Thank you for your time.

orangecontrib/spectroscopy/preprocess/concaverubberband.py
```diff
@@ -0,0 +1,191 @@
...
+    if M < _PARALLEL_THRESHOLD:
+        for i, row in enumerate(X):
+            out[i], baselines[i] = _correct_spectrum(row, n_iter)
+        return out, baselines
+
+    chunk_size = max(1, M // (os.cpu_count() * 4))
+    offsets = list(range(0, M, chunk_size))
+    chunks = [X[off: off + chunk_size] for off in offsets]
+
+    with ThreadPoolExecutor(max_workers=os.cpu_count()) as ex:
```
markotoplak, line 148: @stuart-cls, in Peak Fit we needed to limit the number of processes because those were separate processes and this eating RAM, yes? Could you confirm that for threads there is no such limit is needed?

@samuelpinilla, what is the actual CPU utilization you get from this threaded code? Does it saturate your processor on big data? I am just wondering how effective it is in this usecase because python's GIL prevents some operation working in parallel - for large numpy operations this parallelism works really well, but for python code usually not.

orangecontrib/spectroscopy/preprocess/concaverubberband.py
```diff
@@ -0,0 +1,191 @@
+"""
+Iterative concave rubberband baseline correction as a Preprocess subclass.
+
+Algorithm ported from algorithm.py (Samuel Pinilla). No Orange/Qt imports
+at the algorithm layer; GUI integration is handled in baseline.py.
```
markotoplak, line 5: Could you perhaps write a paragraph about how algorithm works here? So that we could eventually put it into the documentation? If I remember correctly, this is your algorithm that hasn't been published separately, right? If it was, it would be nice to have a reference, but if it is not, a good description is crucial for both users and maintainers. 

The description should also point out how it differs from from the "rubberband algorithm" and what exactly is the iterative aspect of it - at least how does increasing the number of iterations change the correction.

markotoplak:
@samuelpinilla, thank you for this PR. I think this is a really useful piece of work.

When you manage, please add a short algorithm description (as I commented before). Then I'll rebase this to master (to remove the merge commit), lint it and merge (if you want, you can also do these steps yourself). Just writing this so that you know I won't bug you for fixes anymore. :)

orangecontrib/spectroscopy/preprocess/concaverubberband.py
samuelpinilla: Truly apologies for the long delay. Thank you for the comment. The last commit I added this. Please let me know.

samuelpinilla:
> @samuelpinilla, thank you for this PR. I think this is a really useful piece of work.
> 
> When you manage, please add a short algorithm description (as I commented before). Then I'll rebase this to master (to remove the merge commit), lint it and merge (if you want, you can also do these steps yourself). Just writing this so that you know I won't bug you for fixes anymore. :)

Thank you very much for your comments, and kind words. Not at all, I really appreciate the time and the comments. Better to review and agree changes :). I added the description please let me know if it is a good one.

orangecontrib/spectroscopy/preprocess/concaverubberband.py
samuelpinilla: Thanks for the question. I did several checks and the CPU utilization is not saturated. And, general to avoid this the size of the processed spectra per processor is given by chunk_size = max(1, M // (os.cpu_count() * 4)). Since it is multiplied by 4 it is small every time. But please let me know about this if you see fit.

markotoplak: I just wondered how much utilization you get. It depends on the code, but if the code keeps GIL locked (and thus could only utilize 1 CPU) I would even skip this.

orangecontrib/spectroscopy/preprocess/concaverubberband.py
markotoplak: Thank you for the explanation. Was very clear and I finally understand what it does. :)

orangecontrib/spectroscopy/preprocess/concaverubberband.py
samuelpinilla: I did some tests on real size images (64x64x3600) here at Diamond, and this method uses 50% of all processors. So when it comes to speed there will be a benefit of using more than 1 CPU. What do you think?

markotoplak: Cool, keep it as-is.

markotoplak:
I added lints into respective commits and am merging. Thank you for your contribution.

PR #856, opened 2026-07-14
Elementwise operations

csbrasnett:
Add elementwise operations to spectroscopy toolkit. These take input table(s) and perform direct operations on them.

elementwise-where: 
- where data is within some threshold of an input value, it gets replaced by a singular new value
- min/max value replacement. I.e. if data are >/< some input value, then replace them with one value

elementwise-operations:
- perform simple arithmetic operations on the primary table, based on the secondary table.

tests are passing and linting checked.


markotoplak:
@borondics, could you also think of widget names suggestions when you are testing this? I don't think "elementwise" carries lots of meaning. I was thinking of "Replace value" and "Calculator" or something in this vein.

borondics:
Hi guys! I am on it. Already checked the widgets and have some suggestions. A long-ish document is to come in a few days.

borondics:
## Elementwise Where

### Consider the following names
- Replace values 
- Search and replace

### Current problems
- When a workflow is loaded with this widget, the data flows into it, but the calculation or emission of the data doesn't happen automatically.

#### Consider the following GUI improvements:
- Currently, the widget functionality description is on the GUI. Please implement this as a help file instead to describe the different 
- Remove the info box and add that also to the documentation
- There are three distinct and mutually exclusive calculations with almost the same parameters (except the "Tolerance"). Instead of having all of them triplicated and using radio buttons to select the type of replacement it would be nice to have the options available as a drop-down selection.
- The "Modify Data Values in Place Automatically" is a very long message. Shorten it to "Replace values" or similar.
- The icon is too large, it sticks out on the corners. 

## Elementwise operations
We already have "Shift and Scale" in Preprocess Spectra. Ideally, we could remove those preprocessors and write a robust "Calculator" widget.

### Consider the following names
- Calculator
- Spectrum Calculator

### Current problems
When a workflow is loaded with this widget, the data flows into it, but the calculation or emission of the data doesn't happen automatically.

#### Consider the following GUI improvements:
- Please start implementing a help file
- Remove the info box and add that also to the documentation
- If we decide to replace the preprocessors, the GUI needs to be designed nicely, like a block-calculator, where `inA + inB / inB` style calculations can be done. It could be graphical or symbolic text based.

csbrasnett:
Thanks @borondics can I ask, when you say 'help files' and 'documentation', do you mean the same feature that the 'help' option on a widget links to the readthedocs site?

markotoplak:
Just a note. If the widget's:

- UI is small (few options to choose)
- and the behaviour is non-obvious

Then it could make sense to put widget's description into the widget. But this depends on a particular widget... 

For example, see this one. 

image

Or maybe

image

csbrasnett:
# Checklist from requested changes:

## Elementwise Where

### Consider the following names
- ~Replace values~
- ~Search and replace~
**widget now called `replace`**


### Current problems
- When a workflow is loaded with this widget, the data flows into it, but the calculation or emission of the data doesn't happen automatically.
**not done - please could you indicate the correct way of doing this? The variety of commit methods is currently confusing me**

### Consider the following GUI improvements:
- ~Currently, the widget functionality description is on the GUI. Please implement this as a help file instead to describe the different~
- ~Remove the info box and add that also to the documentation~
**Docs page now added for this widget**


- ~There are three distinct and mutually exclusive calculations with almost the same parameters (except the "Tolerance"). Instead of having all of them triplicated and using radio buttons to select the type of replacement it would be nice to have the options available as a drop-down selection.~
**This is now done.**

- ~The "Modify Data Values in Place Automatically" is a very long message. Shorten it to "Replace values" or similar.~
**message changed now**
- ~The icon is too large, it sticks out on the corners.~
**icon edited**

## Elementwise operations
We already have "Shift and Scale" in Preprocess Spectra. Ideally, we could remove those preprocessors and write a robust "Calculator" widget.

### Consider the following names
- ~Calculator~
- ~Spectrum Calculator~
**now renamed `spectrum calculator`**

### Current problems
When a workflow is loaded with this widget, the data flows into it, but the calculation or emission of the data doesn't happen automatically.
**as above**

#### Consider the following GUI improvements:
- ~Please start implementing a help file~
- ~Remove the info box and add that also to the documentation~
**docs file now started**

- If we decide to replace the preprocessors, the GUI needs to be designed nicely, like a block-calculator, where `inA + inB / inB` style calculations can be done. It could be graphical or symbolic text based. 
**I can see arguments both ways for this. I don't mind ultimately what happens, but one key difference here IMO is that the preprocessor only takes a single input file, and this widget takes two to perform operations using both. So that we should bear in mind.**

csbrasnett:
Hi @borondics if you have time to look at this again it'd be greatly appreciated, thanks!

PR #858, opened 2026-07-21
Phase unwrapping

csbrasnett:
This PR adds an unwrapping widget, originally developed at DLS by Ben Nicholson. The widget is a wrapper around the [numpy unwrap function](https://numpy.org/doc/stable/reference/generated/numpy.unwrap.html). 

For the moment, it uses default values for the function as documented. I've been assured that these are sufficient, but we could have the option of varying them. However, I'm unsure atm what would be the best route for this. Some possible options from the top of my head - 

- radio buttons with 'default' and 'custom', where `custom` then has activated input boxes
- sliders with default values. This may be tricky because I imagine we really want regular intervals (1/4,1/2..1, 2, etc) of \pi, rather than arbitrary float values

The user feedback on the widget is happy with it as it is, so potentially adding more functionality to the GUI could be out of scope of this PR and can be added later once the basic widget is integrated.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -37,6 +37,7 @@ Widgets
    widgets/peakfit
    widgets/snr
    widgets/polar
+   widgets/unwrap
 
 Indices and tables
 ------------------
--- a/doc/widgets/unwrap.md
+++ b/doc/widgets/unwrap.md
@@ -1,0 +1,10 @@
+Unwrap signal
+=============
+
+This widget is a wrapper for the [NumPy signal unwrapping function](https://numpy.org/doc/stable/reference/generated/numpy.unwrap.html). It uses the default arguments of:
+
+```
+wrapped signal = unwrap(input signal, discont = pi, period = 2*pi)
+```
+
+This unwrapping is calculated for every row of the input data table.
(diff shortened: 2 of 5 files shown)


borondics:
We have also been doing some pahse unwrapping work with @ngergihun previously. I just want to make sure that there is no duplication and everything is working as intended. 

@ngergihun and @csbrasnett could you please discuss this PR?

markotoplak:
@ngergihun, please don't forget about this one. You can also join us on regular meetings with the Diamond team.

ngergihun:
Hi all,

Thanks, @markotoplak, for the reminder. @csbrasnett, correct me if I misunderstand it, but as far as I can see, it is a simple NumPy unwrap. Of course, it would be nice to have more control.
We already have a simple phase unwrap preprocessor in the Preprocess Spectra and Preprocess Image widgets. I think we should advance that preprocessor further with some options to define wrapping intervals.

I don't think the unwrapping needs as much control as a separate widget. It's also easier to integrate as a preprocessor. And makes more sense to me as all the other spectral transformations are implemented as preprocessors.

csbrasnett:
@ngergihun nope, it's really just np.unwrap. Apologies, please could you point to where this is in the existing Preprocess Spectra widget? I can't seem to identify what you're mentioning here. Happy to work on adding to the current preprocessor if that's what's more useful! I'll also discuss this with users at DLS to make sure they're aware.

csbrasnett:
@ngergihun just checking up on this again, are you thinking of the phase unwrapping in the SNOM module? https://github.com/Quasars/orange-snom/blob/master/orangecontrib/snom/preprocess/phase_unwrap.py

